PR #23, opened 2021-08-20
Add workflow for publishing cdk and connector images

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


cjwooo:
@tovbinm Please add our standard NPM publishing token to the Github repo secrets as `GITHUB_TOKEN`.

.github/workflows/ci.yml
```diff
@@ -59,3 +59,91 @@ jobs:
           done
         env:
           EXAMPLE_SOURCE_TEST_CREDS: ${{ secrets.EXAMPLE_SOURCE_TEST_CREDS }}
+
+  publish-cdk:
+    name: Publish CDK
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
```
tovbinm, line 66: 5 minutes

.github/workflows/ci.yml
```diff
@@ -59,3 +59,91 @@ jobs:
...
+        with:
+          node-version: 14
+
+      - name: Extract version
+        id: extract-version
+        run: |
+          CDK_VERSION="$(node -p 'require("./lerna.json").version')"
+          echo "CDK_VERSION=$CDK_VERSION" >> $GITHUB_ENV
+
+      - name: Check tag exists
```
tovbinm, line 86: So what the process of release would be?
1. bump the version in the code
2. the release triggers
3. creates a tag on Github

Correct?

.github/workflows/ci.yml
```diff
@@ -59,3 +59,91 @@ jobs:
...
+          node-version: 14
+
+      - name: Docker login
+        uses: docker/login-action@v1.10.0
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build connector packages
+        run: lerna bootstrap --hoist
```
tovbinm, line 135: Dont we do run `lerna bootstrap --hoist` in the docker container already?

package.json
```diff
@@ -1,6 +1,14 @@
 {
   "name": "root",
   "private": true,
+  "scripts": {
+    "bump": "lerna version --no-git-tag-version --preid rc",
+    "build": "lerna run build",
+    "clean": "lerna run clean && rm -rf node_modules out",
+    "cover": "lerna run test -- -- --coverage --color",
```
tovbinm, line 8: rename to `test-cov`

tovbinm (commented):
lets also exclude `*.log` and `*.out` files in `.dockerignore` file + all the `devDependencies` we dont need

.github/workflows/ci.yml
cjwooo: Correct

cjwooo:
@tovbinm Please also add our standard docker hub creds into Github Secrets: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`

tovbinm:
@cjwooo done

.github/workflows/ci.yml
```diff
@@ -59,3 +59,88 @@ jobs:
...
+        id: check-tag
+        with:
+          tag: v${{ env.CDK_VERSION }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Publish packages
+        if: steps.check-tag.outputs.exists == 'false'
+        run: |
+          lerna bootstrap --hoist
```
tovbinm, line 97: `lerna bootstrap --hoist && cd cdk && npm publish --access public --tag latest`

tovbinm: also perhaps better?
```
lerna bootstrap --hoist && lerna run npm publish --scope cdk -- --access public --tag latest`
```

tovbinm (approved):

.github/workflows/ci.yml
```diff
@@ -59,3 +59,88 @@ jobs:
...
+
+      - name: Setup Node
+        uses: actions/setup-node@v2.3.0
+        with:
+          node-version: 14
+
+      - name: Extract version
+        id: extract-version
+        run: |
+          CDK_VERSION="$(node -p 'require("./lerna.json").version')"
```
tovbinm, line 83: let's rename `CDK_VERSION` to `RELEASE_VERSION` since we use it to tag all the images as well

.github/workflows/ci.yml
```diff
@@ -59,3 +59,88 @@ jobs:
           done
         env:
           EXAMPLE_SOURCE_TEST_CREDS: ${{ secrets.EXAMPLE_SOURCE_TEST_CREDS }}
+
+  publish-cdk:
```
tovbinm, line 63: let's put the CDK & Docker images release flows into `release.yml` file

tovbinm (commented):
Also can you please add instructions the README on how to release the connectors

.github/workflows/ci.yml
cjwooo: Clarification, it doesn't trigger a "GitHub Release". It just creates a new tag.

.github/workflows/ci.yml
cjwooo: `lerna run npm publish` is not supported, and putting all the commands on one line seems less readable.

.github/workflows/ci.yml
cjwooo: I'm tagging the connector images by sha, as we normally do, not by package.json version.

cjwooo: The connector images are released with every push. Only the CDK relies on the tag creation to be published to NPM.

tovbinm: Hmm, I think we should also have some versioning for connector containers as well.

.github/workflows/ci.yml
tovbinm: Then perhaps it's better to have us manually trigger the release? Same as we do for the feeds.

cjwooo: The problem My concern with that is the release, tag version, and package.json can differ, and `npm publish` uses the package.json version, so it could get confusing if the release/tag are created manually and the package.json version doesn't match.

cjwooo: The advantage of the current approach is that the developer only needs to run `npm run bump` on their machine and commit it, and the tag and NPM package will automatically be created with the correct version.

tovbinm: ok, let's keep it like you're proposing for now.

.github/workflows/ci.yml
cjwooo: I'm ok with that.

tovbinm: I think lerna support versioning of individual sub-projects. We could use that version to tag containers as well.

cjwooo: That would require the workflow to check the change diff to identify the connector packages that bumped versions. Not sure how doable that is with GHA.

cjwooo: Because we're gonna use git tags to mark CDK releases.

cjwooo: Actually, we can check if a docker tag already exists for the package.json version for each connector and use that to determine whether or not to publish.

tovbinm: Great! Let's do it.

tovbinm:
🥳 

@cjwooo let's make our first release ;)

PR #25, opened 2021-08-23
Handle raw Airbyte records in Faros destination + more GH converters

tovbinm:
## Description
1. Handle raw Airbyte records in Faros destination. This enables Faros destination to handle inputs in two scenarios:
- GitHub source -> Faros destination
- GitHub source -> Postgres destination (raw), and then Postgres source (raw) -> Faros destination

2. Added more GitHub converters

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


cdk/src/protocol.ts
```diff
@@ -128,6 +129,18 @@ export class AirbyteRecord implements AirbyteMessage {
     }
   ) {}
 
+  unpackRaw(): AirbyteRecord {
+    const stream = this.record.stream;
+    if (!stream || !stream.startsWith('_airbyte_raw_')) {
```
cjwooo, line 134: Make a global const for `_airbyte_raw_`

destinations/faros-destination/src/converters/github/issues.ts
```diff
@@ -0,0 +1,97 @@
+import {AirbyteRecord} from 'cdk';
+import {Utils} from 'faros-feeds-sdk/lib';
```
cjwooo, line 2: Why is the `/lib` there?

tovbinm: removed

cdk/src/protocol.ts
tovbinm: will do

cjwooo (approved):

PR #62, opened 2021-09-03
Dynamicaly load converters by stream name in Faros destination

tovbinm:
## Description

Dynamicaly load converters by stream name in Faros destination.
No more manual registering of classes. We now dynamically load the converter by name.

Example: for stream `blahblahblah__github__pull_request_stats` we dynamically import module `./github/pull_request_stats` and create an instance of the converter class `GithubPullRequestStats`.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


cjwooo (approved):
Cool!

eskrm (approved):

tovbinm:
FYI `require` also works instead of `import`

PR #113, opened 2021-10-01
FAI-441 - Added revisions & users streams to Phabricator source + bump version

tovbinm:
## Description

1. Added revisions & users streams to Phabricator source + bump version
2. Jenkins source fixes

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change

# Fixes
FAI-441

### DIFF
--- a/sources/phabricator-source/src/phabricator.ts
+++ b/sources/phabricator-source/src/phabricator.ts
@@ -1,3 +1,4 @@
+import AsyncLock from 'async-lock';
 import {Condoit} from 'condoit';
 import iDiffusion from 'condoit/dist/interfaces/iDiffusion';
 import {
@@ -88,6 +89,8 @@ interface PagedResult<T> extends ErrorCodes {
 }
 
 export class Phabricator {
+  private static lock = new AsyncLock();
+  private static phabricator: Phabricator;
   private static repoCacheById: Dictionary<Repository, phid> = {};
   private static repoCacheByName: Dictionary<Repository, string> = {};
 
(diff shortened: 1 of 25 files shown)


sources/phabricator-source/src/phabricator.ts
```diff
@@ -1,6 +1,11 @@
 import {Condoit} from 'condoit';
 import iDiffusion from 'condoit/dist/interfaces/iDiffusion';
-import {ErrorCodes, phid} from 'condoit/dist/interfaces/iGlobal';
+import {
```
cjwooo, line 3: Why are these imports from the `condoist/dist` folder and not just from `condoit`?

sources/phabricator-source/src/phabricator.ts
```diff
@@ -217,19 +270,101 @@ export class Phabricator {
...
+
+    const constraints = {repositoryPHIDs, modifiedStart: modified};
+    const attachments = {projects: false, subscribers: true, reviewers: true};
+
+    yield* this.paginate(
+      limit,
+      (after) =>
+        this.client.differential.revisionSearch({
+          queryKey: 'all',
+          order: 'updated',
```
cjwooo, line 313: You're using `stateCheckpointInterval()` to update the source state and more frequent intervals. This requires the order of the stream's outputted records to be in order of oldest to newest, so that a failed or canceled sync does not cause the next sync to miss records. Are you positive that both this stream and the users stream meets this ordering requirement? I don't see a "direction" parameter in the calls to the Phabricator API.

sources/phabricator-source/src/phabricator.ts
tovbinm: I dont think they export it properly. I will double check.

sources/phabricator-source/src/phabricator.ts
tovbinm: Hmm, I am actually querying newest -> oldest. Should I remove `stateCheckpointInterval` completely?

tovbinm: FYI, for example for commits search: 
Screen Shot 2021-10-01 at 12 59 11 PM

sources/phabricator-source/src/phabricator.ts
tovbinm: Nope, they dont export interfaces in the index.ts

sources/phabricator-source/src/phabricator.ts
cjwooo: Either change the search queries to sort from oldest -> newest or remove `stateCheckpointInterval`.

tovbinm: Same for Jenkins?

tovbinm: I actually like to be able to query from newest -> oldest this make incremental runs very fast.

cjwooo: Yes same for Jenkins, I didn't notice you added it there too.

tovbinm: Removed. Good catch!!

tovbinm:
@cjwooo please have another look

cjwooo (approved):

PR #114, opened 2021-10-01
FA-441 - Added projects stream to Phabricator + cleanup

tovbinm:
## Description

Added projects stream to Phabricator + cleanup

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Issue
FA-441

### DIFF
--- a/sources/phabricator-source/src/phabricator.ts
+++ b/sources/phabricator-source/src/phabricator.ts
@@ -20,7 +20,8 @@ export interface PhabricatorConfig {
   readonly server_url: string;
   readonly token: string;
   readonly start_date: string;
-  readonly repositories: string;
+  readonly repositories: string | string[];
+  readonly projects: string | string[];
   readonly limit: number;
 }
 
@@ -99,6 +100,7 @@ export class Phabricator {
     readonly client: Condoit,
     readonly startDate: Moment,
     readonly repositories: string[],
+    readonly projects: string[],
     readonly limit: number,
     readonly logger: AirbyteLogger
   ) {}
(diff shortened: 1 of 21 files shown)


sources/phabricator-source/src/phabricator.ts
```diff
@@ -376,4 +390,38 @@ export class Phabricator {
...
+
+    const constraints = {phids: filter.projectIds ?? []};
+    const attachments = {members: true, ancestors: true, watchers: true};
+
+    yield* this.paginate(
+      limit,
+      (after) => {
+        return this.client.project.search({
+          queryKey: 'all',
+          order: 'newest' as any,
```
cjwooo, line 412: I'm suprised the `as any` is needed.

sources/phabricator-source/src/phabricator.ts
```diff
@@ -376,4 +390,38 @@ export class Phabricator {
       }
     );
   }
+
+  async *getProjects(
+    filter: {
+      projectIds?: phid[];
```
cjwooo, line 396: I don't see where `projectIds` is ever populated.

sources/phabricator-source/src/phabricator.ts
tovbinm: the Condoit types are outdated

sources/phabricator-source/src/phabricator.ts
tovbinm: added

tovbinm: ok, it's not used actually, but let's keep it

tovbinm: ok! I added project filter as well

PR #115, opened 2021-10-03
Fix Jenkins url

Roma-Kyrnis:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/faros-destination/src/converters/jenkins/jobs.ts
+++ b/destinations/faros-destination/src/converters/jenkins/jobs.ts
@@ -1,6 +1,11 @@
 import {AirbyteRecord} from 'faros-airbyte-cdk';
 
-import {Converter, DestinationModel, DestinationRecord, StreamContext} from '../converter';
+import {
+  Converter,
+  DestinationModel,
+  DestinationRecord,
+  StreamContext,
+} from '../converter';
 import {JenkinsCommon, Job} from './common';
 
 export class JenkinsJobs extends Converter {
@@ -8,7 +13,7 @@ export class JenkinsJobs extends Converter {
     'cicd_Organization',
     'cicd_Pipeline',
   ];
-  
+
   id(record: AirbyteRecord): any {
     return record?.record?.data?.fullName;
   }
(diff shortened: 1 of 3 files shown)


tovbinm (approved):

tovbinm (changes requested):
Can you please provide a clean description what is this change about?

Roma-Kyrnis:
Description updated

destinations/faros-destination/src/converters/jenkins/jobs.ts
```diff
@@ -19,11 +24,12 @@ export class JenkinsJobs extends Converter {
...
     const source = this.streamName.source;
     const job = record.record.data as Job;
-    
+
     const jenkinsUrl = JenkinsCommon.parseJenkinsUrl(job.url);
     if (!jenkinsUrl) return [];
     const organization = JenkinsCommon.cicd_Organization(jenkinsUrl, source);
-    const orgKey = {uid: organization.record.uid, source}
+    const orgKey = {uid: organization.record.uid, source};
+    job.url = jenkinsUrl.url;
```
tovbinm, line 32: why do you need to update job url?

Roma-Kyrnis: To use parsed url everywhere

Roma-Kyrnis: When parsing url it can add slash character at the end if It doesn't exist

Roma-Kyrnis: Do you advice not to do it?

tovbinm: I think we should keep the job url untouched, as it is returned from the API response.

tovbinm: What are we doing in the original Jenkins feed?

Roma-Kyrnis: In the feed we save Jenkins's origin url

PR #153, opened 2021-11-09
Add Harness source and Faros destination's converter

### LINKED ISSUE #152: Add Harness source and Faros converter
Add a Harness source and converter based on [Harness](https://github.com/faros-ai/feeds/tree/master/feeds/cicd/harness-feed) feed

Roma-Kyrnis:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix #152 

## Migration notes

## Extra info


sources/harness-source/resources/spec.json
```diff
@@ -0,0 +1,51 @@
...
+    "properties": {
+      "api_key": {
+        "type": "string",
+        "title": "Harness API key",
+        "airbyte_secret": true
+      },
+      "account_id": {
+        "type": "string",
+        "title": "Harness account ID",
+        "airbyte_secret": true
```
cjwooo, line 21: Does the account id need to be secret?

cjwooo:
Is there a test Harness server we can spin up to test, similar to how we do Jenkins, or do we need to test against a live Harness account?

destinations/faros-destination/src/converters/harness/executions.ts
```diff
@@ -0,0 +1,326 @@
...
+    } else {
+      return undefined;
+    }
+  }
+
+  private parseWorkflow(
+    execution: HarnessExecutionNode,
+    source: string
+  ):
+    | {
```
cjwooo, line 146: This seems pretty clunky read. Can you make the non-undefined return type a local interface?

sources/harness-source/resources/spec.json
Roma-Kyrnis: Account id can be public. I'll change it to not be a secret

destinations/faros-destination/src/converters/harness/executions.ts
Roma-Kyrnis: Ok. I've created interface

Roma-Kyrnis:
Harness has [delegate](The Harness Delegate is a service). It's service but not Harness's local instance

cjwooo (approved):

tovbinm:
Yay!

Please update README

PR #271, opened 2022-02-14
Community edition: support resetting data

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


destinations/faros-destination/src/converters/jira/common.ts
```diff
@@ -10,10 +10,10 @@ export interface Assignee {
 }
 
 export enum RepoSource {
-  BITBUCKET = 'Bitbucket',
-  GITHUB = 'GitHub',
-  GITLAB = 'GitLab',
-  VCS = 'VCS',
+  BITBUCKET = 'bitbucket',
```
tovbinm, line 13: This is a breaking change! In our feeds and events we use the camel casing for the repo sources.

destinations/faros-destination/src/index.ts
```diff
@@ -575,6 +602,7 @@ class FarosDestination extends AirbyteDestination {
     );
 
     return {
+      origin: origins[0],
```
tovbinm, line 605: Lets revert this for SaaS version and create a Faros Destination per account.

destinations/faros-destination/src/converters/github/common.ts
```diff
@@ -157,6 +157,12 @@ export abstract class GithubConverter extends Converter {
   id(record: AirbyteRecord): any {
     return record?.record?.data?.id;
   }
+
+  get streamName(): StreamName {
```
tovbinm, line 161: where do you use this function?

cjwooo: https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/faros-destination/src/converters/github/repositories.ts#L19 and similar for other github converters.

tovbinm: Oh, it's on the `GithubConverter` base. Makes sense.

tovbinm: We do provide a default implementation thought in most base `Converter` class
https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/faros-destination/src/converters/converter.ts#L13

Not sure why we override this.

cjwooo: As I said in today's meeting, I need to match the source of the records outputted by the github converters with the references written by the jira converters so they will be properly connected. The faros destination writes all the sources as lowercase, which doesn't match what our existing feeds and therefore our jira converters write.

tovbinm: Do you have any ideas on how can we fix this properly?

cjwooo: Stop using streamName.source to populate records' source field and make source a property of the Converter class. Every source system has its own way of casing its name.

tovbinm (approved):

PR #279, opened 2022-02-16
Update Harness and Pagerduty sources

Roma-Kyrnis:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


cjwooo:
I think we should keep the cutoff parameters. They're useful for users with a lot of source data who don't want to pull all that historical data. If the problem was the cutoff breaking the tests, let's just set the cutoff parameter for testing to an extreme value like 5 years ago or something. @tovbinm

tovbinm:
Yes, we absolutely need to have either cutoff or start date.

Roma-Kyrnis:
Can you add `"cutoff_days": 50000` option to Harness and Pagerduty config?

Do we need to add cutoff to other connectors?

tovbinm:
@Roma-Kyrnis what's the wait here?

Roma-Kyrnis:
I'm updating to main branch and you can merge it

cjwooo (approved):

PR #281, opened 2022-02-16
Adds Segment tracking support to Community Edition

ypc-faros:
## Description

Added a segment_user_id field under the Community edition configuration. If specified, it will be used to track the write stats with Segment.
If no segment user id is specified we just not track the stats.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/faros-destination/resources/spec.json
+++ b/destinations/faros-destination/resources/spec.json
@@ -43,26 +43,10 @@
                 "description": "The Community Edition Hasura URL.",
                 "default": "http://localhost:8080"
               },
-              "segment_config" : {
-                "type": "object",
-                "title": "Segment configuration",
-                "required": [
-                  "segment_api_key",
-                  "segment_user_id"
-                ],
-                "properties": {
-                  "segment_api_key": {
-                    "type": "string",
-                    "title": "Segment API key",
-                    "description": "Segment API key",
-                    "default": "YEu7VC65n9dIR85pQ1tgV2RHQHjo2bwn"
-                  },
-                  "segment_user_id": {
-                    "type": "string",
-                    "title": "Segment User Id",
-                    "description": "The User Id with which to track events in Segment"
-                  }
-                }
+              "segment_user_id": {
+                "type": "string",
+                "title": "Segment User Id",
+                "description": "The User Id with which to track events in Segment. If not present, then reporting is disabled."
               }
             }
           },
--- a/destinations/faros-destination/resources/spec.json
+++ b/destinations/faros-destination/resources/spec.json
@@ -42,6 +42,27 @@
                 "title": "Hasura URL",
                 "description": "The Community Edition Hasura URL.",
                 "default": "http://localhost:8080"
+              },
+              "segment_config" : {
+                "type": "object",
+                "title": "Segment configuration",
+                "required": [
+                  "segment_api_key",
+                  "segment_user_id"
+                ],
+                "properties": {
+                  "segment_api_key": {
+                    "type": "string",
+                    "title": "Segment API key",
+                    "description": "Segment API key",
+                    "default": "YEu7VC65n9dIR85pQ1tgV2RHQHjo2bwn"
+                  },
+                  "segment_user_id": {
+                    "type": "string",
+                    "title": "Segment User Id",
+                    "description": "The User Id with which to track events in Segment"
+                  }
+                }
               }
             }
           },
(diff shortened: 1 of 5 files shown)


destinations/faros-destination/resources/spec.json
```diff
@@ -42,6 +42,27 @@
...
+              },
+              "segment_config" : {
+                "type": "object",
+                "title": "Segment configuration",
+                "required": [
+                  "segment_api_key",
+                  "segment_user_id"
+                ],
+                "properties": {
+                  "segment_api_key": {
```
cjwooo, line 54: cc @tovbinm I think we should not allow changing the api key.

cjwooo: Also wondering if we should not mention segment in the configuration and just refer to a "User Id". And should users be able to change their user id?

package.json
```diff
@@ -41,5 +41,8 @@
     "ts-jest": "^27.1.3",
     "ts-node": "^10.4.0",
     "typescript": "^4.5.5"
+  },
+  "dependencies": {
```
cjwooo, line 45: I'm guessing you ran `npm i analytics-node` from the repository root. Please `npm uninstall analytics-node` from the root, then cd to `destinations/faros-destination` and run `npm i analytics-node` there. Our structure is we specify submodule-specific dependencies in their respective package.jsons and keep devDependencies (dependencies that are only required for compilation and not during runtime) in the root package.json.

destinations/faros-destination/resources/spec.json
tovbinm: sounds good. let's put the key in the code.

destinations/faros-destination/src/index.ts
```diff
@@ -136,6 +138,18 @@ class FarosDestination extends AirbyteDestination {
     } catch (e) {
       throw new VError(`Invalid Hasura url. Error: ${e}`);
     }
+
+    if (config.edition_configs.segment_config) {
+      if (!config.edition_configs.segment_config.segment_api_key) {
+        throw new VError('Community Edition Segment API key is not set');
+      }
+      if (!config.edition_configs.segment_config.segment_user_id) {
```
tovbinm, line 146: is `segment_user_id` is not set - no need to error, simply skip reporting.

destinations/faros-destination/resources/spec.json
```diff
@@ -42,6 +42,27 @@
                 "title": "Hasura URL",
                 "description": "The Community Edition Hasura URL.",
                 "default": "http://localhost:8080"
+              },
+              "segment_config" : {
+                "type": "object",
+                "title": "Segment configuration",
```
tovbinm, line 48: lets put description here explaining that if segment_user_id the reporting is disabled and link to docs where we explain what is being tracked

destinations/faros-destination/src/index.ts
```diff
@@ -373,7 +387,16 @@ class FarosDestination extends AirbyteDestination {
         );
       }
 
-      // TODO: report stats to Segment if segment key is provided
+      if (this.analytics) {
+        this.logger.info('Sending stats to Segment.');
+        this.analytics.track({
+          event: 'Records written',
```
tovbinm, line 393: `event: 'Write Stats'`

destinations/faros-destination/src/index.ts
```diff
@@ -373,7 +387,16 @@ class FarosDestination extends AirbyteDestination {
         );
       }
 
-      // TODO: report stats to Segment if segment key is provided
+      if (this.analytics) {
+        this.logger.info('Sending stats to Segment.');
```
tovbinm, line 391: `this.logger.info('Sending write stats to Segment.')`

destinations/faros-destination/src/index.ts
```diff
@@ -373,7 +387,16 @@ class FarosDestination extends AirbyteDestination {
...
+        this.analytics.track({
+          event: 'Records written',
+          userId: config.edition_configs.segment_config.segment_user_id,
+          stats,
+        });
+      } else {
+        this.logger.info('Segment tracking not configured.');
```
tovbinm, line 398: no need to log this.

tovbinm (approved):
very cool! just a few comments

destinations/faros-destination/resources/spec.json
cjwooo: Also don't need to wrap the segment_user_id param in a segment_config object since it's only one param now.

tovbinm: good point!

ypc-faros: > lets put description here explaining that if segment_user_id the reporting is disabled and link to docs where we explain what is being tracked

I'll update the description once we have these docs.

tovbinm:
@ypc-faros you need to bump the version of the destination to release it

PR #317, opened 2022-03-02
Update dependencies and bump all versions

cjwooo:
## Description

Bumping individual source/destination package versions leads to future complications with our setup, so let's just bump all versions using `npm run bump` from the repository root folder like we used to do.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


tovbinm:
Thank you @cjwooo !!!

tovbinm (approved):

tovbinm:
Yep, let's keep doing `npm run bump` - better.

PR #322, opened 2022-03-04
Datadog application mapping

willmarks:
## Description

Adding logic to correctly handle the application mapping configuration input for Datadog.

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/faros-destination/src/converters/datadog/common.ts
```diff
@@ -11,10 +11,8 @@ export enum IncidentSeverityCategory {
   Custom = 'Custom',
 }
 
-type ApplicationMapping = Record<string, {name: string; platform?: string}>;
-
 interface DatadogConfig {
-  application_mapping?: ApplicationMapping;
+  application_mapping?: string;
```
willmarks, line 15: So this was super weird, but the Record type that I had borrowed from another converter was not correctly handling the application mapping input. Not sure how the other converts are working (they might not be). But to get this to work I needed to parse the string -> JSON myself.

destinations/faros-destination/src/converters/datadog/incidents.ts
```diff
@@ -72,6 +74,34 @@ export class DatadogIncidents extends DatadogConverter {
       });
     }
 
+    let applicationMapping;
+    try {
+      applicationMapping = JSON.parse(config.application_mapping);
+    } catch (err: any) {
+      throw new VError('Could not parse application mapping');
```
willmarks, line 81: Is this the correct way to throw an error within a converter?

destinations/faros-destination/src/converters/datadog/common.ts
tovbinm: I think there is a way to set a dictionary type in source spec.

tovbinm: Where else did you see this?

willmarks: the other two converters that have application_mappings: victorops and pagerduty

tovbinm: they all use `string` type in spec:
```
  "application_mapping": {
                        "type": "string",
                        "title": "Application Mapping",
                        "description": "JSON map of StatusPage service(s) name, to compute platform specific app name and platform name. Used to reference compute_Application object, from an ims_Incident object.",
                        "multiline": true,
                        "default": "{}",
                        "examples": [
                          "{ \"Aion\": { \"name\": \"aion\", \"platform\": \"ECS\" } }"
                        ]
                      }
```

tovbinm: They all use `parseObjectConfig` helper function which deals with it

destinations/faros-destination/src/converters/datadog/incidents.ts
tovbinm: Yes, it ok to throw an error we catch it here - https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/faros-destination/src/index.ts#L480.

But you dont need it if you use `parseObjectConfig` which already does it.

tovbinm (approved):

tovbinm:
dont forget to bump the version `npm run bump`

PR #324, opened 2022-03-08
Avoid failing on undefined converter + add tests for converter registry

tovbinm:
## Description

Avoid failing on undefined converter + add tests for converter registry

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change


cjwooo:
@tovbinm Pull latest commits from main. Backlog source was added and version needs to be bumped.

PR #331, opened 2022-03-11
Fix use of memoize

cjwooo:
## Description

The typescript-memoize library does not work for generators. Only the first call was yielding results. Therefore, I converted all the functions that utilize the `@Memoize` annotation to return normal lists.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -198,11 +198,11 @@
 			}
 		},
 		"node_modules/@aws-sdk/abort-controller": {
-			"version": "3.53.0",
-			"resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-3.53.0.tgz",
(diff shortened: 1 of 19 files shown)


tovbinm (approved):
bummer that it does not work with AsyncGenerators, oh well.

cjwooo:
Each stream is tested independently so there were no cases yet where the memoized function was called more than once.

tovbinm:
Can we add a test an check that it's being called exactly once? so we could test for memoize being used

Roma-Kyrnis:
I'll try to check. As I understand we need to check that function called exactly only and this function is in AsyncGenerator func?

PR #333, opened 2022-03-11
Fix converter-by-name lookup

cjwooo:
## Description

- Converter class names should only be the name of the stream itself, without the strictly capitalized "source prefix", e.g. `Issues` instead of `GithubIssues`
- Base Converter class has an additional required property: `source`, that takes the place of the strictly capitalized prefix that is now removed. This string should follow the casing of the actual source system, e.g. "GitHub" not "Github".

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/faros-destination/src/converters/converter-registry.ts
+++ b/destinations/faros-destination/src/converters/converter-registry.ts
@@ -1,4 +1,4 @@
-import {camelCase, upperFirst} from 'lodash';
+import {camelCase, find, findKey} from 'lodash';
 import {Dictionary} from 'ts-essentials';
 import {VError} from 'verror';
 
@@ -39,7 +39,17 @@ export class ConverterRegistry {
       const mod = require(`./${streamName.source}/${streamName.name}`);
 
       // Create converter instance by name
-      const className = upperFirst(camelCase(name));
+      const className =
+        find(
+          Object.keys(mod),
+          (c: string) =>
+            c.toLowerCase() === camelCase(streamName.name).toLowerCase()
+        ) ?? findKey(mod, (f: any) => typeof f === 'function');
+      if (!className) {
+        throw new VError(
+          `Could not find converter from module for stream ${name}`
+        );
+      }
       const converter = new mod[className]();
 
       // Keep the converter instance in the registry
--- a/destinations/faros-destination/src/converters/github/common.ts
+++ b/destinations/faros-destination/src/converters/github/common.ts
@@ -3,7 +3,7 @@ import {Utils} from 'faros-feeds-sdk';
 import {toLower} from 'lodash';
 import {Dictionary} from 'ts-essentials';
 
-import {Converter, DestinationRecord, StreamName} from '../converter';
+import {Converter, DestinationRecord} from '../converter';
 
 /** Common functions shares across GitHub converters */
 export class GithubCommon {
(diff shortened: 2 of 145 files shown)


cjwooo:
cc @Roma-Kyrnis @huongtn

tovbinm (commented):
@cjwooo can you please justify this change?

cjwooo:
This fixes https://github.com/faros-ai/airbyte-connectors/pull/271#discussion_r806412007. Plus we don't want developers who are creating custom converters for in-house sources to be tripped up by our current capitalization requirements ("Github" instead of "GitHub", "Googlecalendar" instead of "GoogleCalendar", etc)

tovbinm:
1. `"Github" instead of "GitHub"` would not be compatible with existing feeds and Events CLI (we are using `GitHub` everywhere). There will be probably more instances like this.
2. Not that important, but removing source from the class name (e.g. `Teams` instead of `GitHubTeams`) would potentially make it more difficult to troubleshoot

cjwooo:
This PR removes the requirement of having to name things with "Github", i.e. the wrong casing.

tovbinm:
OK, I see.

destinations/faros-destination/src/converters/converter-registry.ts
```diff
@@ -39,7 +39,17 @@ export class ConverterRegistry {
       const mod = require(`./${streamName.source}/${streamName.name}`);
 
       // Create converter instance by name
-      const className = upperFirst(camelCase(name));
+      const className =
+        find(
```
tovbinm, line 43: `find` is probably O(n). can we do case insensitive lookups quicker?

destinations/faros-destination/src/converters/github/common.ts
```diff
@@ -153,16 +153,12 @@ export class GithubCommon {
 
 /** Github converter base */
 export abstract class GithubConverter extends Converter {
+  source = 'GitHub';
```
tovbinm, line 156: I guess we'd better have `abstract class XYZConverter extends Converter` defined for each `XYZ` source

destinations/faros-destination/src/converters/googlecalendar/common.ts
```diff
@@ -131,6 +131,8 @@ export class GooglecalendarCommon {
 
 /** GoogleCalendar converter base */
 export abstract class GooglecalendarConverter extends Converter {
```
tovbinm, line 133: nit: can we maintain proper capitalization in name and rename `GooglecalendarConverter` to `GoogleCalendarConverter`? and the same for others, e.g AgileAccelerator

destinations/faros-destination/test/converter-registry.test.ts
```diff
@@ -40,9 +40,6 @@ describe('converter registry', () => {
       if (converter && converter.convert) {
         const streamName = converter.streamName.asString.toLowerCase();
 
-        // Skip Okta Faros converter aliases
```
tovbinm, line 43: oh nice!

tovbinm (approved):
LGTM 

Great stuff!!

chalenge (approved):
🙇🏽

PR #335, opened 2022-03-15
Fix Squadcast

Roma-Kyrnis:
## Description

Fix 400 error returned from Squacast because owner_id must be provided. Fix next iteration on events

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change


Roma-Kyrnis:
> Can the connector still pull incidents and events for all owners if an owner id is not provided?

Sure. I've added thing that pull all teams or users (if teams empty) if there is no owner_id

PR #370, opened 2022-04-06
changes for fetch max user and groups in single call and also remove …

prakashcc:
## Type of change
Bug Fix- Performance Issue

Below are the fixes -
1.) Currently, API is not using its full potential and only fetching 100 users and groups which is the default. this is 9 times in comparison with current
2.) inside the getUser function it is making multiple calls to fetch the same user data which I have combined and now getting in a single call. This will make this function from log (2n) to log(n). 

## Migration notes

## Extra info

### DIFF
--- a/sources/azureactivedirectory-source/src/azureactivedirectory.ts
+++ b/sources/azureactivedirectory-source/src/azureactivedirectory.ts
@@ -105,7 +105,7 @@ export class AzureActiveDirectory {
 
   async *getUsers(): AsyncGenerator<User> {
     const res = await this.httpClient.get<UserResponse>(
-      'users/?$select=Department,postalCode,createdDateTime,identities,streetAddress&$top=999'
+      'users/?$select=department,postalCode,createdDateTime,identities,streetAddress&$top=999'
     );
     for (const item of res.data.value) {
       try {
--- a/sources/azureactivedirectory-source/src/azureactivedirectory.ts
+++ b/sources/azureactivedirectory-source/src/azureactivedirectory.ts
@@ -104,9 +104,19 @@ export class AzureActiveDirectory {
   }
 
   async *getUsers(): AsyncGenerator<User> {
-    const res = await this.httpClient.get<UserResponse>(
-      'users/?$select=department,postalCode,createdDateTime,identities,streetAddress&$top=999'
-    );
+    const maxResults = 999;
+    const res = await this.httpClient.get<UserResponse>('users', {
+      params: {
+        $select: [
+          'department',
+          'postalCode',
+          'createdDateTime',
+          'identities',
+          'streetAddress',
+        ],
+        $top: maxResults,
+      },
+    });
     for (const item of res.data.value) {
       try {
         const managerItem = await this.httpClient.get<User>(
@@ -123,7 +133,12 @@ export class AzureActiveDirectory {
   }
 
   async *getGroups(): AsyncGenerator<Group> {
-    const res = await this.httpClient.get<GroupResponse>('groups/?$top=999');
+    const maxResults = 999;
+    const res = await this.httpClient.get<GroupResponse>('groups', {
+      params: {
+        $top: maxResults,
+      },
+    });
     for (const item of res.data.value) {
       const memberItems = await this.httpClient.get<UserResponse>(
         `groups/${item.id}/members`
--- a/sources/azureactivedirectory-source/src/azureactivedirectory.ts
+++ b/sources/azureactivedirectory-source/src/azureactivedirectory.ts
@@ -118,7 +118,6 @@ export class AzureActiveDirectory {
       } catch (error) {
         this.logger.error(error.toString());
       }
-      console.log(item);
       yield item;
     }
   }
--- a/sources/azureactivedirectory-source/src/azureactivedirectory.ts
+++ b/sources/azureactivedirectory-source/src/azureactivedirectory.ts
@@ -103,9 +103,37 @@ export class AzureActiveDirectory {
     }
   }
 
+  private async *paginate<T>(path: string, param = {}): AsyncGenerator<T> {
+    let after = null;
+    do {
+      try {
+        const res = await this.httpClient.get<T[]>(path, param);
+        const linkHeader = res.data['@odata.nextLink'];
+        if (linkHeader) {
+          after = linkHeader.split('$skiptoken=').pop();
+          param['params']['$skiptoken'] = after;
+        } else {
+          after = null;
+        }
+        const result = res.data['value'];
+        for (const item of result) {
+          yield item;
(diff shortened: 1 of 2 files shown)


sources/azureactivedirectory-source/src/azureactivedirectory.ts
```diff
@@ -128,12 +118,13 @@ export class AzureActiveDirectory {
       } catch (error) {
         this.logger.error(error.toString());
       }
+      console.log(item);
```
tovbinm, line 121: remove the console log

sources/azureactivedirectory-source/src/azureactivedirectory.ts
```diff
@@ -104,20 +104,10 @@ export class AzureActiveDirectory {
   }
 
   async *getUsers(): AsyncGenerator<User> {
-    const res = await this.httpClient.get<UserResponse>('users');
+    const res = await this.httpClient.get<UserResponse>(
+      'users/?$select=Department,postalCode,createdDateTime,identities,streetAddress&$top=999'
```
tovbinm, line 108: why aren't we paginating?

sources/azureactivedirectory-source/src/azureactivedirectory.ts
prakashcc: i removed it

sources/azureactivedirectory-source/src/azureactivedirectory.ts
prakashcc: I am going to add pagination next but these fixes will still be relevant after adding the pagination.

tovbinm: I would make the page size configurable with a sane default. Would 999 be a good one? or it's safer to go with a smaller value, eg. 250, 500?

prakashcc: this size is recommended by the azure-ad team and I have tested it with 999 users working fine. As soon as I will implement the pagination we will make it as custom properties like we have client_id etc.

tovbinm: @huongtn FYI

tovbinm: 1. I would like to have you add pagination 
2. I think there is a better way to pass query arguments than this (see params - https://masteringjs.io/tutorials/axios/get)

prakashcc: @tovbinm based on your suggestion. changes are completed.

tovbinm (approved):
just one comment left, otherwise lgtm!

cjwooo:
Please also fix the failing test https://github.com/faros-ai/airbyte-connectors/runs/5994598081?check_suite_focus=true#step:6:65

sources/azureactivedirectory-source/src/azureactivedirectory.ts
```diff
@@ -103,59 +103,90 @@ export class AzureActiveDirectory {
...
       }
-      yield item;
+      yield user;
     }
   }
 
   async *getGroups(): AsyncGenerator<Group> {
-    const res = await this.httpClient.get<GroupResponse>('groups');
-    for (const item of res.data.value) {
+    const maxResults = 999;
```
tovbinm, line 163: `async *getGroups(maxResults = 999)` same above

sources/azureactivedirectory-source/src/azureactivedirectory.ts
```diff
@@ -103,59 +103,90 @@ export class AzureActiveDirectory {
...
-        item.streetAddress = extraUserInfo.data.streetAddress;
-        item.identities = extraUserInfo.data.identities;
+  private async *paginate<T>(path: string, param = {}): AsyncGenerator<T> {
+    let after = null;
+    do {
+      try {
+        const res = await this.httpClient.get<T[]>(path, param);
+        const linkHeader = res.data['@odata.nextLink'];
+        if (linkHeader) {
+          after = linkHeader.split('$skiptoken=').pop();
```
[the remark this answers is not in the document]
tovbinm: @prakashcc FYI ^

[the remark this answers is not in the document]
prakashcc: @tovbinm I tried to parse with the format-link-header but it is not working as per expectation. I am getting an empty object in return.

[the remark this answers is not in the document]
tovbinm: can you share an example of `linkHeader` value?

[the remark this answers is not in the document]
prakashcc: this is the "@odata.nextLink": "https://graph.microsoft.com/v1.0/users?$top=5&$skiptoken=X%274453707 ... 6633B900000000000000000000%27"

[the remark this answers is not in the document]
tovbinm: How about using the `URL`?
```
> new URL('https://graph.microsoft.com/v1.0/users?$top=5&$skiptoken=X%274453707 ... 6633B900000000000000000000%27')?.searchParams.get('$skiptoken')
"X'4453707 ... 6633B900000000000000000000'"
```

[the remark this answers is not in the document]
prakashcc: @tovbinm The suggested solution is working but as per my understanding, the split and pop are more memory optimized.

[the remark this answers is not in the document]
prakashcc: But I have made the changes based on your recommendation.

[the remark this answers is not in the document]
tovbinm: Memory is not an issue. I would like to have pagination work even when there an extra url argument / different order is returned. That's why URL parsing is preferred.

sources/azureactivedirectory-source/src/azureactivedirectory.ts
```diff
@@ -1,5 +1,6 @@
 import axios, {AxiosInstance} from 'axios';
 import {AirbyteLogger, wrapApiError} from 'faros-airbyte-cdk/lib';
+import formatLinkHeader from 'format-link-header';
```
tovbinm, line 3: you dont need `formatLinkHeader`

prakashcc: @tovbinm fixed it sorry

tovbinm (approved):

tovbinm:
test is failing, otherwise lgtm

PR #374, opened 2022-04-08
Remove repoSourceFallback

willmarks:
## Description

If we are not going to allow for a configurable repo source fallback, it should just default to `VCS`.


tovbinm (approved):

tovbinm:
version bump needed

PR #386, opened 2022-04-11
Update deps

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


.github/workflows/release.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       HUSKY: 0
     steps:
       - name: Check out
-        uses: actions/checkout@v2.4.0
+        uses: actions/checkout@v3.0.0
```
tovbinm, line 22: I think we can simply do `actions/checkout@v3` which would pick the latest v3 everytime, no?

.github/workflows/release.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         uses: actions/checkout@v2.4.0
 
       - name: Docker login
-        uses: docker/login-action@v1.14.0
+        uses: docker/login-action@v1.14.1
```
tovbinm, line 87: same here

tovbinm (approved):

PR #389, opened 2022-04-11
actions/checkout@v2

tovbinm:
## Description

Use major version only

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -46,7 +46,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -46,7 +46,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2.4.0
+      uses: actions/checkout@v2
```
willmarks, line 49: is `v3` available now?

willmarks (approved):

.github/workflows/codeql-analysis.yml
cjwooo: Yes

PR #391, opened 2022-04-12
Destination: Create application from VictorOps incident

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
+++ b/destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
@@ -164,7 +164,6 @@ export class Incidents extends VictorOpsConverter {
         : incident.service;
     let application = {name: service, platform: ''};
 
-    // if we have an app mapping specified, only then write the application and impact records
     if (service && service in appMapping && appMapping[service].name) {
       this.logger.info(
         `Incident ${incident.incidentNumber} impacted service: ${service}`
--- a/destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
+++ b/destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
@@ -162,6 +162,7 @@ export class Incidents extends VictorOpsConverter {
       appField in incident && typeof incident[appField] === 'string'
         ? incident[appField]
         : incident.service;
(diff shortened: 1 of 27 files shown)


destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
```diff
@@ -162,30 +162,31 @@ export class Incidents extends VictorOpsConverter {
       appField in incident && typeof incident[appField] === 'string'
         ? incident[appField]
         : incident.service;
+    let application = {name: service, platform: ''};
 
     // if we have an app mapping specified, only then write the application and impact records
```
willmarks, line 167: Comment no longer applies.

destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
```diff
@@ -162,30 +162,31 @@ export class Incidents extends VictorOpsConverter {
...
+      record: application,
+    });
+
+    res.push({
+      model: 'ims_IncidentApplicationImpact',
+      record: {
+        incident: incidentRef,
+        application,
+      },
+    });
```
willmarks, line 189: Interesting. I was under the impression that we only cared about applications in IMS when there was an explicit mapping. If we always care now, we will need to make a this change to all the other IMS connectors I think. Why do we want to do this?

willmarks (approved):

destinations/airbyte-faros-destination/src/converters/victorops/incidents.ts
cjwooo: Consistency for IMS-related sources https://faros-ai.atlassian.net/browse/FAI-1751. Pagerduty and Statuspage converters already create the application when there's no explicit mapping.

willmarks: ah ok sounds good. I will make the same change for Datadog.

willmarks (approved):
Datadog looks good!

PR #398, opened 2022-04-19
FAI-1803: Add ServiceNow source/converter

willmarks:
## Description

Add source and converter for pulling Incidents and Users from ServiceNow.

## Notes

- I am still trying to find a way to use anything other than basic auth for API access...
- There does not appear to be any notion of a service linked to an incident (as far as I can tell...)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


cjwooo:
@willmarks  Do you have a config.json we can use for source acceptance testing? If so, can you send it to me

willmarks:
I do not have one for the acceptance tests as the instance I was using for testing is a developer instance that hibernates whenever not in use and will be reclaimed after not being used for 10 days. Any tests pointed to it would start failing as soon as the instance hibernates.

sources/servicenow-source/src/streams/incidents.ts
```diff
@@ -0,0 +1,59 @@
...
+    protected readonly logger: AirbyteLogger
+  ) {
+    super(logger);
+  }
+
+  getJsonSchema(): Dictionary<any, string> {
+    return require('../../resources/schemas/incident.json');
+  }
+  get primaryKey(): StreamKey {
+    return ['number.value'];
```
cjwooo, line 28: Change to `['number', 'value']` here and throughout (catalog jsons)

sources/servicenow-source/src/streams/incidents.ts
```diff
@@ -0,0 +1,59 @@
...
+  }
+  get cursorField(): string | string[] {
+    return ['sys_updated_on.value'];
```
cjwooo, line 31: Change to `['sys_updated_on', 'value']` here and throughout

sources/servicenow-source/acceptance-test-config.yml
```diff
@@ -0,0 +1,25 @@
...
+      configured_catalog_path: "test_files/full_configured_catalog.json"
+  full_refresh:
+    - config_path: "test_files/config.json"
+      configured_catalog_path: "test_files/full_configured_catalog.json"
+  incremental:
+    - config_path: "test_files/config.json"
+      configured_catalog_path: "test_files/incremental_configured_catalog.json"
+      future_state_path: "test_files/abnormal_state.json"
+      cursor_paths:
+        incidents: [ "sys_updated_on.value" ]
```
cjwooo, line 24: I know we're not running these tests for real but the cursor paths here should be the key in the resulting state object, not the cursor field specified by the stream in the catalog. This enables the test container to read the correct state key.

sources/servicenow-source/resources/spec.json
```diff
@@ -0,0 +1,49 @@
...
+      "url"
+    ],
+    "additionalProperties": false,
+    "properties": {
+      "username": {
+        "order": 0,
+        "type": "string",
+        "title": "ServiceNow Username",
+        "description": "The users should have the roles: sn_incident_read, cmdb_query_builder_read",
+        "airbyte_secret": true
```
cjwooo, line 19: Does the username need to be secret?

sources/servicenow-source/src/streams/incidents.ts
cjwooo: This is how to represent nested fields in the resulting stream catalog

cjwooo (approved):

sources/servicenow-source/acceptance-test-config.yml
willmarks: Ahh. ok that makes sense, thanks!

tovbinm:
@willmarks please add it to the README + catalogs

Issue #396, opened 2022-04-19
Add CircleCI connector / converters


sasher:

tovbinm:
Original implementation - https://github.com/faros-ai/feeds/tree/master/feeds/cicd/circleci-feed

PR #402, opened 2022-04-21
Add CircleCI connector

huongtn:
## Description

#396 

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


willmarks:
When you have a chance, please resolve the conflict. Thanks!

sources/circleci-source/resources/spec.json
```diff
@@ -0,0 +1,49 @@
...
+        "title": "repository name",
+        "description": "CircleCI repository name"
+      },
+      "cutoff_days": {
+        "type": "integer",
+        "title": "Cutoff Days",
+        "default": 90,
+        "description": "Only fetch data updated after cutoff"
+      },
+      "rejectUnauthorized": {
```
cjwooo, line 37: Why is this option necessary?

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -0,0 +1,163 @@
...
+    if (!config.token) {
+      throw new VError('No token provided');
+    }
+    if (!config.org_slug) {
+      throw new VError('No org_slug provided');
+    }
+    const parts = config.org_slug.split('/').filter((p) => p);
+    if (parts.length != 2) {
+      throw new VError(`Organization slug %s does not match the expected format' +
+      ' {vcs_slug}/{org_name}, e.g gh/my-org. We will try to query the' +
```
cjwooo, line 39: If we're throwing an error on misformatted org_slug, why say `we will try to query anyways`? We should just enforce proper formatting and not make any assumptions (which it looks like you're doing, but the error message indicates otherwise).

cjwooo:
Can multiple repo names be associated with a single org-slug?

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -0,0 +1,163 @@
...
+      config.org_slug,
+      config.repo_name,
+      startDate
+    );
+  }
+
+  projectSlug(): string {
+    return encodeURIComponent(`${this.orgSlug}/${this.repoName}`);
+  }
+  async checkConnection(): Promise<void> {
```
cjwooo, line 72: NIT: put newlines between functions

cjwooo:
Can we support multiple org slugs and repositories? In doing so, we need to track states per org-slug/repository for applicable incremental streams. Example stream that does this: https://github.com/faros-ai/airbyte-connectors/blob/main/sources/bitbucket-source/src/streams/commits.ts

cjwooo:
@huongtn Can you look at my latest comments

huongtn:
> @huongtn Can you look at my latest comments

ok, let me fix

sources/circleci-source/resources/spec.json
huongtn: I copied from Feed

sources/circleci-source/src/circleci/circleci.ts
huongtn: Ok, I updated the message

cjwooo:
Let's remove the org_slugs parameter and just have the user specify the repositories array as `orgSlug/repoName`

huongtn:
> @huongtn See my last comment, then I think we're good to ship.

ok, let me update

sources/circleci-source/resources/spec.json
```diff
@@ -0,0 +1,47 @@
...
+        "title": "Repository Names",
+        "description": "Names should be in the format \"project_type/organization_name/repository_name\""
+      },
+      "cutoff_days": {
+        "type": "integer",
+        "title": "Cutoff Days",
+        "default": 90,
+        "description": "Only fetch data updated after cutoff"
+      },
+      "reject_unauthorized": {
```
cjwooo, line 34: What is the purpose of this reject_unauthorized option? When would a user want to disable this?

cjwooo: @huongtn ^

huongtn: I copied from Feed

cjwooo (approved):

tovbinm:
Yay!!!

Issue #397, opened 2022-04-19
Add TravisCI  connector / converters


sasher:

tovbinm:
Original implementation - https://github.com/faros-ai/feeds/tree/master/feeds/cicd/travisci-feed

PR #423, opened 2022-05-03
Add TravisCI connector

huongtn:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ #397 ] New feature
- [ ] Breaking change

## Related issues

> Fix [#397 ]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/converters/travisci/common.ts
```diff
@@ -0,0 +1,82 @@
...
+        return {category: 'Queued', detail};
+      case 'passed':
+        return {category: 'Success', detail};
+      case 'failed':
+      case 'errored':
+        return {category: 'Failed', detail};
+      case 'canceled':
+        return {category: 'Canceled', detail};
+      default:
+        return {category: 'Custom', detail};
```
willmarks, line 56: Lets do `{..., detail: status}` so that we have the exact value of `status` instead of the lowercased `detail`. You can still switch on the lowercased detail.

PR #437, opened 2022-05-06
Bitbucket source fixes: part 1

cjwooo:
## Description

> Provide description here

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/bitbucket-source/src/bitbucket/bitbucket.ts
```diff
@@ -113,11 +106,8 @@ export class Bitbucket {
       ];
     }
 
-    if (!config.workspace) {
-      return [false, 'No workspace provided'];
-    }
-    if (!config.repositories) {
-      return [false, 'No repository provided'];
+    if (!config.workspaces) {
```
tovbinm, line 109: `!config.workspaces || config.workspaces.length < 1`

sources/bitbucket-source/src/bitbucket/bitbucket.ts
```diff
@@ -272,20 +292,25 @@ export class Bitbucket {
...
+        workspace,
         repoSlug
       );
     }
   }
 
-  @Memoize((repoSlug: string): string => repoSlug)
-  async getPipelines(repoSlug: string): Promise<ReadonlyArray<Pipeline>> {
+  @Memoize((workspace: string, repoSlug: string): string =>
+    workspace.concat(repoSlug)
```
tovbinm, line 302: `${workspace}${repoSlug}`

sources/bitbucket-source/src/bitbucket/types.ts
```diff
@@ -10,7 +10,7 @@ export interface BitbucketConfig extends AirbyteConfig {
   readonly username?: string;
   readonly password?: string;
   readonly token?: string;
-  readonly workspace: string;
+  readonly workspaces: string[];
```
tovbinm, line 13: `workspaces: ReadOnlyArray<string>;`

sources/bitbucket-source/src/streams/branches.ts
```diff
@@ -32,8 +31,11 @@ export class Branches extends AirbyteStreamBase {
     cursorField?: string[],
     streamState?: Dictionary<any>
   ): AsyncGenerator<StreamSlice> {
-    for (const repository of this.repositories) {
-      yield {repository};
+    const bitbucket = Bitbucket.instance(this.config, this.logger);
+    for (const workspace of this.config.workspaces) {
+      for await (const repo of bitbucket.getRepositories(workspace)) {
+        yield {workspace, repository: repo.name};
+      }
```
tovbinm, line 38: should we parallelize here by workspace / repository using p-limit or it's not something recommended to do in Airbyte source?

tovbinm: (same in other streams)

sources/bitbucket-source/src/streams/commits.ts
```diff
@@ -36,8 +35,12 @@ export class Commits extends AirbyteStreamBase {
     cursorField?: string[],
     streamState?: Dictionary<any>
   ): AsyncGenerator<StreamSlice> {
-    for (const repository of this.repositories) {
-      yield {repository};
+    const bitbucket = Bitbucket.instance(this.config, this.logger);
+    for (const workspace of this.config.workspaces) {
+      for await (const repo of bitbucket.getRepositories(workspace)) {
+        console.log(workspace + repo.name);
```
tovbinm, line 41: remove `console.log`

sources/bitbucket-source/src/streams/issues.ts
```diff
@@ -36,8 +35,11 @@ export class Issues extends AirbyteStreamBase {
     cursorField?: string[],
     streamState?: Dictionary<any>
   ): AsyncGenerator<StreamSlice> {
-    for (const repository of this.repositories) {
-      yield {repository};
+    const bitbucket = Bitbucket.instance(this.config, this.logger);
+    for (const workspace of this.config.workspaces) {
+      for await (const repo of bitbucket.getRepositories(workspace)) {
+        yield {workspace, repository: repo.name};
```
tovbinm, line 41: why all these streams yield the same tuples - workspace and repo?

tovbinm: also should we be using repo.slug instead?

sources/bitbucket-source/src/streams/branches.ts
cjwooo: They don't make any recommendation one way or the other. If we do parallelize repositories, we would need to NOT use stream slices to divide the work, because stream slices are processed sequentially, and we need to fix the stream state handling to have separate states per stream per repository (this is what we do in our feeds and what I'm doing in the next PR).

cjwooo: It does somewhat go against the general recommendation of using stream slices, but technically that is an optional feature of the airbyte source cdk.

cjwooo: Nothing, to enable the state saving logic for incremental streams with stream slices. Also, the default configuration for Airbyte is to allocate 1 CPU for a connector sync. We'll need to tune this (or switch to Kubernetes(?))

tovbinm: Even with 1 CPU we should be able to invoke multiple API calls in parallel thanks for Node.js event loop ;)

cjwooo: I think we should make this feature in the base source.

PR #443, opened 2022-05-10
Bitbucket: Track stream state by repo and fix linter errors

cjwooo:
## Description

Also add optional repositories param.

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/bitbucket-source/src/bitbucket/bitbucket.ts
```diff
@@ -253,7 +253,7 @@ export class Bitbucket {
   }
 
   @Memoize(
-    (workspace: string, repoSlug: string): string => `${workspace}${repoSlug}`
+    (workspace: string, repoSlug: string): string => `${workspace};${repoSlug}`
```
tovbinm, line 256: why `;` ?

sources/bitbucket-source/resources/spec.json
```diff
@@ -45,6 +45,17 @@
...
+        "type": "array",
+        "items": {
+          "type": "string"
+        },
+        "title": "Repositories",
+        "description": "Names of your Bitbucket repositories in the format 'workspace/repo-slug'. If none are provided, data from all repositories for the specified workspaces will be pulled.",
+        "examples": [
+          "blaze-lib/blaze"
+        ]
+      },
```
tovbinm, line 58: what if both `workspaces` and `repositories` list are provided?

cjwooo: Workspaces is always required. Repositories is optional and will indicate only certain repository data (commits, PRs, etc) is to be pulled. Workspace data (users, repo metadata) is always pulled.

sources/bitbucket-source/src/bitbucket/bitbucket.ts
cjwooo: Just putting a separator in between the parameters. Following examples in the memoize library.

tovbinm (approved):

tovbinm:
I am getting annoyed by the constant version bumps. Perhaps we can automatically release from the main branch on every merge instead of bumping it in the code?

cjwooo:
We could try that.

PR #470, opened 2022-05-24
Fix missing uid in gitlab releases converter

ypc-faros:


tovbinm (approved):

destinations/airbyte-faros-destination/src/converters/gitlab/releases.ts
```diff
@@ -71,6 +71,7 @@ export class Releases extends GitlabConverter {
           name: release.tag_name,
           repository: {
             name: repo?.toLowerCase(),
+            uid: repo?.toLowerCase(),
```
willmarks, line 74: Why is this needed? Isn't a vcs_Repository's key constructed from 'name' and 'organization'?

tovbinm: `uid` needed in the Faros CE schema

PR #502, opened 2022-06-08
Wrap and print full error in destination runner + tests

tovbinm:
## Description
Wrap and print full error in destination runner + tests

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/faros-airbyte-cdk/src/destinations/destination-runner.ts
+++ b/faros-airbyte-cdk/src/destinations/destination-runner.ts
@@ -1,6 +1,7 @@
 import {Command} from 'commander';
 import path from 'path';
 
+import {wrapApiError} from '../errors';
 import {AirbyteLogger} from '../logger';
 import {PACKAGE_VERSION, redactConfig} from '../utils';
 import {AirbyteDestination} from './destination';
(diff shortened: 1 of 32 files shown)


faros-airbyte-cdk/src/destinations/destination-runner.ts
```diff
@@ -82,10 +83,10 @@ export class AirbyteDestinationRunner {
               this.logger.write(message);
             }
           } catch (e: any) {
+            const w = wrapApiError(e);
+            const s = JSON.stringify(w);
             this.logger.error(
-              `Encountered an error while writing to destination: ${
-                e.message ?? JSON.stringify(e)
-              }`
+              `Encountered an error while writing to destination: ${w} - ${s}`
```
willmarks, line 89: does having `w` here only print the error message?

willmarks (approved):

faros-airbyte-cdk/src/destinations/destination-runner.ts
tovbinm: and we want the full error if possible.

PR #507, opened 2022-06-08
Bump dev deps

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "@types/jest": "^27.5.0",
     "@types/lodash": "^4.14.173",
     "@types/luxon": "^2.3.2",
-    "@types/node": "^17.0.31",
+    "@types/node": "^14.14.20",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "@types/jest": "^27.5.0",
     "@types/lodash": "^4.14.173",
     "@types/luxon": "^2.3.2",
-    "@types/node": "^17.0.31",
+    "@types/node": "^14.14.20",
```
tovbinm, line 22: downgrade?!

cjwooo: Yes we should be the major version that corresponds to the minimum node version we support, 14, to avoid using features that aren't available in version 14.x

PR #516, opened 2022-06-09
Only write unique apps in all converters for Faros destination

tovbinm:
## Description

Only write unique apps in all converters for Faros destination

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change


willmarks (approved):

willmarks:
We will probably need this same change for all our ims sources

PR #522, opened 2022-06-13
Fix PagerDuty spec

danieldiamond:
## Description

Cutoff days and page_size fields in spec currently require `integer`, however when you change these default values in the airbyte source spec, it fails the schema check as airbyte receives `string` from the UI.

Changing the spec to expect string values and converting where applicable in the `incident` and `logentry` classes.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/package.json
+++ b/destinations/airbyte-faros-destination/package.json
@@ -1,6 +1,6 @@
 {
   "name": "airbyte-faros-destination",
-  "version": "0.1.93",
+  "version": "0.1.94",
   "description": "Faros Destination for Airbyte",
   "keywords": [
     "airbyte",
@@ -36,7 +36,7 @@
     "analytics-node": "^6.0.0",
     "axios": "^0.26.0",
     "date-format": "^4.0.6",
-    "faros-airbyte-cdk": "^0.1.93",
+    "faros-airbyte-cdk": "^0.1.94",
     "faros-feeds-sdk": "^0.10.0",
     "fs-extra": "^10.0.0",
     "git-url-parse": "^11.6.0",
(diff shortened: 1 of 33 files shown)


cjwooo:
@danieldiamond I'm unable to reproduce the error behavior you described. Can you attach a video or gif of the Airbyte UI when you change the fields.

danieldiamond:
@cjwooo you might have simply run `Retest source` and not `Save changes and test`

https://user-images.githubusercontent.com/33811744/173460204-c682bca3-3700-44c5-b250-93ba606be146.mov

tovbinm:
@danieldiamond which Airbyte version are you using?

danieldiamond:
@tovbinm `0.39.13-alpha`

tovbinm:
OK, this explains it. We are on `0.35.5-alpha`.

@cjwooo any ideas on how can we make it work for both versions?

danieldiamond:
@tovbinm isn't my PR backward compatible??

danieldiamond:
@cjwooo can you verify that my solution works for your airbyte version

cjwooo:
@danieldiamond Are you using the version of pagerduty source that comes by default with Airbyte, version 0.1.23? If so, update it to the latest version, 0.1.93. The Airbyte UI unfortunately still does not handle `numbers` in JSON schemas properly, so we updated all of our source specs to specify `integers` instead of `numbers`. However, the source version that comes with Airbyte does not have that fix. We will contribute a patch to the Airbyte repo to update the versions of our sources to the latest ones.

danieldiamond:
oh thats annoying `0.1.97` doesn't show up in the airbyte UI as "this isnt the latest version"

danieldiamond:
closing this. latest version works

PR #526, opened 2022-06-14
Test api errors thrown up

willmarks:
## Description

Add explicit tests to confirm that errors are being thrown out of readRecords.

Probably not worth bumping the version for this.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/servicenow-source/test/index.test.ts
+++ b/sources/servicenow-source/test/index.test.ts
@@ -118,6 +118,22 @@ describe('index', () => {
     expect(items).toStrictEqual(incidents);
   });
 
+  test('streams - incidents, throws up API error', async () => {
+    const expectedMessage = 'API Error';
+    listIncidents.mockRejectedValue(new VError(expectedMessage));
+    const source = new sut.ServiceNowSource(logger);
+    const streams = source.streams({});
+    const stream = streams[0];
+    try {
+      const iter = stream.readRecords(SyncMode.FULL_REFRESH);
+      await iter.next();
+    } catch (err: any) {
+      expect(err.message).toBe(expectedMessage);
+      return;
+    }
+    expect(0).toBe(1);
+  });
+
   test('streams - users, use incremental sync mode', async () => {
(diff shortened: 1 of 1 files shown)


sources/servicenow-source/test/index.test.ts
```diff
@@ -118,6 +118,22 @@ describe('index', () => {
...
+    const streams = source.streams({});
+    const stream = streams[0];
+    try {
+      const iter = stream.readRecords(SyncMode.FULL_REFRESH);
+      await iter.next();
+    } catch (err: any) {
+      expect(err.message).toBe(expectedMessage);
+      return;
+    }
+    expect(0).toBe(1);
```
tovbinm, line 134: remove `expect(0).toBe(1);`

sources/servicenow-source/test/index.test.ts
```diff
@@ -153,4 +169,20 @@ describe('index', () => {
...
+    const streams = source.streams({});
+    const stream = streams[1];
+    try {
+      const iter = stream.readRecords(SyncMode.FULL_REFRESH);
+      await iter.next();
+    } catch (err: any) {
+      expect(err.message).toBe(expectedMessage);
+      return;
+    }
+    expect(0).toBe(1);
```
tovbinm, line 186: same

sources/servicenow-source/test/index.test.ts
willmarks: If an error is not thrown/caught then the test passes when it should fail.

tovbinm: I see, then do 
```
expect(() => {
      const iter = stream.readRecords(SyncMode.FULL_REFRESH);
      await iter.next();
      fail('Error should have been thrown.');
}).toThrow(....)
```

tovbinm: fail also looks ok.

tovbinm (approved):

PR #529, opened 2022-06-15
Adds Faros Feeds converter

ypc-faros:
## Description

Adds Faros Feeds converter

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


ypc-faros:
> why is the logger private? where is it being used?

It is used by the `handleRecordProcessingError`. The HasuraWriter is instantiated in `FarosDestination`:

```
const writer = new HasuraWriter(
          hasura,
          origin,
          stats,
          this.handleRecordProcessingError
        );
```

and the `handleRecordProcessingError` implementation references `this.logger`. Since there's no logger in HasuraWriter, it throws with `Cannot read property 'error' of undefined `

tovbinm:
Hmm, this is weird. Checking on that...

tovbinm (approved):

PR #542, opened 2022-06-20
Upgrade to Node.js 16

tovbinm:
## Description

Upgrade to Node.js 16

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,10 +44,10 @@ jobs:
       - name: Set up Node
         uses: actions/setup-node@v3
         with:
-          node-version: 14
+          node-version: 16
 
       - name: Install dependencies
-        run: npm install -g npm@7 lerna && lerna bootstrap --hoist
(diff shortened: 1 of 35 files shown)


cjwooo (changes requested):
Each source/destination package.json should be updated to indicate the use of version 16 https://github.com/faros-ai/airbyte-connectors/blob/main/sources/example-source/package.json#L18

tovbinm:
Technically, you can still use node 14 to run it. We just package it with Node 16 now.

cjwooo:
If we want to keep supporting 14.x then we should revert this change to the @types/node version https://github.com/faros-ai/airbyte-connectors/pull/542/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R21

tovbinm:
ok, let me bump then

tovbinm:
bumped. we should release it under `0.2.0`

cjwooo (approved):

Issue #528, opened 2022-06-15
Add source for Linear


tovbinm:
https://linear.app/

tovbinm:
Apparently there is an official source for Linear already - https://github.com/linear/linear-airbyte-source

I think we should remove our own source implementation and just use the keep the converters for Faros Destination.

thomas-gerber:
Our source seems better than the community one (incremental, labels, ...). We need to figure out the way forward.

thomas-gerber:
@cjwooo suggested the following course of action:
* remove the source from https://github.com/faros-ai/airbyte-connectors/pull/547
* fork the community source to make important streams incremental, and submit our changes upstream
* have the converter support that instead
@cjwooo is creating the fork.

igorsechyn:
hey @thomas-gerber, 

i am developer at linear and was working on the source implementation mentioned here. did not realise there is someone else working on it already. 

one important thing to know, is that connector developed and maintained by linear is using a separate export api, which is only available on plus paid plans. 

We do plan to support incremental updates, but it is not the highest priority right now. Also we do support `labels` under `issueLabel` stream.

Having said that, we are always happy about external contributions. Let us know, if we can assist.

PR #547, opened 2022-06-24
Add Linear connector

huongtn:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ #528 ] New feature
- [ ] Breaking change

## Related issues

> Fix [#528]() 

## Migration notes

## Extra info


thomas-gerber:
There now is a airbyte linear community source: https://linear.app/docs/airbyte
See conversation here: https://github.com/faros-ai/airbyte-connectors/issues/528. @tovbinm suggested removing our source and just keeping our converters.

thomas-gerber:
Although, it seems the community source does not support incremental yet...

thomas-gerber:
so, maybe we keep our source until we are satisfied with the community source?

cjwooo:
@huongtn Linear already has an Airbyte source for pulling their data. We want to use their version instead of maintaining our own version, but it is missing a couple features that you implemented in your source implementation. The missing features I found so far are: "Labels" stream, and incremental support for "Issues" stream.

We've made a fork of Linear's Airbyte source here https://github.com/faros-ai/linear-airbyte-source. Please make a PR adding those missing features (and any other missing features I missed) to that version of the source, and update your destination converters to support Linear's source. I suspect only minimal changes to the converters will be needed, as the fetched data from each stream should be largely the same.

We can then contribute your source additions upstream to the official Linear source repo.

PR #554, opened 2022-06-30
OpsGenie: paginate users

willmarks:
## Description

Users were not being paginated but should be. Note: I also checked Teams, the API does not allow pagination.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/opsgenie-source/src/opsgenie/opsgenie.ts
+++ b/sources/opsgenie-source/src/opsgenie/opsgenie.ts
@@ -110,7 +110,7 @@ export class OpsGenie {
         if (new Date(incident.createdAt) >= startTimeMax) {
           const incidentItem = incident;
           const timeLineResponse =
-            await this.restClient.get<IncidentTimeLinePaginateResponse>(
+            await this.retryApi<IncidentTimeLinePaginateResponse>(
               `v2/incident-timelines/${incident.id}/entries`
             );
           if (timeLineResponse.status === 200)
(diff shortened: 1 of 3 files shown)


sources/opsgenie-source/src/opsgenie/opsgenie.ts
```diff
@@ -88,16 +93,18 @@ export class OpsGenie {
...
-
     const serviceMap = new Map<string, string>();
 
+    let offset = 0;
     do {
+      const params = {offset, limit: this.pageSize, sort: 'createdAt'};
       const response = await this.retryApi<PaginateResponse<Incident>>(
-        `v1/incidents?sort=createdAt&order=asc&limit=${this.pageSize}&offset=${offset}`
+        `v1/incidents`,
+        params
```
willmarks, line 107: `order=asc` is the default.

sources/opsgenie-source/src/opsgenie/opsgenie.ts
```diff
@@ -88,22 +93,24 @@ export class OpsGenie {
...
       );
       for (const incident of response?.data?.data ?? []) {
         if (new Date(incident.createdAt) >= startTimeMax) {
           const incidentItem = incident;
           const timeLineResponse =
-            await this.restClient.get<IncidentTimeLinePaginateResponse>(
+            await this.retryApi<IncidentTimeLinePaginateResponse>(
```
willmarks, line 113: Not sure why this helper for rate limiting backoff was not being used for everything... but now I am using it everywhere.

cjwooo (approved):

PR #569, opened 2022-07-07
[datadog-source] Add support for fetching metrics via the /api/v1/query endpoint

iMacTia:
## Description

* [datadog-source] Add support for fetching metrics via the /api/v1/query endpoint

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues
N/A

## Migration notes
N/A

## Extra info

This is my first time working with Airbyte and the Faros CDK, so comments/suggestions are welcome!
The biggest challenge I had building this connector is that the `api/v1/query` endpoint does not map directly to a specific resource like `users` or `incidents`. So I had to get creative in how to allow users to specify the metrics they'd like to fetch and keep a state for each of them.

### DIFF
--- a/sources/datadog-source/src/datadog.ts
+++ b/sources/datadog-source/src/datadog.ts
@@ -142,7 +142,7 @@ export class Datadog {
   // Retrieve the specified metric between from and to unix timestamps
   async *getMetrics(
     query: string,
-    query_hash: string,
+    queryHash: string,
     from: number,
     to: number
   ): AsyncGenerator<MetricPoint, any, any> {
@@ -155,7 +155,7 @@ export class Datadog {
       for (const metadata of res.series) {
         for (const point of metadata.pointlist) {
           yield {
-            id: `${query_hash}-${metadata.metric}-${point[0]}`,
+            id: `${queryHash}-${metadata.metric}-${point[0]}`,
             displayName: metadata.displayName,
             metric: metadata.metric,
             timestamp: point[0],
--- a/sources/datadog-source/src/datadog.ts
+++ b/sources/datadog-source/src/datadog.ts
@@ -1,4 +1,4 @@
-import {v2} from '@datadog/datadog-api-client';
+import {v1, v2} from '@datadog/datadog-api-client';
 import {AirbyteLogger} from 'faros-airbyte-cdk';
 import VError from 'verror';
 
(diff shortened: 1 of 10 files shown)


sources/datadog-source/src/streams/metrics.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  async *readRecords(
+    syncMode: SyncMode,
+    _cursorField?: string[],
+    _streamSlice?: Dictionary<any, string>,
+    streamState?: Dictionary<string, number>
+  ): AsyncGenerator<Dictionary<any, string>, any, unknown> {
+    for (const metric of this.datadog.config.metrics ?? []) {
+      let from = 0;
+      const query_hash = createHash('md5').update(metric.query).digest('hex');
```
tovbinm, line 48: nit: rename `query_hash` to `queryHash`, and `max_to` to `maxTo`

sources/datadog-source/src/datadog.ts
```diff
@@ -112,6 +139,39 @@ export class Datadog {
...
+  ): AsyncGenerator<MetricPoint, any, any> {
+    try {
+      const res = await this.client.metrics.queryMetrics({
+        from,
+        to,
+        query: query,
+      });
+      for (const metadata of res.series) {
+        for (const point of metadata.pointlist) {
+          yield {
```
tovbinm, line 157: Why instead of converting just return the Datadog API results as is?

Based on Airbyte philosophy, a source should be responsible to fetch and return data as is, and the conversion should be done later on, e.g in the destination if necessary.

sources/datadog-source/src/streams/metrics.ts
iMacTia: Sorry! That's my Ruby habit 😂
I'll get those renamed to follow TS conventions 👍

sources/datadog-source/src/datadog.ts
iMacTia: So this is the biggest issue I had conceptually, and what I tried to explain in the PR description.
The way resources are managed in sources works pretty well for "RESTful" resources: you call an endpoint that gives you access to a collection of resources. You get those resources based on a timestamp so that any new/updated resource is returned, and you add them to the "flow".

With Datadog metrics I had the issue where I'm not really accessing a resource but instead, I'm querying metrics.
There's one endpoint (`/query`) that lets you query the metrics and returns a time series graph ([timestamp + value] pairs), rather than the actual resource. To make things harder, you can pass parameters to your query, effectively creating an infinite number of "views" (one for each "combination" of filters).

The way I decided to tackle this is to create a fictitious "MetricPoint" resource that encapsulates a given value at a given point in time for a given metric (e.g. "value for metric `system.cpu.idle` at 2 pm today is 10").
The `query` (the combination of parameters) is used with the metric and the timestamp to create an ID so that requesting the same "time series" will not result in duplicates.

As I said, this was quite different from what things are normally done for other sources, but I'm very much open to changing the implementation following your direction. If you could point me to a similar source that deals with such a scenario, that would be really helpful 🙏

tovbinm: Thanks for explaining the rationale. This makes sense. Let's go with your solution for now.

tovbinm (approved):

tovbinm:
Released under v0.2.11 - https://github.com/faros-ai/airbyte-connectors/releases/tag/v0.2.11

tovbinm:
Thank you for the contribution! @iMacTia

iMacTia:
Thank you for the speedy and thorough review!

PR #577, opened 2022-07-13
[datadog-source] Fix source configuration

iMacTia:
## Description

Airbyte UI does not support arrays of objects as configuration parameters for sources.
This change makes the Datadog source `metrics` config an array of strings instead of an array of objects.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

This is a follow-up fix from #569 

## Extra info

When we tried the new stream introduced in #569 on the Airbyte UI, we got the following error:

> Internal Server Error: malformed JsonSchema array type, must have items field in {"item":{"type":"object","properties":{"query":"string"},"additionalProperties":false},"type":"array","order":3,"title":"Metrics","description":"list of metrics to fetch and their configuration"}

### DIFF
--- a/sources/datadog-source/resources/spec.json
+++ b/sources/datadog-source/resources/spec.json
@@ -36,11 +36,7 @@
         "title": "Metrics",
         "description": "list of metrics to fetch and their configuration",
         "item": {
-          "type": "object",
-          "additionalProperties": false,
-          "properties": {
-            "query": "string"
-          }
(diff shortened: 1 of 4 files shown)


tovbinm (approved):

tovbinm:
The simpler the better ;)

iMacTia:
Once again, thank you for the speedy review!

PR #579, opened 2022-07-14
Fix lerna --scope option.

iMacTia:
## Description

This change allows us to pass the `-- --scope ...` option when running `npm run test` or `npm run test-cov` from the project root (as stated in the README).
Before this change, the additional options configured in the global `package.json` would cause the `--scope` option to be forwarded to `jest` instead, causing an error.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

Fixes #567


tovbinm:
Thank you!

PR #585, opened 2022-07-21
FAI-2713  - ServiceNow configurable cmdb_ci/cmdb_ci_service identifier

willmarks:
## Description

Support a configurable service identifier so that any fields can be used to populate the `compute_Application.name`

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/servicenow-source/resources/spec.json
+++ b/sources/servicenow-source/resources/spec.json
@@ -29,22 +29,29 @@
         "title": "ServiceNow url",
         "description": "ServiceNow instance url"
       },
-      "cutoff_days": {
+      "service_id_field": {
         "order": 3,
+        "type": "string",
+        "title": "ServiceNow service identifier field",
+        "description": "The cmdb_ci or cmdb_ci_service field to use as the service identifier",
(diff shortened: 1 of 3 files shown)


sources/servicenow-source/resources/spec.json
```diff
@@ -29,22 +29,29 @@
...
+      "cutoff_days": {
+        "order": 4,
         "type": "integer",
         "title": "Cutoff Days",
         "default": 90,
         "description": "Only fetch data updated after cutoff"
       },
       "page_size": {
-        "order": 4,
+        "order": 5,
```
tovbinm, line 47: why change the order and not add `service_id_field` at the end? (or it is not a breaking change anyways?)

tovbinm (approved):

sources/servicenow-source/resources/spec.json
willmarks: Everything is referenced by key so this is not breaking. This just changes where the field shows up when configuring a source in airbyte.

PR #586, opened 2022-07-25
FAI-1920: Added support for Faros GraphQL V2 in Faros Destination

ted-faros:
## Description

This PR introduces support for using Faros's v2 graphql APIs in the cloud edition of the faros airbyte destination.  

To this end, we've generalized the `HasuraClient` and `HasuraWriter` formerly used exclusively by CE into a `GraphQLClient` and `GraphQLWriter`.  

While generalizing `HasuraClient` we factored it into 3 parts:

- `SchemaLoader`: interface providing a Faros-defined graphql `Schema`.
- `Backend`: interface providing standard GraphQL API for queries and mutations.
- `GraphQLClient`: uses `Schema` to translate records into mutations and writes them to `Backend`

With this refactoring, `GraphQLClient` can now be used directly against Hasura (the CE use case) or with Faros via the `FarosClient` (the cloud v2 use case).

_Note_: the Hasura schema loading logic of `HasuraClient` has been factored into `HasuraSchemaLoader` and moved to `feeds-sdk` repo.  This permits the logic to be reused in new `/graphql/schema` endpoint of Faros.

Also in this PR is new feature for automatic application of defaults to `AirbyteConfig`.  With addition of `withDefaults` utility, defaults supplied in `spec.json` are substituted for undefined values.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

[clio #432](https://github.com/faros-ai/clio/pull/432)
[feeds-sdk #296](https://github.com/faros-ai/feeds-sdk/pull/296)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -355,6 +400,8 @@ export class FarosDestination extends AirbyteDestination {
           stats,
           writer
         );
+        // flush any buffered writes
+        await graphQLClient.flush();
```
tovbinm, line 404: we should flush the writer, not the client. Let's move it there.

ted-faros:
With batch mutations, the Hasura-based graphql engine write latency is comparable to Clio revisions.

Here is Clio revisions:

```
{"log":{"level":"INFO","message":"Opening a new revision on graph ted with expiration of 5 seconds"},"type":"LOG"}
{"log":{"level":"INFO","message":"Processed 335 records"},"type":"LOG"}
bin/main write --config $TEST_DIR/dest_cloud_config.json --catalog  <   0.56s user 0.10s system 25% cpu 2.600 total
```

Here is v2 with batch size of 1 (aka no batching):

```
{"log":{"level":"INFO","message":"Using GraphQLClient for write with batch size 1"},"type":"LOG"}
{"log":{"level":"INFO","message":"Processed 335 records"},"type":"LOG"}
bin/main write --config $TEST_DIR/dest_cloud_v2_config.json --catalog  <   1.07s user 0.23s system 21% cpu 5.951 total
```

And v2 with batch size of 100:

```
{"log":{"level":"INFO","message":"Using GraphQLClient for write with batch size 100"},"type":"LOG"}
{"log":{"level":"INFO","message":"Processed 335 records"},"type":"LOG"}
bin/main write --config $TEST_DIR/dest_cloud_v2_config.json --catalog  <   0.60s user 0.11s system 42% cpu 1.681 total
```

destinations/airbyte-faros-destination/package.json
```diff
@@ -37,7 +37,7 @@
     "axios": "^0.26.0",
     "date-format": "^4.0.6",
     "faros-airbyte-cdk": "0.0.1",
-    "faros-feeds-sdk": "^0.12.0",
+    "faros-feeds-sdk": "0.13.0",
```
tovbinm, line 40: we should be ok to keep doing `^0.13.0`

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -53,6 +53,12 @@
                 "title": "Segment User Id",
                 "description": "The User UUID with which to track events in Segment. If not present, then reporting is disabled. See https://community.faros.ai/docs/telemetry for more details.",
                 "format": "uuid"
+              },
+              "graphql_mutation_batch_size": {
+                "type": "integer",
+                "title": "GraphQL mutation batch size",
+                "description": "Maximum number of mutations to execute in a single request to GraphQL backend",
+                "default": 1
```
tovbinm, line 61: was this the original default? shall we increase this to a larger value?

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -91,6 +97,24 @@
                 "description": "The Faros graph name.",
                 "default": "default"
               },
+              "use_graphql_v2": {
```
tovbinm, line 100: let's instead of the boolean do an enum for the GraphQL API version with a drop down for values: v1, v2

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -91,6 +97,24 @@
...
+                "type": "boolean",
+                "title": "GraphQL Version 2",
+                "description": "Use version 2 of graphql engine.",
+                "default": false
+              },
+              "graphql_mutation_batch_size": {
+                "type": "integer",
+                "title": "GraphQL mutation batch size",
+                "description": "Maximum number of mutations to execute in a single request to GraphQL backend.",
```
tovbinm, line 109: `GraphQL API`

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -91,6 +97,24 @@
...
+                "default": 100
+              },
+              "check_tenant": {
```
tovbinm, line 112: what is this for? if this is only used with tests, no need to expose it in the spec.json?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -295,14 +181,88 @@ export class HasuraClient {
...
+    const gql = GraphQLClient.batchMutation(queries);
+    if (gql) {
+      const res = await this.backend.postQuery(gql);
+      if (res.errors) {
+        this.logger.warn(
+          `Error while saving batch: ${JSON.stringify(
+            res.errors
+          )}. Query: ${gql}`
+        );
+        // now try mutations individually and fail on the first bad one
```
tovbinm, line 205: Dont we get enough information from the batch mutation call in `res.errors`? rerunning all the mutations so we can just isolate a single mutation seems nice to have, or do you think it's necessary?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -295,14 +181,88 @@ export class HasuraClient {
...
+   *   }
+   *
+   *  Notable here are the i1/i2 aliases.  These required when multiple operations
+   *  share the same name (e.g. insert_cicd_Artifact_one) and are supported in
+   *  jsonToGraphQLQuery with __aliasFor directive.
+   *
+   *  @return batch gql mutation or undefined if the input is undefined, empty
+   *  or doesn't contain any mutations.
+   */
+  static batchMutation(queries: any[]): string | undefined {
```
tovbinm, line 248: how large a mutation can be? is there a server / client side limit?

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -711,14 +765,12 @@ export class FarosDestination extends AirbyteDestination {
...
       let isTimestamped = false;
-      if (writer) {
-        if (writer instanceof HasuraWriter) {
-          isTimestamped = await writer.write(result);
-        } else {
-          const obj: Dictionary<any> = {};
-          obj[result.model] = result.record;
-          writer.write(obj);
-        }
+      if (writer instanceof GraphQLWriter) {
```
tovbinm, line 768: just confirming that `writer instanceof GraphQLWriter` would return false on null / undefined / invalid type?

destinations/airbyte-faros-destination/test/index.test.ts
```diff
@@ -160,3 +162,175 @@ describe('index', () => {
...
+            id: true,
+          },
+        },
+      },
+    ];
+    expect(GraphQLClient.batchMutation(json)).toBeUndefined();
+  });
+});
+
+describe('utils withDefaults', () => {
```
tovbinm, line 294: do we have file for utils tests? if not, let's add one and move this there

tovbinm (commented):
looks good, see comments

destinations/airbyte-faros-destination/src/destination.ts
ted-faros: yes and fall into the `else if (writer)` test.  My eye reads the diff as equivalent.  If I'm missing something, please tell me.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: This is a nice to have for operators.  My experience in isolating errors that occur in batches is you need to set the batch size to 1 and rerun.  This would give the operator a precise errors message without the rerun.

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: This allows us to turn off the check without altering the code.  Can I use an alternate method for toggling this?  Commenting/uncommenting code is a hassle/liability.

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: Prior to this PR there was no batching.  I just verify batching works against CE's Hasura instance; I'll set this to 100 for CE.

destinations/airbyte-faros-destination/package.json
ted-faros: fixed

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: I haven't found anything in the docs. I think we'll need to test to determine the practical limit.

destinations/airbyte-faros-destination/resources/spec.json
tovbinm: When do you need to toggle this? In tests / local runs from ui or?

destinations/airbyte-faros-destination/src/destination.ts
tovbinm: Yes, `writer` is null then dry run is enabled. Please revert this change.

destinations/airbyte-faros-destination/test/index.test.ts
ted-faros: moved to `faros-airbyte-cdk` package

destinations/airbyte-faros-destination/src/destination.ts
ted-faros: reverted

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: fixed

tovbinm:
@cjwooo please have a look as well

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: local runs.  without the toggle you need to run posidone for the /tenant api.

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: fixed

destinations/airbyte-faros-destination/resources/spec.json
tovbinm: more generically, we'd rather have a way to skip connection check, because this is essentially what this is for Faros API. wdyt? @cjwooo

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -91,6 +97,28 @@
...
+              "cloud_graphql_batch_size": {
+                "type": "integer",
+                "title": "GraphQL mutation batch size",
+                "description": "Maximum number of mutations to execute in a single request to GraphQL backend.",
+                "default": 100
+              },
+              "check_connection": {
+                "type": "boolean",
+                "title": "Check Faros API Connection",
+                "description": "Check Faro API connection on init.",
```
ypc-faros, line 119: nit: Faro => Faros

ypc-faros (approved):
lgtm

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -180,7 +180,7 @@ export class FarosDestination extends AirbyteDestination {
     } catch (e) {
       throw new VError(`Failed to initialize Faros Client. Error: ${e}`);
     }
-    if (!config.dry_run && config.edition_configs.check_tenant) {
+    if (!config.dry_run && config.edition_configs.check_connection) {
```
tovbinm, line 183: @ted-faros 
Let apply the same check prior to all remote calls invoked from `init(..)` and `initX(..)` functions
1. `await this.getFarosClient().tenant()`
2. `await this.getGraphQLClient().healthCheck()`
3. etc.

tovbinm: (updated ^ )

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: good eye @ypc-faros!  fixed

destinations/airbyte-faros-destination/src/destination.ts
tovbinm: Hmm, I am not sure. You added one in an incorrect spot. Since you added the `check_connection` flag in Cloud section of the spec, only add it in the initCloud path, where we call the remote APIs

ted-faros: good catch. fixed

PR #607, opened 2022-08-03
ServiceNow: Correct destination model list

willmarks:
## Description

Add missing destination models. @cjwooo this is needed to delete all written models during `full_refresh` mode, right?

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


tovbinm (approved):

tovbinm:
You might need to fix the tests

willmarks:
The tests already were considering these models. We should probably add deletion model tests across the board.

PR #608, opened 2022-08-03
Allow overriding spec in Faros Destination runner + adding config check handler

tovbinm:
## Description

Allow overriding spec in Faros Destination runner + adding config check handler

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -76,6 +77,8 @@ export class FarosDestination extends AirbyteDestination {
     super();
   }
 
+  public onConfigCheck: (config: AirbyteConfig) => Promise<void> = undefined;
```
cjwooo, line 80: Don't need `public`

PR #628, opened 2022-08-11
FAI-2934 - Upgrade Datadog client to 1.2.0

willmarks:
## Description

Upgrade the Datadog API client to 1.2.0. There was some slight changes to the client's creation that needed to be updated. Also annoyingly they did not document correctly how to set unstable operations until yesterday.

**QUESTION:**
Are the changes in the package-lock.json correct? seems like a lot is changing...

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "name": "root",
       "dependencies": {
-        "@datadog/datadog-api-client": "1.0.0-beta.8",
+        "@datadog/datadog-api-client": "1.2.0",
         "@gitbeaker/node": "^35.6.0",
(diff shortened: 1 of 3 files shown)


package-lock.json
```diff
@@ -7677,6 +7678,22 @@
         "node >=0.6.0"
       ]
     },
+    "node_modules/faros-airbyte-cdk": {
```
tovbinm, line 7681: `faros-airbyte-cdk` should not be in this file (same below)

tovbinm (commented):
fix package-lock.json please

tovbinm (approved):

PR #632, opened 2022-08-16
one more check

tovbinm:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/destination.ts
+++ b/destinations/airbyte-faros-destination/src/destination.ts
@@ -180,18 +180,20 @@ export class FarosDestination extends AirbyteDestination {
     } catch (e) {
       throw new VError(`Failed to initialize Faros Client. Error: ${e}`);
     }
-    if (!config.dry_run && config.edition_configs.check_connection) {
-      try {
+    try {
+      if (!config.dry_run && config.edition_configs.check_connection) {
         await this.getFarosClient().tenant();
-      } catch (e) {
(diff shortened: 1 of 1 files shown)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -180,18 +180,20 @@ export class FarosDestination extends AirbyteDestination {
...
+    } catch (e) {
+      throw new VError(`Invalid Faros API url or API key. Error: ${e}`);
     }
     this.farosGraph = config.edition_configs.graph;
     try {
-      const exists = await this.getFarosClient().graphExists(this.farosGraph);
-      if (!exists) {
-        throw new VError(`Faros graph ${this.farosGraph} does not exist`);
+      if (!config.dry_run && config.edition_configs.check_connection) {
+        const exists = await this.getFarosClient().graphExists(this.farosGraph);
```
ted-faros, line 193: In the mode I'm running with Clio local, skipping this will only delay the graph check until the first graphql operation.

tovbinm: sure, but it sill makes sense to me to disable any remote checks if `config.edition_configs.check_connection` is set to false. because if config.dry_run is true, we wont write any entries.

PR #655, opened 2022-08-23
FAI-3003 - Remove exact address

willmarks:
## Description

For privacy reasons, we no longer want to save the users' addresses.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


nat-casey (approved):

cjwooo:
Let's remove zip code as well.

tovbinm (commented):
Should we have a param to control this?

PR #659, opened 2022-08-25
FAI-3026 - Retrieve active status from employmentHistoryStatus

willmarks:
## Description

This PR allows for the configuration of the custom 'employmentHistoryStatus' value that is equivalent to 'inactive' that should be used to determine if a user is inactive in Faros. If this configuration is not provided, the converter will fall back to doing what it was doing before: looking at the 'status' field.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/bamboohr/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/bamboohr/common.ts
@@ -4,6 +4,7 @@ import {Converter, StreamContext} from '../converter';
 
 export interface BambooHRConfig {
   bootstrap_teams_from_managers?: boolean;
+  inactive_employment_history_status?: string;
 }
 
 /** BambooHR converter base */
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/converters/bamboohr/common.ts
```diff
@@ -21,4 +22,8 @@ export abstract class BambooHRConverter extends Converter {
   protected bootstrapTeamsFromManagers(ctx: StreamContext): boolean {
     return this.bamboohrConfig(ctx).bootstrap_teams_from_managers ?? false;
   }
+
+  protected inactiveEmploymentHistoryStatus(ctx: StreamContext): string {
```
cjwooo, line 26: nit: `string | undefined`

cjwooo (approved):

PR #660, opened 2022-08-25
FAI-3026 - BambooHR support array of employmentHistoryStatus values

willmarks:
## Description

Support more than one employmentHistoryStatus value that should be equated to inactive within Faros.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -674,7 +674,7 @@
                       "inactive_employment_history_status": {
                         "type": "string",
                         "title": "Inactive employmentHistoryStatus",
-                        "description": "The configured value for the employeeHistoryStatus that is equivalent to an employee being inactive"
+                        "description": "Comma separated array of employmentHistoryStatus values that are equivalent to an employee being inactive"
```
ypc-faros, line 677: Can this be an actual array of strings instead?

willmarks: It could be and I had it that way at first, but that seemed more annoying for customers to express:
```
["One","Two"]
```
vs
```
One,Two
```
The first requires an understanding of how a JSON array is created. I felt the second way was easier for anyone to write.. does that make sense? or do you think the first way is worth it?

ypc-faros: I don't have a strong opinion on either, but I'd favor string array :)

Airbyte renders a nice UI element for string arrays. E.g., pipelines in the screenshot below:

Screen Shot 2022-08-26 at 12 29 03 PM

willmarks: Ah that is nice. I'll switch it

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -672,9 +672,9 @@
                         "default": false
                       },
                       "inactive_employment_history_status": {
-                        "type": "string",
+                        "type": "array",
```
ypc-faros, line 675: Also include `"items": {"type": "string"}` under `inactive_employment_history_status`

ypc-faros (approved):

destinations/airbyte-faros-destination/resources/spec.json
willmarks: oh good call thanks

PR #661, opened 2022-08-26
Fix error handling in batch operations

ypc-faros:
## Description

This PR fixes two issues:

1. The `errors` (if present) is a field under `response.data` instead of under `response`
2. We should always remove from the write buffer. We currently do not remove operations that fail in Hasura, so they are always kept around in the front of the buffer.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


ted-faros (approved):
Looks good.  Thanks for debugging this and making the change @ypc-faros !

ypc-faros:
> Looks good. Thanks for debugging this and making the change @ypc-faros !

Thank you @ted-faros !!!

PR #663, opened 2022-08-29
FAI-3028 Add uid to compute application model in ServiceNow

ypc-faros:
## Description

Add uid to compute application model in ServiceNow (for Faros Community Edition)

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/common/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/common/common.ts
@@ -1,5 +1,3 @@
-import {toLower} from 'lodash';
-
 /** Common functions shared across converters */
 export class Common {
   static computeApplication(
@@ -17,7 +15,7 @@ export class Common {
     name: string,
     platform?: string
   ): string {
-    if (!platform) return toLower(name);
-    return [toLower(name), toLower(platform)].join('_');
+    if (!platform) return name;
+    return [name, platform].join('_');
   }
 }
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/converters/common/common.ts
```diff
@@ -0,0 +1,23 @@
...
+      uid: Common.computeApplicationUid(name, platform),
+    };
+  }
+
+  private static computeApplicationUid(
+    name: string,
+    platform?: string
+  ): string {
+    if (!platform) return toLower(name);
+    return [toLower(name), toLower(platform)].join('_');
```
willmarks, line 21: This would definitely be an edge case, but could we potentially have an issue with uniqueness because the uid is case insensitive, but the key for compute_Application within SaaS is case sensitive?
Screen Shot 2022-08-29 at 10 46 43 AM

willmarks: Maybe this is an expected difference between CE and SaaS?

ypc-faros: Good point. I'm actually not totally sure what's the correct behavior here (and I haven't see enough field data to make a decision). Any thoughts @thomas-gerber ?

willmarks (approved):

PR #674, opened 2022-09-02
FAI-3038: More GraphQL V2 Schema Updates

ted-faros:
## Description

Standardize response across `GraphQLBackend` implementations by extracting and returning the `data` field from Axios response in `HasuraBackend`.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/community/hasura-backend.ts
+++ b/destinations/airbyte-faros-destination/src/community/hasura-backend.ts
@@ -20,8 +20,10 @@ export class HasuraBackend implements GraphQLBackend {
   }
 
   async postQuery(query: any): Promise<any> {
-    return await this.api.post('/v1/graphql', {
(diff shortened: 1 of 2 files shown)


destinations/airbyte-faros-destination/src/community/hasura-backend.ts
```diff
@@ -20,8 +20,10 @@ export class HasuraBackend implements GraphQLBackend {
   }
 
   async postQuery(query: any): Promise<any> {
```
tovbinm, line 22: seems like a breaking change. is there anything else that can be affected?

ted-faros: It was a bug.  It worked for CE but not Clio v2.  Now it works for both

tovbinm (approved):

PR #714, opened 2022-09-29
Add `uid` to `compute_Application` + use common function

tovbinm:
## Description

Add 'uid' to 'compute_Application' + use common function

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/azurepipeline/pipelines.ts
+++ b/destinations/airbyte-faros-destination/src/converters/azurepipeline/pipelines.ts
@@ -8,12 +8,14 @@ import {Pipeline} from './models';
 
 export class Pipelines extends AzurePipelineConverter {
   readonly destinationModels: ReadonlyArray<DestinationModel> = [
+    'compute_Application',
     'cicd_Deployment',
     'cicd_Organization',
     'cicd_Pipeline',
   ];
 
   private seenOrganizations = new Set<string>();
+  private seenApplications = new Set<string>();
 
   async convert(
     record: AirbyteRecord,
@@ -47,6 +49,11 @@ export class Pipelines extends AzurePipelineConverter {
           mappedApp?.name ?? runItem.name,
           mappedApp?.platform
         );
+        const appKey = application.uid;
+        if (!this.seenApplications.has(appKey)) {
+          res.push({model: 'compute_Application', record: application});
+          this.seenApplications.add(appKey);
+        }
       }
 
       const startedAt = Utils.toDate(runItem.createdDate);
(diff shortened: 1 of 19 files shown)


thomas-gerber (changes requested):
@tovbinm I think your are missing https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/bitbucket/deployments.ts#L51 / https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/bitbucket/common.ts#L16 (and possibly in other sources as well.

destinations/airbyte-faros-destination/src/converters/azurepipeline/pipelines.ts
```diff
@@ -36,10 +37,18 @@ export class Pipelines extends AzurePipelineConverter {
       });
     }
 
+    const applicationMapping = this.applicationMapping(ctx);
```
tovbinm, line 40: @thomas-gerber @cjwooo @willmarks should we be writing `compute_Application` entries from Azure pipelines converter? same question for Bitbucket deployments converter?

destinations/airbyte-faros-destination/src/converters/azurepipeline/pipelines.ts
```diff
@@ -36,10 +37,18 @@ export class Pipelines extends AzurePipelineConverter {
...
+
     for (const runItem of pipeline.runs) {
-      const applicationMapping = this.applicationMapping(ctx);
-      const application =
-        (applicationMapping && applicationMapping[runItem.name]) ?? null;
+      let application = null;
+      if (runItem?.name) {
+        const mappedApp = applicationMapping[runItem.name];
+        application = Common.computeApplication(
```
tovbinm, line 46: @thomas-gerber @cjwooo @willmarks should we be writing `compute_Application` entries from Azure `pipelines` converter? same question for Bitbucket `deployments` converter?

tovbinm: Right now we are only writing out the refs from the deployments (but not the `compute_Application` entries)

cjwooo: Yea we should.

thomas-gerber (approved):
Great! 
Also appreciated that applications are written for Bitbucket / Azure even if mappings are not provided.

destinations/airbyte-faros-destination/src/converters/azurepipeline/pipelines.ts
tovbinm: ok, I adding it.

PR #724, opened 2022-10-08
CircleCI: Fixes differences in pipeline and organization UIDs

thomas-gerber:
Fixes differences in pipeline and organization UIDs

## Type of change
- [x] Bug fix

## Migration notes

> Full resync will be necessary

## Extra info

> We always use the full project slug and common methods, because the Circle CI API documentation is incorrect on fields like organization_slug.


thomas-gerber:
@cjwooo I didn't find a simple way to add a test that loads [this](https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/test/resources/circleci/all-streams.log) and checks that the 2 streams compute the same cicd_Pipeline ID.

tovbinm (approved):

PR #735, opened 2022-10-19
GitLab: Harmonize project and board names with repo and pipeline name

thomas-gerber:
Important to properly filter "git" dashboard on one repository (including its issues).

Right now, repo/pipeline are e.g. `farosce-test` whereas board/project name are `FarosCE test` for https://gitlab.com/farosai/farosce-test

## Type of change
- [x] New feature
- [x] Breaking change

## Related issues
https://github.com/faros-ai/faros-community-edition/pull/231

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/gitlab/projects.ts
+++ b/destinations/airbyte-faros-destination/src/converters/gitlab/projects.ts
@@ -28,7 +28,7 @@ export class Projects extends GitlabConverter {
     res.push(
       ...GitlabCommon.tms_ProjectBoard_with_TaskBoard(
         {uid: `${project.id}`, source},
-        project.path?.toLowerCase(),
+        repository.name,
         project.body,
         project.created_at,
         project.updated_at
@@ -37,8 +37,8 @@ export class Projects extends GitlabConverter {
     res.push({
       model: 'cicd_Pipeline',
       record: {
-        uid: project.path?.toLowerCase(),
-        name: project.name,
+        uid: repository.uid,
+        name: repository.name,
         description: project.description?.substring(
           0,
           GitlabCommon.MAX_DESCRIPTION_LENGTH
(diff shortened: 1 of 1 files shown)


destinations/airbyte-faros-destination/src/converters/gitlab/projects.ts
```diff
@@ -28,7 +28,7 @@ export class Projects extends GitlabConverter {
     res.push(
       ...GitlabCommon.tms_ProjectBoard_with_TaskBoard(
         {uid: `${project.id}`, source},
-        project.name,
+        project.path?.toLowerCase(),
```
tovbinm, line 31: not good. project.path might include the group(s), since the projects on gitlab can be nested. try it out.

thomas-gerber: @tovbinm do you agree that vcs_Repository, cicd_Pipeline, tms_Project, tms_TaskBoard should have the same UID and name (i.e. there is a 1-1 mapping between all those)?

thomas-gerber: If so, I am thinking then using the parseRepositoryKey.uid everywhere (incl. cicd_Build name).
https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts#L85

thomas-gerber: I think that should take care of the nested case consistently.

tovbinm: I presume you are only referring to GitLab in this case.

UIDs should be typically constructed as from id of the source system if possible without modifications as its unique properties already guaranteed by the source system.

Names are more presentational and can vary, but generally should be recognizable by users.

BTW, Where does it help you to unify these?

thomas-gerber: In my latest commit (see below), I use consistently the `parseRepositoryKey` function in common. Its purpose was already to generate a UID and a name (set to the same thing presently).
As stated in the PR, unifying all of those allows the creation of dashboards where a filter can apply at the same time to PRs, builds, tasks.

tovbinm (approved):

PR #772, opened 2022-11-05
[Buildkite] Fix missing builds when large number of pipelines

thomas-gerber:
## Description
Use paginated GraphQL API to list pipelines in build stream

Currently, the builds stream (contrary to the pipelines stream) uses the REST API to get the list of pipelines to iterate over. That method does not paginate, which means that when one has more than 30 pipelines, builds will be missed.

We switch to the GraphQL API for listing pipelines, for which we paginate properly.

## Type of change
- [x] Bug fix

### DIFF
--- a/sources/buildkite-source/src/buildkite/buildkite.ts
+++ b/sources/buildkite-source/src/buildkite/buildkite.ts
@@ -343,13 +343,11 @@ export class Buildkite {
     cursor?: string,
     createdAtFrom?: Date
   ): AsyncGenerator<Build> {
-    const res = await this.restClient.get<PipelineSlug[]>(
-      `organizations/${organization}/pipelines`
-    );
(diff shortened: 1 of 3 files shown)


thomas-gerber:
@cjwooo could you help fix the test? I am not familiar with that library.

tovbinm:
@thomas-gerber you need to updated the mock here - https://github.com/faros-ai/airbyte-connectors/blob/main/sources/buildkite-source/test/index.test.ts#L147

thomas-gerber:
@cjwooo @tovbinm I properly wired the builds stream test. 
Not sure if I can easily test large number of pipelines as this would require crafting lots of data with cursor in `pipelines_input`. Do you think what test we have is acceptable?

tovbinm (approved):

PR #780, opened 2022-11-11
[Circle CI] makes pipeline stream state slice specific

thomas-gerber:
## Description

Current pipeline stream state is global, which can lead to missing pipelines during incremental syncs if one has more than one repo (aka slices).

## Type of change
- [x] Bug fix

### DIFF
--- a/sources/circleci-source/src/streams/pipelines.ts
+++ b/sources/circleci-source/src/streams/pipelines.ts
@@ -9,9 +9,8 @@ type StreamSlice = {
   repoName: string;
 };
 
-interface PipelineState {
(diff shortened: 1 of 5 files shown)


thomas-gerber:
@cjwooo do I have a way in tests to read the state post readRecords (the same way we read the records)?

thomas-gerber:
> @cjwooo do I have a way in tests to read the state post readRecords (the same way we read the records)?

thanks @cjwooo , followed your suggestion!

cjwooo (approved):

tovbinm:
I am thinking if we should do it in every of our sources...

PR #791, opened 2022-11-17
Set package version in Docker images + update Lerna to 6.0.3

tovbinm:
## Description

Package version wasn't set correctly for published Docker images. This change fixes it + updates Lerna to 6.0.3

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,19 +2,24 @@ FROM node:16-alpine
 
 WORKDIR /home/node/airbyte
 
-ARG version
-RUN test -n "$version" || (echo "'version' argument is not set, e.g --build-arg version=x.y.z" && false)
-ENV CONNECTOR_VERSION $version
-
 COPY lerna.json .tsconfig.json package.json package-lock.json ./
 RUN sed -i "/jest\|mockttp/d" package.json
 COPY ./faros-airbyte-cdk ./faros-airbyte-cdk
 COPY ./faros-airbyte-common ./faros-airbyte-common
 COPY ./sources ./sources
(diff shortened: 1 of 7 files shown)


cjwooo (changes requested):
Currently doesn't work with source acceptance tests and makes unnecessary changes to package-lock.json. I'll take ownership of this PR.

Dockerfile
```diff
@@ -8,9 +8,18 @@ COPY ./faros-airbyte-cdk ./faros-airbyte-cdk
...
+
 RUN apk add --no-cache --virtual .gyp python3 make g++ \
     && npm install -g lerna tsc
 RUN lerna bootstrap --hoist
+
+ARG version
+RUN test -n "$version" || (echo "'version' argument is not set, e.g --build-arg version=x.y.z" && false)
+ENV CONNECTOR_VERSION $version
+
+RUN cp package-lock.json .package-lock.json.tmp \
```
cjwooo, line 20: Couldn't find a lerna command option that maintains the package-lock formatting. I confirmed that the corrected version still applies.

cjwooo (approved):

Dockerfile
tovbinm: Does the `package-lock.json` contain any references to local packaged, e.g `faros-airbyte-cdk`? If so, they would keep pointing to `0.0.1`, no?

cjwooo: There's no mention of any package that is implemented in the repo.

PR #524, opened 2022-06-13
FAI-1803: ServiceNow client timeout

willmarks:
## Description

Timeout configuration.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

PR #793, opened 2022-11-17
FAI-3971: Batch mutations in Faros Destination 

ted-faros:
## Description

This PR improves the write throughput of the Faros destination with "smart" batching of upserts.  The high-level algorithm for this change is:

1. For each record, build a tree of `Upsert`s.  The tree has more than one node if the current record represents a nested entity (e.g. `branch` record referencing `repo` which, in-turn, references `org).`  
2. Buffer `Upsert`s and index each by model (e.g. `vcs_Branch`)
3. When batch size limit is reached, execute a single insert mutation per model.  Start with leaves of `Upsert` tree. 
4. After inserting a batch, copy the `id` of each record back to the `Upsert` object.  When inserting subsequent batches, required foreign keys will be read from the `Upsert` tree and copied to the appropriate batch mutation.   

_Note_: there is a complication in this algorithm for self-referent model (i.e. `org_Employee`'s manager relationship).  For these models, we split the batch into "levels".  The first level consists of records with no dependencies on records of the same type.  The second depends on the first and so on. 

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

[clio #524](https://github.com/faros-ai/clio/pull/524)
[hasura #35](https://github.com/faros-ai/hasura/pull/35)
[feeds-sdk #358](https://github.com/faros-ai/feeds-sdk/pull/358)


ted-faros:
fyi @tovbinm @ypc-faros I just found in a bug in this PR for self-referential models (i.e. ims_Team, org_Employee,  org_Team,  tms_Task, tms_Team and vcs_Team).  I should have a fix shortly.

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -55,9 +55,15 @@
                 "format": "uuid"
               },
               "community_graphql_batch_size": {
+                "type": "integer",
+                "title": "GraphQL upsert batch size",
+                "description": "Maximum number of records to execute in a single GraphQL upsert.",
+                "default": 1000
+              },
+              "community_graphql_mutation_batch_size": {
```
tovbinm, line 63: let's just keep the original batch size parameter with default of 1000

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,37 +34,118 @@ interface ConflictClause {
 
 export interface GraphQLBackend {
   healthCheck(): Promise<void>;
+
   postQuery(query: any): Promise<any>;
 }
 
 interface WriteOp {
-  query: string;
+  query: any;
```
tovbinm, line 42: add `readonly` to all fields that you arent planning to modify (in this and other interfaces)

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,37 +34,118 @@ interface ConflictClause {
...
+  id?: string; // will be set after object is inserted
+  model: string;
+  object: Dictionary<any>;
+  foreignKeys: Dictionary<Upsert>;
+}
+
+export class UpsertBuffer {
+  private readonly upsertBuffer: Map<string, Upsert[]> = new Map();
+
+  add(upsert: Upsert): void {
```
tovbinm, line 61:
```
const existing = this.upsertBuffer.get(upsert.model);
if (!existing) {
    this.upsertBuffer.set(upsert.model, [upsert]);
    return; 
}
existing.push(upsert);
```

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,37 +34,118 @@ interface ConflictClause {
...
+    const size = this.size();
+    this.upsertBuffer.clear();
+    return size;
+  }
+}
+
+export function serialize(obj: any): string {
+  return keys(obj)
+    .sort()
+    .map((k) => `${k}:${obj[k]}`)
```
tovbinm, line 97: this wont work for nested objects. 

also, isnt there already some function in lodash for this?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -122,12 +217,61 @@ export class GraphQLClient {
...
+        'failed to flush upsert buffer. abandoned %d records',
+        numRemoved
+      );
+    }
+  }
+
+  private async doFlushUpsertBuffer(): Promise<void> {
+    for (const model of this.tableDependencies) {
+      const op = this.toUpsertOp(model);
+      if (op) {
```
tovbinm, line 242: better: `if (!op) continue;` so you can avoid unnecessary nesting

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -188,12 +332,21 @@ export class GraphQLClient {
...
     }
   }
 
   async flush(): Promise<void> {
+    // sequentially flush upsert then write buffer
+    // we do this sequentially because we may have updates
+    // that operate on records in the upsert buffers and
+    // therefore need the upserts written first
+    await this.flushUpsertBuffer();
+    await this.flushWriteBuffer();
```
tovbinm, line 346: what's the difference between the two buffers?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -319,6 +472,108 @@ export class GraphQLClient {
...
+    const modelUpserts = this.upsertBuffer.get(model);
+    if (modelUpserts) {
+      const upsertsByKey = new Map();
+      // add FKs and index upsert by primary key
+      // the index is needed after write to find source
+      // upsert and set id
+      const all = modelUpserts.map((u) => {
+        const fullObj = this.objectWithForeignKeys(u);
+        const key = this.serializedPrimaryKey(model, fullObj);
+        if (!upsertsByKey.has(key)) {
```
tovbinm, line 542: same as https://github.com/faros-ai/airbyte-connectors/pull/793/files#r1029512603

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: There are 3 kinds of writes: upserts, updates and deletes.  We are doing the smart batching for upserts only.  The updates and deletes (which are much less common) cannot be batched like the upserts. They are buffered to the write buffer and unchanged by this PR.

ted-faros:
> fyi @tovbinm @ypc-faros I just found in a bug in this PR for self-referential models (i.e. ims_Team, org_Employee, org_Team, tms_Task, tms_Team and vcs_Team). I should have a fix shortly.

this is fixed and ready to review.

destinations/airbyte-faros-destination/resources/spec.json
ted-faros: There are 2 types of batching going on and they have very different characteristics.  The upsert batch size will be in the 1000s and the update/delete batch size (aka mutation batch size) will be in the 100s.  Some options to simplify: 

- drop the mutation batching completely
- hard-code the mutation batch size to 100

Either way, these operations are infrequent and shouldn't impact throughput materially.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: done

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: done

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,37 +34,118 @@ interface ConflictClause {
...
+  }
+
+  size(): number {
+    return Array.from(this.upsertBuffer.values()).reduce(
+      (sum, arr) => sum + arr.length,
+      0
+    );
+  }
+
+  get(model: string, remove = true): Upsert[] | undefined {
```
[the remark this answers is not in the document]
ted-faros: done

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: these will be always be simple objects.  I don't see a lodash equivalent.  Besides, the semantics of this method are crucial to the foreign key accounting.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: done

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,12 +37,165 @@ interface ConflictClause {
...
+}
+
+/**
+ * Converts an Upsert (which is a tree structure) into an array of tuples.
+ * Each tuple contains an Upsert and its level in the tree structure.
+ * The level indicates the number of steps from a leaf where leaves are level 0.
+ * If an Upsert has multiple paths to a leaf, the level is the maximum number of
+ * steps.
+ */
+function addLevel(u: Upsert, level = 0): [Upsert, number][] {
```
ypc-faros, line 186: Is the `level` param ever used?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,12 +37,165 @@ interface ConflictClause {
...
+  const result = [];
+  let maxAncestorLevel = 0;
+  for (const ancestor of Object.values(u.foreignKeys)) {
+    if (ancestor.model === u.model) {
+      const ancestorsWithLevels = addLevel(ancestor, level + 1);
+      maxAncestorLevel = max([ancestorsWithLevels[0][1], maxAncestorLevel]);
```
ypc-faros, line 192: Is `ancestorsWithLevels` guaranteed to be a single element here?

destinations/airbyte-faros-destination/resources/spec.json
tovbinm: Let's keep the original argument as is to avoid breaking changes and just add a new one then: `community_graphql_upsert_batch_size` with default of 1000 (same below)

destinations/airbyte-faros-destination/src/common/graphql-client.ts
tovbinm: if you only expect a simple object, then change from `obj: any` to `Dictionary<numeric | string | string[], string>` or something

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -22,12 +37,165 @@ interface ConflictClause {
...
+    async *[Symbol.asyncIterator](): AsyncIterator<T> {
+      for (const batch of batches) {
+        yield await callback(batch);
+      }
+    },
+  };
+}
+
+/**
+ * Separates upserts into levels. Level 0 has no dependencies on other levels.
```
tovbinm, line 161: It looks like you need a topological sort here. Why not use some existing library?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
tovbinm: great! it's worth adding this comment somewhere in the code. Your future self would appreciate this ;)

destinations/airbyte-faros-destination/test/index.test.ts
```diff
@@ -285,3 +298,405 @@ describe('graphql-client', () => {
...
+    expect(serialize({z: 1, a: 'bar'})).toEqual('a:bar|z:1');
+    expect(serialize({})).toEqual('');
+  });
+  test('strictPick', async () => {
+    expect(strictPick({z: 1, a: 'bar'}, ['z', 'b'])).toEqual({z: 1, b: 'null'});
+    expect(strictPick({z: 1, a: 'bar'}, ['b'])).toEqual({b: 'null'});
+  });
+});
+
+describe('toLevels', () => {
```
tovbinm, line 575: it might be worth to split this test file into several smaller ones

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: it is guaranteed to have at least one element here.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: good catch!  going to replace this with a library call per @tovbinm comment

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: I looked into this.  I could replace [sortByLevel](https://github.com/faros-ai/airbyte-connectors/blob/tt/destination-perf-fix/destinations/airbyte-faros-destination/src/common/graphql-client.ts#L186-L199) with a call to toposort library.  However, just building the input to the sort and mapping it back to the original data structure is a bunch of code.  The 12 line roll your own with [appropriate tests](https://github.com/faros-ai/airbyte-connectors/blob/tt/destination-perf-fix/destinations/airbyte-faros-destination/test/graphql-client.test.ts#L418) seems clearer to me.

ted-faros: You see, the nodes in this graph are `Upsert` objects and they don't have unique ids.  I'd have to assign UIDs and then map back and forth through the library.

ted-faros: @tovbinm thanks for the comment here.  I thought this through and realized a much simpler algorithm for these self-referential models is to run 2 queries per batch:

1.  insert all the nodes
2. insert all the edges (i.e. update the FKs)

I'll do this a remove a bunch of complexity

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ypc-faros: Why is it sufficient to compare with the level of the first element (`ancestorsWithLevels[0][1]`) though?

Shouldn't we compare with all `ancestorsWithLevels`'s levels ?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
tovbinm: Great! Not to complicate things too much, but how about self referencing types? e.g `org_Employee -> manager: org_Employee` ? https://github.com/faros-ai/canonical-models/blob/main/resources/models/org.gql#L51

ted-faros: That is exactly why there is this sorting logic.  There is a [branch here](https://github.com/faros-ai/airbyte-connectors/blob/tt/destination-perf-fix/destinations/airbyte-faros-destination/src/common/graphql-client.ts#L362) to deal with self-referent models.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: we only care about the adjacent ancestor.  The recursion takes care of ordering the ancestors depth-wise. We do the max because there could be more than one ancestor and one side of the ancestor tree could be deeper.  I realized now that ancestor is kinda backwards.  They are really dependents.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: Turns out the nodes and edges approach is simple conceptually but not easily implemented.  I'm going to stick with the current `toLevels` way of sequentially inserting a batch of possibly self-referent records.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: done

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: [done](https://github.com/faros-ai/airbyte-connectors/blob/tt/destination-perf-fix/destinations/airbyte-faros-destination/src/common/graphql-client.ts#L201-L220)

ypc-faros (approved):

ypc-faros:
This is awesome @ted-faros 🔥 🔥 🔥

tovbinm:
🎉🎉🎉

PR #795, opened 2022-11-18
Better logging for sources

tovbinm:
## Description

1. Removed logging on every stream slice checkpoint operation (this should make logging less noisy)
2. Added logging on start & end of each stream slice with number of read records

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/faros-airbyte-cdk/src/sources/source-base.ts
+++ b/faros-airbyte-cdk/src/sources/source-base.ts
@@ -230,6 +230,11 @@ export abstract class AirbyteSourceBase<
         recordCounter
       );
     }
+    this.logger.info(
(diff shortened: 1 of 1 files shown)


cjwooo:
Let's also add a log statement indicating the start of processing a stream state.

tovbinm:
Yeah, we already log the initial stream state in the beginning of "readIncremental" function.

cjwooo:
Sorry I mean processing a stream slice, not state. It's good to know when the connector has finished a slice and started the next one.

tovbinm:
@cjwooo I updated the PR and description. please have a look again

cjwooo (approved):

PR #797, opened 2022-11-21
Modify the records read from the graph + add converter

ypc-faros:
## Description

Introduces a `result_model` field to the spec to control how to return the records read from the graph.

Example:

For the query:

```
query MyQuery {
  vcs {
    pullRequests {
      nodes {
        number
      }
    }
  }
}
```

We can return the records as:
```
"vcs": {
      "pullRequests": {
        "nodes": [
                {ORIGINAL_RECORD}
        ]
      }
    }
```
when `result_model` is `Nested`

or as:
```
"vcs_PullRequest": {ORIGINAL_RECORD}
```
when `result_model` is `Flat`

The main goal is to be able to control how to output the records in the most appropriate format for piping through a jsonata converter (`Nested`) or using the Faros graph converter (`Flat`).

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/converters/faros-graph/faros_graph.ts
```diff
@@ -0,0 +1,32 @@
...
+import {
+  Converter,
+  DestinationModel,
+  DestinationRecord,
+  StreamContext,
+} from '../converter';
+import {FarosFeed as OriginalFarosFeed} from '../faros-feeds/faros_feed';
+
+/**
+ * This converter is identical to FarosFeed converter.
```
tovbinm, line 12: a nice hack! :)

sources/faros-graphql-source/resources/spec.json
```diff
@@ -58,6 +58,17 @@
         "title": "Page Size",
         "description": "Page size to use when paginating through query results.",
         "default": 100
+      },
+      "path_to_record": {
```
tovbinm, line 62: how about this instead?
```
"result_model": {
  "order": 6,
  "type": "string",
  "title": "Result model",
  "description": "Where to place each of the query results in their corresponding output records. E.g Nested - { vcs { pullRequests { nodes: [<record>]}}}, Flat - { vcs_PullRequest: <record> }",
  "default": "Nested",
  "enum": [
    "Nested",
    "Flat"
  ]
}
```

sources/faros-graphql-source/src/streams/faros-graph.ts
```diff
@@ -141,7 +163,21 @@ export class FarosGraph extends AirbyteStreamBase {
       refreshedAtMillis = max([refreshedAtMillis, recordRefreshedAtMillis]);
 
       this.state = {...this.state, [query]: {refreshedAtMillis}};
```
tovbinm, line 165: This is not very efficient: `this.state = {...this.state, [query]: {refreshedAtMillis}};` avoid the copy `...this.state`

tovbinm: `this.state[query] = {refreshedAtMillis};`

tovbinm: + perhaps also check for undefined before that

tovbinm (changes requested):
see comments

sources/faros-graphql-source/resources/spec.json
ypc-faros: Done, thanks!

destinations/airbyte-faros-destination/src/converters/faros-graph/faros_graph.ts
ypc-faros: 😉

sources/faros-graphql-source/src/streams/faros-graph.ts
ypc-faros: Done!

I think `this.state` is guaranteed to be defined. Or did I misunderstand your comment?

tovbinm (approved):
can you also please add a test where `result_model` is set to `Flat`?

sources/faros-graphql-source/src/index.ts
```diff
@@ -16,13 +16,19 @@ export enum GraphQLVersion {
   V2 = 'v2',
 }
 
+export enum ResultModel {
+  NESTED = 'Nested',
+  FLAT = 'Flat',
+}
+
```
tovbinm, line 23: I would just keep the values simple capital case:
```
export enum ResultModel {
  Nested = 'Nested',
  Flat = 'Flat',
}
```

tovbinm (approved):

PR #803, opened 2022-11-29
FAI-3971: Batch mutations in Faros Destination

ted-faros:
## Description

Fix batch size configuration and set defaults at current optimal.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -235,8 +235,8 @@ export class GraphQLClient {
     logger: AirbyteLogger,
     schemaLoader: SchemaLoader,
     backend: GraphQLBackend,
-    upsertBatchSize = 1000,
-    mutationBatchSize = 1
+    upsertBatchSize = 10000,
+    mutationBatchSize = 100
   ) {
     this.logger = logger;
     this.schemaLoader = schemaLoader;
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -235,8 +235,8 @@ export class GraphQLClient {
     logger: AirbyteLogger,
     schemaLoader: SchemaLoader,
     backend: GraphQLBackend,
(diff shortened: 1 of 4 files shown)


ypc-faros (approved):

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -235,8 +235,8 @@ export class GraphQLClient {
     logger: AirbyteLogger,
     schemaLoader: SchemaLoader,
     backend: GraphQLBackend,
-    upsertBatchSize = 1000,
-    mutationBatchSize = 1
+    upsertBatchSize = 10000,
```
tovbinm, line 238: Why keep the defaults in client ctor at all?

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -145,8 +145,8 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
         this.logger,
         schemaLoader,
         backend,
-        config.edition_configs.community_graphql_batch_size,
-        config.edition_configs.community_graphql_mutation_batch_size
+        config.edition_configs.community_graphql_upsert_batch_size,
```
tovbinm, line 148: I think you need to pass the default here if config value is null or undefined, e.g "config.value ?? default"

ogusak: looks like in case of passed value as `undefined`, TS would use the default value of the param

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: removed

ogusak (approved):

destinations/airbyte-faros-destination/src/destination.ts
ted-faros: The config value has a default and we [automatically provide it](https://github.com/faros-ai/airbyte-connectors/blob/main/faros-airbyte-cdk/src/destinations/destination-runner.ts#L113) now.

PR #804, opened 2022-11-29
Remove metadata/refreshedAt from result record

ypc-faros:
## Description

Returning these fields causes issues in the destination:

```
errors":[{"extensions":{"code":"validation-failed","path":"$.selectionSet.insert_vcs_User.args.objects[0].refreshedAt"},"message":"field 'refreshedAt' not found in type: 'vcs_User_insert_input'"}]}}}
```

We could instead remove them in the destination, but since we only use them as a watermark for the incremental syncs and they get automatically updated on write, they're really not useful in the returned record.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/faros-graphql-source/src/streams/faros-graph.ts
+++ b/sources/faros-graphql-source/src/streams/faros-graph.ts
@@ -174,6 +174,10 @@ export class FarosGraph extends AirbyteStreamBase {
 
       this.state[query] = {refreshedAtMillis};
 
+      // Remove metadata/refreshedAt
+      // We only use these fields for updating the incremental state
+      const {metadata, refreshedAt, ...result} = item;
+
       yield _.set(
         {},
         this.config.result_model === ResultModel.Nested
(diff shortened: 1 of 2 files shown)


sources/faros-graphql-source/src/streams/faros-graph.ts
```diff
@@ -174,6 +174,10 @@ export class FarosGraph extends AirbyteStreamBase {
 
       this.state[query] = {refreshedAtMillis};
 
+      // Remove metadata/refreshedAt
+      // We only use these fields for updating the incremental state
+      const {metadata, refreshedAt, ...result} = item;
```
ted-faros, line 179: lodash `omit` might be a little more to the point here

ted-faros (approved):

sources/faros-graphql-source/src/streams/faros-graph.ts
ypc-faros: Oh, that's great 👍 !

tovbinm (approved):

PR #805, opened 2022-11-29
Disable upsert batching for CE

ted-faros:
## Description

Disabling the recently added upsert batching for CE.  The proximate issue is odd behavior from Hasura for batch inserts.  Specifically, the following query fails if it has more than one object:

 ```
mutation {
  insert_tms_User(objects: [
    {uid: "u1", source: "GitHub", origin: "mytestsource"},
    {uid: "u2", name: "chalenge", source: "GitHub", origin: "mytestsource"}, 
  ], on_conflict: {constraint: tms_User_pkey, 
    update_columns: [emailAddress, name, origin, refreshedAt]}) {
    returning {
      id
      source
      uid
    }
  }
}
```

The error given is:

```
{
  "error": {
    "exec_status": "FatalError",
    "hint": null,
    "message": "cannot insert into column 'id'",
    "status_code": "42601",
    "description": "Column 'id' is a generated column."
  }
}
```
There is no attempt to insert `id` here.  Our use of permissions must be preventing this in the SaaS product. 

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -202,7 +202,7 @@ function addLevel(u: Upsert): [Upsert, number][] {
  * Client for writing records as GraphQL mutations.  The client supports 3
  * kinds of writes: Upserts, Updates and Deletes.
  *
- * For upserts, the high-level algorithm is:
+ * For upserts (when upsertBatchSize is greater than 0), the high-level algorithm is:
  *
  * > For each record, build a tree of Upserts. The tree has more than one node if the current record represents a
  *   nested entity (e.g. branch record referencing repo which, in-turn, references org).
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by ypc-faros — APPROVED** [review]

File: destinations/airbyte-faros-destination/src/common/graphql-client.ts [hunk 1]
```diff
@@ -241,6 +241,14 @@ export class GraphQLClient {
     this.logger = logger;
     this.schemaLoader = schemaLoader;
     this.backend = backend;
+    assert(
+      upsertBatchSize >= 0,
```
ypc-faros, line 245: I think you'll have to allow `undefined` here for the tests to pass.

`upsertBatchSize === undefined || upsertBatchSize >= 0`

ted-faros: Yeah.  Decided to handle this w/ defaults in the constructor.  An `undefined` will lead to weird runtime errors if it gets past here.

PR #819, opened 2022-12-02
Upgrade faros-js-client + tweak state

ypc-faros:
## Description

Upgrade faros-js-client and use references info when creating incremental queries (`V2` only) 
Also, tweak state management by using the input query (if provided) as state key. Otherwise we use the model names as keys

### DIFF
--- a/sources/faros-graphql-source/src/streams/faros-graph.ts
+++ b/sources/faros-graphql-source/src/streams/faros-graph.ts
@@ -12,6 +12,7 @@ import {
   PathToModel,
   pathToModelV1,
   pathToModelV2,
+  Schema,
   toIncrementalV1,
   toIncrementalV2,
 } from 'faros-js-client';
@@ -28,7 +29,7 @@ const INFINITY = 7258118400000;
 const INFINITY_ISO_STRING = new Date(INFINITY).toISOString();
 
 type GraphQLState = {
-  [query: string]: {refreshedAtMillis: number};
+  [queryOrModelName: string]: {refreshedAtMillis: number};
 };
 
 type StreamSlice = {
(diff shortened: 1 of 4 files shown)


sources/faros-graphql-source/src/streams/faros-graph.ts
```diff
@@ -110,12 +117,14 @@ export class FarosGraph extends AirbyteStreamBase {
     streamState?: GraphQLState
   ): AsyncGenerator<Dictionary<any, string>, any, undefined> {
     const {query, incremental, pathToModel}: StreamSlice = streamSlice;
+    const stateKey = incremental ? pathToModel.modelName : query;
```
ted-faros, line 120: since the query might be large, would it make sense to use a hash of the query here?  Granted it would be less readable, but you could log the mapping of query -> hash for debugging.  Just a thought

ted-faros (approved):
looks good.  a suggestion in there.

sources/faros-graphql-source/src/streams/faros-graph.ts
tovbinm: I said exactly the same to @ypc-faros 😄 

1. with user provided input query, we could use query name, if not present - hash the query
2. when no user provided input query is present, i.e where we query all models, we can use the model name as a type

ypc-faros: For 1) I'd have to parse the input query just to see if the user entered a named query. I think I'll just hash it directly. Thoughts? @tovbinm @ted-faros ?

2) Is already done here 👍

tovbinm: 1. it's pretty easy to do with `parse` / `visit` function from `graphql` package. see
Screenshot 2022-12-02 at 13 46 34

see - https://stackoverflow.com/questions/49047259/how-to-parse-graphql-request-string-into-an-object

2. cool!!

ted-faros: the user-supplied name is not guaranteed to be unique so hashing it seems fast and safe.

ypc-faros: Good point @ted-faros 🙏 
I'll just hash then

tovbinm: Convincing. Please log the hash though.

ypc-faros: Done, thank you both 🙏 !

PR #822, opened 2022-12-06
Upgrade faros-js-client to 0.2.6

ypc-faros:
## Description

Upgrade faros-js-client to 0.2.6 and cuts most of the dependencies on faros-feeds-sdk


tovbinm (approved):
lgtm!

tovbinm:
@ypc-faros I presume eventually we should be able to remove Faros client from feeds-sdk, and instead just add faros-js-client as dependency?

ypc-faros:
> @ypc-faros I presume eventually we should be able to remove Faros client from feeds-sdk, and instead just add faros-js-client as dependency in other projects?

Yes! That's the goal.

I didn't want to remove it completely from feeds-sdk just yet to not force people to upgrade in case there's many changes needed and an urgency to get a PR merged.

PR #831, opened 2022-12-09
Paginate pull requests & branch commits in Azure Repositories source

tovbinm:
## Description

Paginate pull requests & branch commits in Azure Repositories source.

Additionally added:
1. `max_commits_per_branch` param to limit number of commits we fetch per branch
2. `page_size` to control the page size while paginating
3. `request_timeout` param to allow controlling the timeout of the requests we make
4.  bump defaults of `api_version` and `graph_version` to 7.x

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change

## Relates issues

We should follow up and split branches and commits into separate streams - https://github.com/faros-ai/airbyte-connectors/issues/832

As a stop gap I added `max_commits_per_branch` argument to limit the number of commits we fetch.

### DIFF
--- a/sources/azure-repos-source/src/azure-repos.ts
+++ b/sources/azure-repos-source/src/azure-repos.ts
@@ -26,6 +26,7 @@ const DEFAULT_API_VERSION = '7.0';
 const DEFAULT_GRAPH_VERSION = '7.1-preview.1';
 export const DEFAULT_PAGE_SIZE = 100;
 export const DEFAULT_MAX_COMMITS_PER_BRANCH = 1000;
+export const DEFAULT_REQUEST_TIMEOUT = 60000;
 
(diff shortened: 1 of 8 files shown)


sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -23,21 +22,29 @@ import {
...
   readonly organization: string;
   readonly project: string;
   readonly api_version?: string;
   readonly graph_version?: string;
+  readonly page_size?: number;
+  readonly max_commits_per_branch?: number;
+  readonly request_timeout?: number;
 }
 
 export class AzureRepo {
```
cjwooo, line 42: Let's rename this class to `AzureRepos`.

cjwooo (approved):

PR #835, opened 2022-12-12
Azure repos: added await & retry logic based on rate limit headers

tovbinm:
## Description

Azure repos: added await & retry logic based on rate limit headers

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Extra info

Assuming Axios response headers are lowercased

### DIFF
--- a/sources/azure-repos-source/src/azure-repos.ts
+++ b/sources/azure-repos-source/src/azure-repos.ts
@@ -190,7 +190,7 @@ export class AzureRepos {
             `(TSTU remaining: ${retryRemaining}, TSTUs total limit: ${retryRatelimit})`
         );
         await this.sleep(Number.parseInt(retryAfterSecs) * 1000);
-        await this.getHandleNotFound(path, conf);
+        return await this.getHandleNotFound(path, conf);
       }
       if (err?.response?.status === 404) {
         return undefined;
--- a/sources/azure-repos-source/src/azure-repos.ts
+++ b/sources/azure-repos-source/src/azure-repos.ts
@@ -1,5 +1,5 @@
 import axios, {AxiosInstance, AxiosRequestConfig, AxiosResponse} from 'axios';
-import {wrapApiError} from 'faros-airbyte-cdk';
+import {AirbyteLogger, wrapApiError} from 'faros-airbyte-cdk';
 import {Dictionary} from 'ts-essentials';
 import {VError} from 'verror';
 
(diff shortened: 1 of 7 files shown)


sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -170,6 +179,19 @@ export class AzureRepos {
       const res = this.httpClient.get<T, R>(path, conf);
       return res;
     } catch (err: any) {
+      const retryAfterSecs = err?.response?.headers['retry-after'];
```
thomas-gerber, line 182: according to the documentation, this header also appears on successful requests; I think it is important to look for it there for proper throttling. Catching it during errors should be a fallback.

sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -170,6 +179,19 @@ export class AzureRepos {
...
+      if (retryAfterSecs) {
```
thomas-gerber, line 183: when we catch it during errors, do we have a way to stop trying after a while?

thomas-gerber (changes requested):

sources/azure-repos-source/src/azure-repos.ts
tovbinm: see - https://learn.microsoft.com/en-us/azure/devops/integrate/concepts/rate-limits?view=azure-devops#recommendations

sources/azure-repos-source/src/azure-repos.ts
tovbinm: I can put a limit of retries.

sources/azure-repos-source/src/azure-repos.ts
thomas-gerber: yes exactly; here you only check if an error is caught; you should also check if the request is successful and wait there as well.
`We recommend that you at least respond to the Retry-After header. If you detect a Retry-After header in **any** response, wait until that amount of time has passed before sending another request. Doing so helps your client application experience fewer enforced delays. Keep in mind that **the response is 200**, so you don't need to apply retry logic to the request.`

tovbinm: added

tovbinm: fixed

thomas-gerber (approved):
Approved with the caveat that there is no test for it

PR #863, opened 2023-01-10
Honor deletion records

ypc-faros:
## Description

I think I added this as a precaution to prevent accidentally dropping records, but it is actually unnecessary and results in different behavior of feed depending on whether it is run directly or via the Faros Feeds wrapper.

It is safe to remove because, if we run a full sync in Airbyte, [all](https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/faros-feeds/model_names.ts) models are reset, so it doesn't matter what happens in the feed. Feeds will always issue deletion records for "non-incremental" models only (unless invoked with `--no-incremental`) so it's safe to process these deletion records.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
+++ b/destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
@@ -37,12 +37,6 @@ export class FarosFeed extends Converter {
 
     const [model, rec] = Object.entries(data).pop();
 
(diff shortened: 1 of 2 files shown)


tovbinm:
@ypc-faros we might need to update the models list here again - https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/faros-feeds/model_names.ts

tovbinm (approved):

PR #865, opened 2023-01-11
Upgrade destination spec w.r.t oneOf

thomas-gerber:
See https://docs.airbyte.com/connector-development/connector-specification-reference/#using-oneof

Needed for a move to v40.26. When trying to use the current spec, Airbyte UI complains about:
```
Spec uses oneOf without a shared const property
```

## Type of change
- [x] Bug fix

Note that this spec appears to load properly, BUT:
- [ ] boolean parameters cannot be switched on (e.g. Dry Run); not sure why
- [ ] I think ALL source specific configs should be valid at once, correct? not sure this is what we have now
- [ ] did not test on an old version of Airbyte
Screenshot 2023-01-10 at 11 37 02 PM
Screenshot 2023-01-10 at 11 36 05 PM


thomas-gerber:
@ypc-faros @cjwooo

tovbinm:
@cjwooo @ypc-faros can you please review?

ypc-faros (commented):
Before moving forward, I think there's a product decision we need to make here.

The first thing to note is that, in the current spec, the `oneOf` under `source_specific_configs` is unnecessary, since it contains a single element (the object with title `Configurations`)

That being said, we need to decide whether we want the user to be able to specify more than one of these source specific cfgs at once (in a single destination configuration). Having the `oneOf` has the benefit of less UI clutter (user selects a single source via the dropdown), but the disadvantage is that then the user has to configure N destinations (one per source-specific configuration they want to enter).

Lastly, if we make any changes to the spec, there's most likely changes needed in the code ([example](https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/destination.ts#L317)). E.g., for the change proposed here, we can no longer reference `source_specific_configs.agileaccelerator` but have to check for `source_specific_configs.source === agileaccelerator` first.

thomas-gerber:
I favor being able to configure all systems in one destination (current behavior).

PR #869, opened 2023-01-13
Add option to accept incoming records origin

ypc-faros:
## Description

Add option to accept incoming records origin and possibly replace them according to a mapping.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/resources/spec.json
+++ b/destinations/airbyte-faros-destination/resources/spec.json
@@ -147,22 +147,33 @@
           "my-faros-destination"
         ]
       },
-      "keep_records_origin": {
+      "accept_input_records_origin": {
         "order": 2,
         "type": "boolean",
-        "title": "Keep input records origin",
-        "description": "Whether to keep the origin field (if present) of input records unchanged",
+        "title": "Accept input records origin",
+        "description": "Whether to accept the origin field (if present) of input records. If enabled, the input records origin will only be changed according to the replacements specified by 'Replace Records Origin'. Otherwise, the input records origin are ignored and overwritten.",
         "default": false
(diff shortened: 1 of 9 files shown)


ted-faros (approved):

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -147,15 +147,22 @@
           "my-faros-destination"
         ]
       },
-      "dry_run": {
+      "keep_records_origin": {
```
tovbinm, line 150: @ypc-faros  how is this going to interact with combination records origin map?

tovbinm (changes requested):
I would like to understand this better before approving

destinations/airbyte-faros-destination/resources/spec.json
ypc-faros: This is meant to be used so that the remapped origin fields are not overwritten. So in the faros-graphql source we configure how to remap the origins and in the destination we toggle this to leave them untouched.

I realize now we could do all this in the destination only, although I'm not sure what's cleaner(remap in source and ask to keep in the dst vs "keep the remapped origin if present" in the dst). Do you have a preference?

ypc-faros: I changed the boolean option to `accept_input_records_origin` since it won't be actually always kept untouched (it will be replaced if it matches any of the origin mapping keys).

Wdyt @tovbinm ?

tovbinm (approved):

Issue #870, opened 2023-01-16
[Jenkins Source] Github OAuth


konfortes:
Hey guys,
We use Github oauth in Jenkins, therefore, configuring a source with user and API token is not possible for me.
Also - Jenkins jobs and builds can be accessed by non-authenticated users.
Any idea what can I do to use this source?

tovbinm:
Are you suggesting that making API key authentication requirement optional would work for you?

konfortes:
Hey @tovbinm 
Yep, it will definitely solve my issue

konfortes:
@tovbinm I created this PR, lmk wdyt:
[https://github.com/faros-ai/airbyte-connectors/pull/871](https://github.com/faros-ai/airbyte-connectors/pull/871)

tovbinm:
Released in https://github.com/faros-ai/airbyte-connectors/releases/tag/v0.4.52

tovbinm:
Thank you for the contribution @konfortes

PR #871, opened 2023-01-18
[Jenkins Source] - Change user and token to be optional

konfortes:
Jenkins instances can be configured to expose jobs and builds to unauthenticated users

## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#870](https://github.com/faros-ai/airbyte-connectors/issues/870) 

## Migration notes

> No migrations

## Extra info

### DIFF
--- a/sources/jenkins-source/src/jenkins.ts
+++ b/sources/jenkins-source/src/jenkins.ts
@@ -13,8 +13,8 @@ const POTENTIAL_MAX_DEPTH = 10;
 
 export interface JenkinsConfig extends AirbyteConfig {
   readonly server_url: string;
-  readonly user: string;
-  readonly token: string;
+  readonly user?: string;
+  readonly token?: string;
   readonly depth?: number;
   readonly pageSize?: number;
   readonly last100Builds?: boolean;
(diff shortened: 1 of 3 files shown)


konfortes:
recheck

sources/jenkins-source/src/jenkins.ts
```diff
@@ -99,8 +99,10 @@ export class Jenkins {
       throw new VError('server_url: must be a valid url');
     }
 
-    jenkinsUrl.username = config.user;
-    jenkinsUrl.password = config.token;
+    if (config.user && config.token) {
```
tovbinm, line 102: I think we should error if one of them is set and other isnt

tovbinm (commented):
you also need to remove the requirement from here - https://github.com/faros-ai/airbyte-connectors/blob/main/sources/jenkins-source/resources/spec.json#L7

sources/jenkins-source/src/jenkins.ts
konfortes: sounds reasonable, will add it

sources/jenkins-source/src/jenkins.ts
```diff
@@ -61,6 +61,34 @@ export class Jenkins {
...
   }
 
+  private static validateConfig(config: JenkinsConfig): void {
+    if (typeof config.server_url !== 'string') {
+      throw new VError('server_url: must be a string');
+    }
+
+    if (config.user || config.token) {
+      if (!config.user || !config.token) {
+        throw new VError('user and token must be both specified or both empty');
```
tovbinm, line 71: `user and token must be either both specified or both be empty`

tovbinm (approved):

sources/jenkins-source/src/jenkins.ts
konfortes: Thanks, Fixed

tovbinm (approved):

tovbinm:
Please resolve conflicts so we could merge

konfortes:
@tovbinm conflicts resolved

tovbinm (approved):

PR #876, opened 2023-01-23
FAI-4937 - Support * to sync all projects

willmarks:
## Description

Support '*' to sync all projects. Update field description. Provide Item to array type field.

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


cjwooo:
Let's also fail if they specify `*` plus some other values.

cjwooo (approved):

tovbinm (approved):

PR #881, opened 2023-01-24
Fix root level id replacement

ypc-faros:
## Description

`_.set` doesn't mutate the object when `path` is `[]`

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


tovbinm (approved):
Thanks for the test

PR #886, opened 2023-01-25
Allow filtering Squadcast services & incidents by service slugs + update deps

tovbinm:
## Description

Allow filtering Squadcast services & incidents by service slugs

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/sources/squadcast-source/resources/spec.json
+++ b/sources/squadcast-source/resources/spec.json
@@ -32,7 +32,7 @@
           "type": "string"
         },
         "title": "Service Slugs",
-        "description": "List of service slugs, e.g payment-api-service. If empty or '*' all available services will be processed."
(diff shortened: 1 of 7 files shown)


sources/squadcast-source/resources/spec.json
```diff
@@ -26,6 +26,14 @@
         "title": "Incident Owner ID",
         "description": "Team ID or User ID"
       },
+      "services": {
+        "type": "array",
+        "items": {
+          "type": "string"
+        },
+        "title": "Services",
+        "description": "Space-delimited list of service names. If empty or '*' all available services will be processed."
```
ypc-faros: nit: I think you can remove "Space-delimited" here. When specifying this in a json file, it is actually a json array.

ypc-faros (approved):

PR #888, opened 2023-01-26
Make it possible to partition full model queries

ypc-faros:
## Description

So that we can split reading a full graph into N buckets concurrently.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/faros-graphql-source/resources/spec.json
+++ b/sources/faros-graphql-source/resources/spec.json
@@ -84,16 +84,19 @@
         "description": "Legacy V1 schema as a gzip compressed, base 64 encoded text. Only used when 'Adapt V1 query' is enabled.",
         "multiline": true
       },
-      "replace_origin_map": {
+      "bucket_id": {
         "order": 9,
-        "type": "string",
-        "title": "Replace Records Origin",
-        "description": "JSON map to replace records origin values, where each entry (origin1, origin2) indicates that the output records' origin property should be replaced from 'origin1' by 'origin2'",
-        "multiline": true,
-        "default": "{}",
-        "examples": [
-          "{ \"originA\": \"originB\" }"
-        ]
+        "type": "integer",
+        "title": "Bucket Number",
+        "description": "Bucket number for this source to determine which portion of models to pull. Use it when distributing the load between multiple sources.",
+        "default": 1
+      },
+      "bucket_total": {
+        "order": 10,
+        "type": "integer",
+        "title": "Total Number of Buckets",
+        "description": "Total number of buckets to distribute models across. Use it when distributing the load between multiple sources.",
+        "default": 1
       }
     }
   }
--- a/sources/faros-graphql-source/resources/spec.json
+++ b/sources/faros-graphql-source/resources/spec.json
@@ -88,14 +88,14 @@
         "order": 9,
         "type": "integer",
         "title": "Bucket Number",
-        "description": "Bucket number for this source to determine which portion of models to pull. Use it when distributing the load between multiple sources.",
+        "description": "Bucket number for this source to determine which portion of models to pull. Use it when distributing the load between multiple sources. Cannot be used in combination with 'GraphQL query'",
         "default": 1
       },
       "bucket_total": {
         "order": 10,
         "type": "integer",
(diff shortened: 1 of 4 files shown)


sources/faros-graphql-source/resources/spec.json
```diff
@@ -84,16 +84,19 @@
         "description": "Legacy V1 schema as a gzip compressed, base 64 encoded text. Only used when 'Adapt V1 query' is enabled.",
         "multiline": true
       },
-      "replace_origin_map": {
```
ypc-faros, line 87: I had forgotten to remove `replace_origin_map`. It was moved to the destination in a previous PR

sources/faros-graphql-source/src/index.ts
```diff
@@ -81,6 +84,24 @@ export class FarosGraphSource extends AirbyteSourceBase<GraphQLConfig> {
             " when 'Adapt V1 query' is enabled"
         );
     }
+
+    if (config.query) {
+      if (config.bucket_id !== undefined)
+        throw new VError('Bucket id cannot be used in combination with query');
+      if (config.bucket_total !== undefined)
+        throw new VError(
+          'Bucket total cannot be used in combination with query'
```
tovbinm, line 93: Let's add this to description

sources/faros-graphql-source/src/index.ts
```diff
@@ -81,6 +84,24 @@ export class FarosGraphSource extends AirbyteSourceBase<GraphQLConfig> {
             " when 'Adapt V1 query' is enabled"
         );
     }
+
+    if (config.query) {
+      if (config.bucket_id !== undefined)
+        throw new VError('Bucket id cannot be used in combination with query');
```
tovbinm, line 90: Let's also add this to description

tovbinm (approved):
@cjwooo can you please have a look? This is quite similar to what you did with GitHub source, but now applied to source slices.

tovbinm:
@ypc-faros why not create a slice per model?

ypc-faros:
> @ypc-faros why not create a slice per model?

We do actually. Now that I think about it this probably doesn't help much if Airbyte runs the slices in parallel. Does it @cjwooo ?

ypc-faros:
> @cjwooo can you please have a look? This is quite similar to what you did with GitHub source, but now applied to source slices.

That's exactly where I got inspiration from :)

cjwooo:
Airbyte (or rather, the Airbyte Connector CDK), does NOT run slices in parallel.

tovbinm:
1. What's the benefit having multiple slices by stream?
2. Should we introduce support for parallelism?

cjwooo:
1. Stream slices allow incremental stream state to be updated in a more granular way than the normal interval-based checkpointing, which is only usable if the stream guarantees that records are outputted in ascending order based on the stream's cursor field. Most streams cannot meet that requirement.
2. More investigation is needed. We'd have to see if the Airbyte server will properly handle the output of parallel slices. This could also break our compatibility with community destinations, which may be operating under the assumption that streams and slices are processed sequentially. Performance benefit may also not be worth the effort, given most sources are network bound and often subject to API rate-limiting.

cjwooo (approved):

ypc-faros:
Thank you for your comments @cjwooo @tovbinm  🙏 !

tovbinm:
@cjwooo @ypc-faros so why offer bucketization instead of just keep slicing by model?

ypc-faros:
> @cjwooo @ypc-faros so why offer bucketization instead of just keep slicing by model?

We can (I'm hoping) speed up copying a graph (by splitting into N concurrent copies). It should also help if there are failures. We can retry copying only the buckets that failed.

ypc-faros:
> > @cjwooo @ypc-faros so why offer bucketization instead of just keep slicing by model?
> 
> We can (I'm hoping) speed up copying a graph (by splitting into N concurrent copies). It should also help if there are failures. We can retry copying only the buckets that failed.

Btw the slicing also stays. We're just providing a way to distribute the slices among N concurrent (or not) syncs.

PR #890, opened 2023-01-26
Allows * shortcut to mean all models in jsonata_destination_models

ypc-faros:
## Description

Allows * shortcut to mean all models in jsonata_destination_models

## Type of change
- [x] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/resources/spec.json
+++ b/destinations/airbyte-faros-destination/resources/spec.json
@@ -181,7 +181,7 @@
           "type": "string"
         },
         "title": "JSONata destination models",
-        "description": "Destination models when using JSONata expression.",
+        "description": "Destination models when using JSONata expression. A single '*' can be provided as a shortcut to include all models.",
         "examples": ["ims_Incident", "vcs_Commit"]
       },
       "source_specific_configs": {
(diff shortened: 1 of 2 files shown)


destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -181,7 +181,7 @@
           "type": "string"
         },
         "title": "JSONata destination models",
-        "description": "Destination models when using JSONata expression.",
+        "description": "Destination models when using JSONata expression. A single '*' can be provided as a shortcut to include all models.",
```
tovbinm, line 184: `Destination models when using JSONata expression. If '*' is provided we assume that all models are being modified.`

destinations/airbyte-faros-destination/src/converters/jsonata.ts
```diff
@@ -10,6 +10,7 @@ import {
   DestinationRecord,
   StreamContext,
 } from './converter';
+import { ALL_MODEL_NAMES } from './faros-feeds/model_names';
```
tovbinm, line 13: did you update the `faros-feeds/model_names` recently? we need to make sure all the model names are there.

tovbinm (approved):

destinations/airbyte-faros-destination/src/converters/jsonata.ts
ypc-faros: I've just checked. They're all there 👍

PR #892, opened 2023-01-26
GitLab Converter: Process nested group repo keys correctly

### LINKED ISSUE #891: GitLab Data Model is Incorrect
Hello Team,

First let me say that I'm not sure if the issues here are with the airbyte source or the converters, but the data being stored in the Faros Destination for GitLab omits some important bits that lead to incorrect data. From what I've seen thus far, all of the problems stem from using partially unique data as the Primary Keys for tables. Specifically, the ids don't seem to be taking into account that GitLab supports nested groups.

If I have projects in the following layout in my GitLab instance. some interesting things happen:

* `foo/bar/baz`
* `foo/bar/qux`
* `foo/qux`

Looking in `vcs_Repository`, you might see:

| id | uid | name | fullName |
| --- | --- | --- | --- |
| GitLab&#124;foo&#124;bar | bar | bar | foo / bar / qux |
| GitLab&#124;foo&#124;qux | qux | qux | foo / qux |

* Only 2 of my projects are listed because `foo / bar / baz` gets clobbered since id is only using a portion of the project path.
* The project name for `foo / bar / qux` is wrong.

This type of issue spreads throughout the model. For instance, `cicd_Releases` only uses the release name as part of the PK, so any projects that have the same release name are not uniquely captured and the description associated with the release may be incorrect.

`cicd_ReleaseTagAssociations` only has the partial repository id and therefore groups all releases across subgroups. So in this example, if we have release `v0.1.0` in `foo / bar / baz`, it would falsely be associated with `foo / bar / qux`

I'll try to dig a bit further into where the problem can be resolved, but wanted to raise the issue in case there's a quick resolution that someone more familiar with the codebase could assist with.

BrentSouza:
## Description

GitLab supports nested groups which need to be considered when parsing the repo key. For example,

`top-level-group / nested-group / project`

Currently, `nested-group` is being treated as the project name and uid. This patch changes the parsing so that:

name = project
uid = nested-group / project

## Type of change
- [x] Bug fix
- [ ] New feature
- [x] Breaking change

## Related issues

Fix #891

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
@@ -69,25 +69,51 @@ export class GitlabCommon {
 
   static parseRepositoryKey(
     webUrl: string | undefined,
-    source: string,
-    startIndex = 3
+    source: string,    
   ): undefined | RepositoryKey {
-    if (!webUrl) return undefined;
-    const repositoryIndex = startIndex + 1;
-    
-    const nameParts: ReadonlyArray<string> = webUrl.split('/');
-    if (nameParts.length <= repositoryIndex) return undefined;
+    if (!webUrl) {
+      return undefined;
+    }
+    return this.parseKeyFromUrl(webUrl, source, false);
+  }
 
+  static parseGroupKey(
+    webUrl: string | undefined,
+    source: string,    
+  ): undefined | OrgKey {
+    if (!webUrl) {
+      return undefined;
+    }
+    return this.parseKeyFromUrl(webUrl, source, true);
+  }
+
+  private static parseKeyFromUrl<B extends boolean>(
+    webUrl: string,
+    source: string,
+    hasGroupPrefix: B
+  ): B extends true ? OrgKey : RepositoryKey;
+
+  private static parseKeyFromUrl(
+    webUrl: string,
+    source: string,
+    hasGroupPrefix: boolean
+  ): OrgKey | RepositoryKey {
+    const startIndex = hasGroupPrefix ? 4 : 3;
+    const nameParts: ReadonlyArray<string> = webUrl.split('/');
     const endIndex = nameParts.indexOf('-') == -1 ? nameParts.length : nameParts.indexOf('-');
+    const uid = nameParts.slice(startIndex, endIndex).join('/').toLowerCase();
     
-    const organization = nameParts[startIndex].toLowerCase();
-    const repositoryUid = nameParts.slice(repositoryIndex, endIndex).join('/').toLowerCase();
-    const repositoryName = nameParts[endIndex - 1].toLowerCase()
+    if (hasGroupPrefix) {
+      return { uid: uid, source };
+    }
     
+    const org = nameParts[startIndex].toLowerCase();
+    const name = nameParts[endIndex - 1].toLowerCase();
+
     return {
-      name: repositoryName,
-      uid: repositoryUid,
-      organization: {uid: organization, source},
+      organization: { uid: org, source },
+      uid,
+      name,
     };
   }
 
--- a/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
@@ -29,13 +29,56 @@ export class GitlabCommon {
   // Max length for free-form description text fields such as issue body
   static readonly MAX_DESCRIPTION_LENGTH = 1000;
 
-  static tms_ProjectBoard_with_TaskBoard(
-    projectKey: ProjectKey,
+  static mapRepositoryHierarchy<T>(
+    repository: RepositoryKey,
+    callback: (k: ProjectKey) => T,
+  ): T[] {
+    return repository.uid.split('/').map((_, i, a) => {
+      const key = {
+        uid: a.slice(0, i + 1).join('/'),
+        source: repository.organization.source,
+      };
+      return callback(key);
+    })
+  }
+
+  static tms_TaskBoard(
+    boardKey: ProjectKey,
     name: string,
+  ): DestinationRecord {
+    return {
+      model: 'tms_TaskBoard',
+      record: {
+        ...boardKey,
+        name,
+      },
+    };
+  }
+
+  static tms_TaskBoardProjectRelationship(
+    boardKey: ProjectKey,
+    projectKey: ProjectKey,
+  ): DestinationRecord {
+    return {
+      model: 'tms_TaskBoardProjectRelationship',
+      record: {
+        board: boardKey,
+        project: projectKey,
+      },
+    };
+  }
+
+  static tms_ProjectBoard_with_TaskBoard(
+    repository: RepositoryKey,
     description: string | null,
     createdAt: string | null | undefined,
     updatedAt: string | null | undefined
   ): DestinationRecord[] {
+    const projectKey = {
+      uid: repository.uid,
+      source: repository.organization.source,
+    };
+    const name = repository.name;
     return [
       {
         model: 'tms_Project',
@@ -50,20 +93,13 @@ export class GitlabCommon {
           updatedAt: Utils.toDate(updatedAt),
         },
       },
-      {
-        model: 'tms_TaskBoard',
-        record: {
-          ...projectKey,
-          name,
-        },
-      },
-      {
-        model: 'tms_TaskBoardProjectRelationship',
-        record: {
-          board: projectKey,
-          project: projectKey,
-        },
-      },
+      this.tms_TaskBoard(projectKey, name),
+      this.tms_TaskBoardProjectRelationship(repository.organization, projectKey),
+      // traverse the hierarchy to link boards for the sub groups and the project itself
+      ...this.mapRepositoryHierarchy<DestinationRecord>(
+        repository,
+        key => this.tms_TaskBoardProjectRelationship(key, projectKey),
+      ),
     ];
   }
 
@@ -101,7 +137,7 @@ export class GitlabCommon {
     const startIndex = hasGroupPrefix ? 4 : 3;
     const nameParts: ReadonlyArray<string> = webUrl.split('/');
     const endIndex = nameParts.indexOf('-') == -1 ? nameParts.length : nameParts.indexOf('-');
-    const uid = nameParts.slice(startIndex, endIndex).join('/').toLowerCase();
+    const uid = nameParts.slice(startIndex + 1, endIndex).join('/').toLowerCase();
     
     if (hasGroupPrefix) {
       return { uid: uid, source };
--- a/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
@@ -94,7 +94,6 @@ export class GitlabCommon {
         },
       },
       this.tms_TaskBoard(projectKey, name),
-      this.tms_TaskBoardProjectRelationship(repository.organization, projectKey),
       // traverse the hierarchy to link boards for the sub groups and the project itself
       ...this.mapRepositoryHierarchy<DestinationRecord>(
         repository,
--- a/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
@@ -74,16 +74,20 @@ export class GitlabCommon {
   ): undefined | RepositoryKey {
     if (!webUrl) return undefined;
     const repositoryIndex = startIndex + 1;
+    
+    const nameParts: ReadonlyArray<string> = webUrl.split('/');
+    if (nameParts.length <= repositoryIndex) return undefined;
 
-    const orgRepo: ReadonlyArray<string> = webUrl.split('/');
-    if (orgRepo.length <= repositoryIndex) return undefined;
-
-    const organization = orgRepo[startIndex];
-    const repositoryName = orgRepo.slice(repositoryIndex, orgRepo.length).join(' / ');
+    const endIndex = nameParts.indexOf('-') == -1 ? nameParts.length : nameParts.indexOf('-');
+    
+    const organization = nameParts[startIndex].toLowerCase();
+    const repositoryUid = nameParts.slice(repositoryIndex, endIndex).join('/').toLowerCase();
+    const repositoryName = nameParts[endIndex - 1].toLowerCase()
+    
     return {
-      name: orgRepo[orgRepo.length - 1].toLowerCase(),
-      uid: repositoryName?.toLowerCase(),
-      organization: {uid: organization?.toLowerCase(), source},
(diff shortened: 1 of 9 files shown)


destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
```diff
@@ -76,12 +76,12 @@ export class GitlabCommon {
     const repositoryIndex = startIndex + 1;
 
     const orgRepo: ReadonlyArray<string> = webUrl.split('/');
-    if (orgRepo.length < repositoryIndex) return undefined;
+    if (orgRepo.length <= repositoryIndex) return undefined;
 
     const organization = orgRepo[startIndex];
-    const repositoryName = orgRepo[repositoryIndex];
+    const repositoryName = orgRepo.slice(repositoryIndex, orgRepo.length).join(' / ');
```
tovbinm, line 82:
```suggestion
const repositoryUid = orgRepo.slice(repositoryIndex, orgRepo.length).join('/').toLowerCase();
const repositoryName = orgRepo[orgRepo.length - 1].toLowerCase();
return {
      name: repositoryName,
      uid: repositoryUid,
      organization: {uid: organization?.toLowerCase(), source},
};
```

destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
```diff
@@ -76,12 +76,12 @@ export class GitlabCommon {
     const repositoryIndex = startIndex + 1;
 
     const orgRepo: ReadonlyArray<string> = webUrl.split('/');
-    if (orgRepo.length < repositoryIndex) return undefined;
+    if (orgRepo.length <= repositoryIndex) return undefined;
```
tovbinm, line 79: generally with `startIndex = 3` hardcoded above, we should just rewrite this function (cc @thomas-gerber / @cjwooo )

destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
BrentSouza: Yeah. I was considering this same change to be honest. Do you know if there is any particular reason `toLowerCase()` is called on everything?

BrentSouza:
I have read the CLA Document and I hereby sign the CLA

destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
cjwooo: Let's remove that index parameter and instead parse the webUrl with https://nodejs.org/docs/latest-v16.x/api/url.html#new-urlinput-base

BrentSouza: I have 0 attachment to this code as I didn't create it, but what does parsing the URL get you in this case? The logic is really in parsing the pathname, which still requires some knowledge of where to start looking for the relevant tokens.

cjwooo: Code readability + eliminate error chance of someone calling this function with an invalid index.

cjwooo: Oh I just saw that this function is called in one place with a different startIndex so we can't remove that parameter.

BrentSouza: Just to provide a little more context about what is going on with this function...

I believe the reason the original author of this converter added the startIndex parameter is that one of the streams (group milestones) returns a slightly different pattern for the webUrl. GitLab prefixes this with the word "groups". You can see this in the patterns below:

| stream | web_url |
| ------ | ------- |
| branches | https://gitlab.com/org/repo/-/tree/main |
| commits | https://gitlab.com/org/repo/-/commit/ed899a2f4b50b4370feeea94676502b42383c746 |
| group_milestones | https://gitlab.com/groups/org/-/milestones/42 |
| jobs | https://gitlab.com/org/repo/-/jobs/7 |
| merge_request_commits | https://gitlab.com/org/repo/-/merge_requests/133 |
| merge_requests | http://gitlab.com/org/repo/-/merge_requests/1 |
| pipelines | https://gitlab.com/org/repo/-/pipelines/47 |
| project_milestones | https://gitlab.com/org/repo/-/milestones/1 |
| projects | https://gitlab.com/org/repo |
| tags | https://gitlab.com/org/repo/-/commit/ed899a2f4b50b4370feeea94676502b42383c746 |

So the intent of `startIndex` is just to know where in the path to extract `org`. So I think creating a URL object from `webUrl` in this case doesn't help really. We would end up doing something like:

```ts
static parseRepositoryKey(webUrl, source, hasGroupPrefix = false) {
    const parsedUrl = new URL(webUrl);
    const nameParts = parsedUrl.pathName.split('/');
    const startIndex = hasGroupPrefix ? 1 : 0;
    //  rest of the logic is the same 
   ...
}
```

BrentSouza: Which I'm happy to do if that's what you prefer. I just don't see value in creating a URL object.

Actually I think this function needs some more work regardless to handle that `group_milestones` special case

cjwooo: Is group_milestones the only special case? It seems like we are not converting those records properly, given that we are currently trying to parse a repository from what seems like a group-level item. @tovbinm @thomas-gerber

BrentSouza: It is, but I am really looking deeper now and I think the way group milestones are being handled in general is not correct. Both project and group milestones are being converted to `tms_Epic`. Projects, are loaded to `tms_Project`, however groups are not, so the linkage for `tms_Epic` to `tms_Project` will be missing/incorrect for group milestones.

I think I'll have to modify the converter for `Groups` to also make an entry in `tms_Project`. Since group ids and project ids in GitLab are independent, the issue then is avoiding collision with the ids, but I think that can be managed if we do something like prefix the group id with a `g`.

With that in place, I think we can change group milestones to ignore parsing `web_url` since the `group_id` is a property in the response object.

That would allow us to remove the `streamIndex` parameter.

cjwooo: I haven't really worked with GitLab myself. @thomas-gerber wdyt^? specifically regarding the proposed converter change.

BrentSouza: I ended up doing something slightly different which I think addresses all of the issues. 

* **Preventing improper use of parseRepositoryKey**: There are now two separate functions for handling the different url patterns, with a private function to handle the common parsing logic. 
* **Prevent key overlap in tms_Project**: Groups and Projects will now register with the same uid as vcs_Repository/vcs_Organization, which are full paths and can't overlap.
* **Incorrect data model**: I went through all of the stream converters and made sure they are using values that will prevent clobbering and improper associations to repos/groups

I think this set of changes is pretty close to modeling the data according to what I can see in my GitLab instance. One thing that's missing still is support for GitLab's epics.

Additionally, I opened airbytehq/airbyte#22124 in the airbyte project to address some data erasure from the source. If that gets fixed, it could simplify some of the stream dependencies present in the converters.

destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
```diff
@@ -69,21 +69,51 @@ export class GitlabCommon {
...
+
+  private static parseKeyFromUrl(
+    webUrl: string,
+    source: string,
+    hasGroupPrefix: boolean
+  ): OrgKey | RepositoryKey {
+    const startIndex = hasGroupPrefix ? 4 : 3;
+    const nameParts: ReadonlyArray<string> = webUrl.split('/');
+    const endIndex = nameParts.indexOf('-') == -1 ? nameParts.length : nameParts.indexOf('-');
+    const uid = nameParts.slice(startIndex, endIndex).join('/').toLowerCase();
```
thomas-gerber, line 104: in order not to break existing users who do not use subgroups, I think uid should NOT include the top level folder (i.e. slice with `startIndex + 1`).
for `top-level-group / nested-group / project`, uid is `nested-group / project`; the org is what has the top level.

destinations/airbyte-faros-destination/src/converters/gitlab/groups.ts
```diff
@@ -20,7 +23,7 @@ export class Groups extends GitlabConverter {
     res.push({
       model: 'cicd_Organization',
       record: {
-        uid: group.path,
+        uid: group.full_path,
```
thomas-gerber, line 26: I think we should only create ONE entry for cicd_Organization and ONE entry for vcs_Organization based on the `top-level-group`.
We also create one tms_TaskBoard per Gitlab group (`uid` comment above applies)
Then, in the project stream, we create ONE tms_Project by Project, link it to all boards from its hierarchy,

destinations/airbyte-faros-destination/src/converters/gitlab/issues.ts
```diff
@@ -88,24 +96,47 @@ export class Issues extends GitlabConverter {
...
         source,
       },
     });
 
-    res.push({
-      model: 'tms_TaskProjectRelationship',
-      record: {
-        task: taskKey,
-        project: projectRef,
+    res.push(
```
thomas-gerber, line 104: see comments above on relationships; I think an issue (tms_Task) should link to a single tms_Project, but could link to several tms_TaskBoards.

thomas-gerber (changes requested):
Summary of the changes requested (open to discussion), assuming `top-level-group / nested-group / project`:
1. vcs_Organization and cicd_Organization should have exactly one entry: `top-level-group`
2. tms_Project should have one entry: `nested-group / project`; the schema is missing a tms_Organization entity, when we add it later it will not break any user if we are consistent this way
3. vcs_Repository should have one entry: `nested-group / project`
4. tms_Board should have 2 entries: `project` and `nested-group / project`; as well as the corresponding 2 tms_TaskBoardProjectRelationship entries between the one project and 2 boards
4. an issue in that project should belong to those 2 tms_TaskBoards (through tms_TaskBoardRelationship) and that one tms_Project (through tms_TaskProjectRelationship)

This will ensure that existing users without subgroup are unchanged.
The limitation that I see is that the UID for tms_Project and tms_TaskBoard is not unique (because `top-level-group` is not part of it), but that can be solved later with a tms_Organization entity.

destinations/airbyte-faros-destination/src/converters/gitlab/issues.ts
BrentSouza: I agree. This makes more sense as issues are a Project level item.

destinations/airbyte-faros-destination/src/converters/gitlab/common.ts
BrentSouza: > The limitation that I see is that the UID for tms_Project and tms_TaskBoard is not unique (because top-level-group is not part of it), but that can be solved later with a tms_Organization entity.

Thanks for the feedback. In addition I think `tms_Epic` and `tms_TaskProjectRelationship` would have the same issue and this limitation that you mention is why I included the `top-level-group` in the `uid`. I felt that if I was making a PR to address broken relationships in the data model that I should be complete with it. However, I agree that if there are plans to create a `tms_Organization` entity, that would be a better path. I'll change this back to exclude the `top-level-group`

BrentSouza:
@thomas-gerber I think `cicd_Release` also needs some attention.

With the changes to `vcs_Repository.uid` to include nested groups, which will then flow `vcs_Tags`, we should be able to link everything correctly. However, `cicd_Release.uid` only has the release name which is not unique, so that would cause inaccuracies with descriptions, dates, and urls across projects that have the same release name, which is fairly common if using semver release names.

Should `cicd_Release` also be aware of Org and include `nested-group/project` in the uid?

BrentSouza:
Latest commits should address the changes requested 😃

thomas-gerber:
> @thomas-gerber I think `cicd_Release` also needs some attention.
> 
> With the changes to `vcs_Repository.uid` to include nested groups, which will then flow `vcs_Tags`, we should be able to link everything correctly. However, `cicd_Release.uid` only has the release name which is not unique, so that would cause inaccuracies with descriptions, dates, and urls across projects that have the same release name, which is fairly common if using semver release names.
> 
> Should `cicd_Release` also be aware of Org and include `nested-group/project` in the uid?

Both your points on linkage working and UID not being unique are valid. **I agree that the uid should be `nested-group/project/releasename` (notice the absence of top-level-group).** Later, we may link to a cicd_Organization, although this already links to vcs_Organization through the tag.

thomas-gerber (changes requested):
One small fix needed:
given `top-level-group/nested-group1/nested-group2/project`, you create the following boards:
* top-level-group
* nested-group1
* nested-group1/nested-group2
* nested-group1/nested-group2/project
The first one (top-level-group) should not be there IMO.
(Same remark for tms_TaskBoardRelationship and tms_TaskBoardProjectRelationship)

Otherwise, I agreed with your comment on cicd_Release, and you can update the UID as you suggested (i.e. append `nested-group/project`) even though it may break backward compatibility.

I checked the rest of entities on my test instances with a nested group, and it looks good. I also sanity checked the other converters, and I haven't found any other places to update, great job!

thomas-gerber:
> One small fix needed: given `top-level-group/nested-group1/nested-group2/project`, you create the following boards:
> 
> * top-level-group
> * nested-group1
> * nested-group1/nested-group2
> * nested-group1/nested-group2/project
>   The first one (top-level-group) should not be there IMO.
>   (Same remark for tms_TaskBoardRelationship and tms_TaskBoardProjectRelationship)
> 
> Otherwise, I agreed with your comment on cicd_Release, and you can update the UID as you suggested (i.e. append `nested-group/project`) even though it may break backward compatibility.
> 
> I checked the rest of entities on my test instances with a nested group, and it looks good. I also sanity checked the other converters, and I haven't found any other places to update, great job!

Unless you have a specific use case in mind that I didn't grasp?

BrentSouza:
My thinking was just keeping inline with what you can see in GitLab. If I navigate to a top-level-group in my instance, I can see all issues for child projects no matter where they sit in a nested group hierarchy. So I think a board to represent that top-level-group is a natural expectation for a GitLab user, but I think you could very easily create a query to get all Tasks in a given top-level-group when tms_Organization is created and linked to tms_TaskProjectRelationship.

I'll remove that and update the uid for cicd_Release.

Side note...Is there anything I can do about the errant CLA user? I goofed with my first commit using my default git profile for work so it didn't link the commit to my GitHub user account correctly.

thomas-gerber (approved):
Thank you @BrentSouza , this was an impressive contribution that went deep into the nuance of the data model!

thomas-gerber:
Bypassing the CLA assistant as the 2 committers are the same person.

thomas-gerber:
@BrentSouza released `v0.4.67` https://github.com/faros-ai/airbyte-connectors/releases/tag/v0.4.67

tovbinm:
Thank you for all the great work!! @BrentSouza

PR #893, opened 2023-01-27
FAI-4901: Hasura End-to-End Testing

ted-faros:
## Description

This PR fixes 2 bugs that present in similar ways but due to different codes paths have different solutions 

### SaaS Bug

In #862, we added a step to batch processing to separate objects into groups by the keys present in the objects.  We then constructed `on_conflict.update_columns` arrays based on the keys of each group.  This was done to fix a bug that nulled out fields of existing rows that were present in the `update_columns` array but not present in the object.

The (flawed) procedure for doing this was roughly:

1. for each batch separate objects into groups by keys of objects
2. add FKs to objects from prior batches
3. using first object of a group as prototype, construct `update_columns` for the group.

This procedure was flawed because the 2nd step could alter keys of a subset of group.  Said another way, not all objects in the group are guaranteed to have the same set of FKs added.  In the specific case we encountered, the first object used for computing `update_columns` did not have data for a particular FK.  A subsequent object did and since the `update_columns` didn't include the FK column, that object's FK was not upsert'ed into the database.

The fix is to reorder steps 1 & 2 so we add all possible data to the objects prior to grouping by keys.  

### CE Bug

WIP

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -158,16 +158,14 @@ export function batchIterator<T>(
 }
 
 /**
- * Separates each level (i.e. Upsert[]) into arrays of upserts
- * that all operate on the same fields.
+ * Separates objects into arrays of objects
+ * that all have the same keys.
  */
-export function groupByAffectedFields(levels: Upsert[][]): Upsert[][] {
-  return flatMap(levels, (level) => {
-    const byObjKeyNames: Dictionary<Upsert[]> = groupBy(level, (u) =>
-      Object.keys(u.object).sort().join('|')
-    );
-    return Object.values(byObjKeyNames);
-  });
+export function groupByKeys(objects: any[]): any[][] {
+  const byObjKeyNames: Dictionary<any[]> = groupBy(objects, (obj) =>
+    Object.keys(obj).sort().join('|')
+  );
+  return Object.values(byObjKeyNames);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -411,18 +409,20 @@ export class GraphQLClient {
...
+            const ops = this.toUpsertOps(model, batch);
+            return Promise.all(ops.map((op) => this.execUpsert(model, op)));
           }
         );
         try {
-          for await (const result of iterator) {
-            this.logger.debug(
-              `executed ${model} upsert with ${result.numObjects} object(s) in ${result.durationMs}ms`
-            );
+          for await (const results of iterator) {
```
tovbinm, line 420: Let's check if debug is enabled to avoid unnecessary iteration over the objects

tovbinm (approved):

tovbinm:
Can you please update the description with what is the actual bug fix.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: I don't see a method for checking levels on `AirbyteLogger`.  Regardless, there will only be 1 or 2 objects in this collection and one collection per model per flush.  It is not a record-level loop.  Maybe it gets called hundreds of times per invocation of the destination.

ted-faros:
> Can you please update the description with what is the actual bug fix.

Done.  Still working on CE validation but please take a look at the description and tell me if it makes sense.

ypc-faros (approved):

PR #897, opened 2023-01-30
fix: update json schema incidents for datadog

jaorr95:
## Description

When we tried to get incidents from datadog,  always throw the following error

image

Also, this json schema was malformed because it had a duplicate `properties` key

After update json schema for incidenst, it works.
> 

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/datadog-source/resources/schemas/incident.json
+++ b/sources/datadog-source/resources/schemas/incident.json
@@ -1,7 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
-  "required": ["id"],
   "properties": {
     "attributes": {
       "type": "object",
@@ -141,10 +140,8 @@
         "lastModifiedByUser": {
           "type": "object",
           "properties": {
-            "properties": {
-              "id": {
-                "type": "string"
-              }
+            "id": {
+              "type": "string"
             }
           }
         }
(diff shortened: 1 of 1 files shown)


sources/datadog-source/resources/schemas/incident.json
```diff
@@ -141,10 +140,8 @@
         "lastModifiedByUser": {
           "type": "object",
           "properties": {
-            "properties": {
```
ypc-faros, line 144: Good catch!

sources/datadog-source/resources/schemas/incident.json
```diff
@@ -1,7 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
   "type": "object",
-  "required": ["id"],
```
ypc-faros, line 4: I don't think this should be removed, given that it is the [primary key](https://github.com/faros-ai/airbyte-connectors/blob/main/sources/datadog-source/src/streams/incidents.ts#L28) for the stream.

Can you give it a try with only the duplicate `properties` removal change?

jaorr95: Done, i try only deleting the duplicate `properties` and it works.

PR #911, opened 2023-02-06
Add example Hygieia converter

ypc-faros:
## Description

Add example [Hygieia](https://github.com/hygieia/hygieia) converter. Converts records from the `collector_items` stream of the `airbyte/source-mongodb-v2` source.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


tovbinm (approved):

tovbinm:
@ypc-faros are you planning to add more converters? Does it work with Airbyte source for MongoDB - https://docs.airbyte.com/integrations/sources/mongodb-v2?

ypc-faros:
> @ypc-faros are you planning to add more converters? Does it work with Airbyte source for MongoDB - https://docs.airbyte.com/integrations/sources/mongodb-v2?

No plans to add more converters for now. This is meant to be used as an example. And yes, this would work for folks with a Hygieia deployment. They would pull from their MongoDB instance using the source you mention.

PR #920, opened 2023-02-15
Remove model & origin in ServiceNow destination + make sure we use the correct values in GraphQL writer

tovbinm:
## Description

Remove model & origin in ServiceNow destination + make sure we use the correct values for origin, model and operation in GraphQL writer

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change


ted-faros (approved):
Ah, I see now that I led @ypc-faros astray with my comment about adding a `model` field.  Thanks for the cleanup, @tovbinm.

PR #931, opened 2023-02-24
Split source specific configs spec

ypc-faros:


tovbinm:
Lgtm, if it's not a breaking change

tovbinm (approved):

ypc-faros:
> Lgtm, if it's not a breaking change

Not breaking 👍

PR #933, opened 2023-02-27
Fix shouldDeleteRecord

ypc-faros:
## Description

Handle the case where the incident wasn't processed at all.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


willmarks (approved):

willmarks:
@ypc-faros I think we may have gotten this backwards. I think we __don't__ want to delete an `ims_IncidentApplicationImpact` if we didn't process the incident at all. Meaning its from some historical incident that is in the graph and hasn't been update.

PR #935, opened 2023-03-01
Honor default branch_pattern

ypc-faros:
## Description

Honor [default](https://github.com/faros-ai/airbyte-connectors/blob/4a42b2d942d3270b5c9ef385d2a09155be20027a/sources/azure-repos-source/resources/spec.json#L31) `branch_pattern`

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/azure-repos-source/src/azure-repos.ts
+++ b/sources/azure-repos-source/src/azure-repos.ts
@@ -32,6 +32,7 @@ import {
 } from './models';
 
 const DEFAULT_API_VERSION = '7.0';
+export const DEFAULT_BRANCH_PATTERN = '^main$';
 const DEFAULT_GRAPH_VERSION = '7.1-preview.1';
 export const DEFAULT_PAGE_SIZE = 100;
 export const DEFAULT_REQUEST_TIMEOUT = 60000;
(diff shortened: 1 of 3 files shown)


sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -133,9 +134,7 @@ export class AzureRepos {
     axiosRetry(httpClient, retryConfig);
     axiosRetry(graphClient, retryConfig);
 
-    const branchPattern = config.branch_pattern
-      ? new RegExp(config.branch_pattern)
-      : undefined;
+    const branchPattern = new RegExp(config.branch_pattern || DEFAULT_BRANCH_PATTERN);
```
tovbinm, line 137: Assuming one wants to sync all branches, how would one does?

ypc-faros: One could use`branch_pattern` => `.*`
I'll add that to the examples. Wdyt about making that (all branches) the default?

tovbinm: then let's note it in the description in spec.json, e.g - https://github.com/faros-ai/airbyte-connectors/blob/main/sources/azure-repos-source/resources/spec.json#L30
```
"branch_pattern": {
        "type": "string",
        "title": "Branch Name Pattern",
        "description": "Regex pattern to match branch names. If you would like to match all branches use '.*'",
        "default": "^main$"
      },
```

tovbinm (approved):

PR #956, opened 2023-03-12
FAI-5403: Mock-data feed sync write failure to V2 graph

ted-faros:
## Description

Ensure null value replacement only happens for nullish keys.  

The use of `||` as a coalesce operator by itself was OK because we map keys to strings and strings to keys consistently.  

However, in the case of the `mock-data` feed we are inconsistent about the type of the `number` field for different model types.  

For example, notice the different types for `number` field in this record:

```
{
  "record": {
    "stream": "myfarosfeedssrc__faros_feeds__faros_feed",
    "emitted_at": 1678482135607,
    "data": {
      "vcs_PullRequestComment": {
        "pullRequest": {
          "number": 0,
          "repository": {
            "name": "solaris",
            "language": "javascript",
            "mainBranch": "main",
            "organization": {
              "uid": "faros-ai",
              "source": "Mock"
            }
          }
        },
        "number": "0",
        "createdAt": "2023-02-26T17:44:21.846Z",
        "author": {
          "uid": "sirius",
          "source": "GitHub"
        }
      }
    }
  },
  "type": "RECORD"
} 
```

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -122,7 +122,7 @@ export function serialize(obj: Dictionary<number | string>): string {
  */
 export function strictPick(obj: any, keys: string[], nullValue = 'null'): any {
   return keys.reduce(
-    (result, key) => set(result, key, obj[key] || nullValue),
+    (result, key) => set(result, key, obj[key] ?? nullValue),
     {}
   );
 }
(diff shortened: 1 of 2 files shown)


ted-faros:
@crupley if you actively working on mock-data feed, you might consider ensuring that number field is alway represented as a numeric.  See the description of this PR for more details.

tovbinm (approved):

crupley:
@ted-faros I only see one place where we generate `vcs_PullRequestComment` in mock data and [it is a BigInt](https://github.com/faros-ai/feeds/blob/master/feeds/mock-data-feed/src/dora.ts#L447). Could the string have come from some place else?

ted-faros:
@crupley does this differ from, say, `vcs_PullRequest`?  I'm wondering if `bigint` is being serialized as a string to avoid overflow.

crupley:
yes, `vcs_PullRequest` is just a number not BigInt

PR #962, opened 2023-03-16
[FAI-5472] Use highest historical severity for status page

nat-casey:
Look through incident's historical updates and grab highest severity from those. This will fix the issue of incidents all having sev5 because they are eventually set back to operational once fixed.

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
+++ b/destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
@@ -55,14 +55,11 @@ export class Incidents extends StatuspageConverter {
     // Use highest severity in incident's history as its severity
     let severity: IncidentSeverity | undefined = undefined;
 
-    for (const update of incident.incident_updates) {
-      if (update.affected_components) {
-        for (const component of update.affected_components) {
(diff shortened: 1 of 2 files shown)


destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
```diff
@@ -51,12 +51,38 @@ export class Incidents extends StatuspageConverter {
...
+            severity = thisSeverity;
+        }
+      }
+      res.push({
+        model: 'ims_IncidentEvent',
+        record: {
+          uid: update.id,
+          type: this.eventType(update.status),
+          createdAt: Utils.toDate(update.created_at),
+          detail: update.body,
```
tovbinm, line 73: should we truncate the `update.body` ?

nat-casey: maybe we should - I moved this chunk of code around I didn't add it so we've been pulling the body not-truncated and it hasn't been an issue so far. I checked where we truncate other fields and it looks like we usually use a config setting for amount we truncate. Since we have to get this out for smartbear I'm not going to add that change in this PR.

PR #967, opened 2023-03-23
Azure sources should all require unaltered access tokens

willmarks:
## Description

Azure sources should require unaltered access tokens

## Type of change
- [ ] Bug fix
- [X] New feature
- [X] Breaking change

### DIFF
--- a/sources/azure-repos-source/src/azure-repos.ts
+++ b/sources/azure-repos-source/src/azure-repos.ts
@@ -89,7 +89,7 @@ export class AzureRepos {
 
     let accessToken = config.access_token;
     if (!isBase64Encoded(accessToken)) {
-      accessToken = base64Encode(`${':'}${accessToken}`);
+      accessToken = base64Encode(`:${accessToken}`);
     }
 
     const httpClient = axios.create({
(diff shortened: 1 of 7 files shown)


sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -82,26 +87,26 @@ export class AzureRepos {
       throw new VError('Projects provided in addition to * keyword');
     }
 
-    const base64EncodedAccessToken = Buffer.from(
-      `${':'}${config.access_token}`,
-      'binary'
-    ).toString('base64');
+    let accessToken = config.access_token;
+    if (!isBase64Encoded(accessToken)) {
+      accessToken = base64Encode(`:${accessToken}`);
```
tovbinm, line 92: why not move the `:` to where you create the token

tovbinm, line 92: why not move the : to where you create the header?

willmarks: It needs to be part of the base64 encoded string.

tovbinm: oh, whoah, sucks!

tovbinm: how people are gonna even use it?

tovbinm (approved):

PR #968, opened 2023-03-23
FAI-5663 - Allow excluding specific model record fields in Faros Destination

tovbinm:
## Description

Allow excluding records fields by model in Faros Destination

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/destination.ts
+++ b/destinations/airbyte-faros-destination/src/destination.ts
@@ -24,7 +24,7 @@ import {
 } from 'faros-js-client';
 import http from 'http';
 import https from 'https';
-import {difference, keyBy, sortBy, uniq} from 'lodash';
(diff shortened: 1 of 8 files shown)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -420,10 +420,11 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
       const streamContext = new StreamContext(
         this.logger,
         config,
-        this.edition === Edition.COMMUNITY ? undefined : this.getFarosClient(),
+        streamsSyncMode,
+        JSON.parse(config.exclude_fields_map ?? '{}'),
```
ypc-faros, line 424: nit: I don't think we need this in the stream context. It can be just a property of `FarosDestination`

ypc-faros (approved):

PR #980, opened 2023-04-10
FAI-5450 - Application Impact startedAt & endedAt for downtime

willmarks:
## Description

Pull impact start and end for downtime calculation. Depends on a canonical-model change.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
+++ b/destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
@@ -4,6 +4,8 @@ import {Utils} from 'faros-js-client';
 import {Common} from '../common/common';
 import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
 import {
+  ApplicationImpact,
+  ApplicationImpactCategory,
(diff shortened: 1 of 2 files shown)


destinations/airbyte-faros-destination/src/converters/statuspage/incidents.ts
```diff
@@ -43,12 +43,21 @@ export class Incidents extends StatuspageConverter {
...
         acknowledgedAt = eventTime;
       }
+
+      if (update.status === StatuspageIncidentStatus.Monitoring) {
+        // The last monitoring status is when the service impact ended
+        impactEndedAt =
+          !impactEndedAt || impactEndedAt < eventTime
+            ? eventTime
+            : impactEndedAt;
+      }
```
willmarks, line 60: Does this make sense? If you switch to monitoring and then resolve the incident after that (that is why we want the last time it was moved to monitoring), then monitoring means that a potential fix has been implemented and the impact should have been resolved.

tovbinm (approved):

PR #986, opened 2023-04-14
Make it possible to output a JSON config instead of Airbyte local CLI args

ypc-faros:
Screenshot 2023-04-14 at 1 50 39 PM


eskrm (approved):
👍

ypc-faros:
Thank you @eskrm 🙏 !

PR #987, opened 2023-04-17
Handle Pagerduty pagination 10000 limit

willmarks:
## Description

pdjs (under the hood `cross-fetch`) does not reject 4xx responses. We need to check for them in non exceptional processing.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/pagerduty-source/src/pagerduty.ts
+++ b/sources/pagerduty-source/src/pagerduty.ts
@@ -172,6 +172,7 @@ export class Pagerduty {
     let fetchNextFunc;
 
     do {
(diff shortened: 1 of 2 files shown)


willmarks:
Btw, is the retry logic doing what we want it to in `errorWrapper`? Not very familiar with the pg js client / cross-fetch so Im not really sure when it will enter the catch statement in that function.

cjwooo (approved):

cjwooo:
@willmarks It definitely worked in the past, as we needed it to deal with flaky user network connections when running the connectors locally before.

PR #991, opened 2023-04-24
Fix pd pagination scenario 2

willmarks:
## Description

There is another way in which the API will response 400 when pagination exceeds 10,000 items. We need to handle that scenario as well.

```
Status code: 400: Bad Request. Data: {\"error\":{\"message\":\"Arguments Caused Error\",\"code\":2002,\"errors\":[\"Offset+limit exceeds maximum allowed value of 10000. Please try to refine your search to reduce the returned set of records below this number, like adding a date range.\"]}}"
```

I also noticed that we are including 2 `limit` params in our requests. We should not explicitly be setting `limit` but instead we should let the pd client set it based off of the configured page size (we are already setting this).

Notice the two limits:
```
https://api.pagerduty.com/incidents?limit=25&time_zone=UTC&since=2023-04-20T17%3A22%3A42.511%2B00%3A00&until=2023-04-21T17%3A22%3A42.511%2B00%3A00&limit=25&offset=10000
``` 

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/pagerduty-source/src/pagerduty.ts
+++ b/sources/pagerduty-source/src/pagerduty.ts
@@ -175,9 +175,7 @@ export class Pagerduty {
       // Deal with PagerDuty 10000 records response limit
       if (
         response?.status == 400 &&
-        (response?.data?.error?.errors?.[0]?.includes(
-          'Offset must be less than'
-        ) ||
+        (response?.data?.error?.errors?.[0]?.includes('Offset must be less') ||
           response?.data?.error?.errors?.[0]?.includes('Offset+limit exceeds'))
       ) {
         this.logger.warn(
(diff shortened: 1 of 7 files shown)


sources/pagerduty-source/src/pagerduty.ts
```diff
@@ -175,7 +175,8 @@ export class Pagerduty {
       // Deal with PagerDuty 10000 records response limit
       if (
         response?.status == 400 &&
-        response?.data?.error?.errors?.[0]?.includes('Offset must be less than')
+        (response?.data?.error?.errors?.[0]?.includes('Offset must be less') ||
```
tovbinm, line 178: perhaps we can just check for `Offset`?

tovbinm (approved):

PR #993, opened 2023-04-26
Octopus fix retry

willmarks:
## Description

axios-retry is not retrying 500s from octopus right now.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/octopus-source/src/octopusClient.ts
+++ b/sources/octopus-source/src/octopusClient.ts
@@ -1,5 +1,8 @@
 import axios, {AxiosInstance, AxiosResponse} from 'axios';
-import axiosRetry, {IAxiosRetryConfig, isRetryableError} from 'axios-retry';
+import axiosRetry, {
+  IAxiosRetryConfig,
+  isIdempotentRequestError,
+} from 'axios-retry';
 import {AirbyteLogger, wrapApiError} from 'faros-airbyte-cdk';
 import isRetryAllowed from 'is-retry-allowed';
 import {Memoize} from 'typescript-memoize';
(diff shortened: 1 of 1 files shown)


sources/octopus-source/src/octopusClient.ts
```diff
@@ -75,7 +72,7 @@ export class OctopusClient {
         shouldResetTimeout: true,
         retries,
         retryCondition: (error: Error): boolean => {
-          return isNetworkError(error) || isIdempotentRequestError(error);
+          return isNetworkError(error) || isRetryableError(error);
```
tovbinm, line 75: Please double check. I recall there was some nuance

tovbinm (approved):

sources/octopus-source/src/octopusClient.ts
willmarks:
```ts
const SAFE_HTTP_METHODS = ['get', 'head', 'options'];
const IDEMPOTENT_HTTP_METHODS = SAFE_HTTP_METHODS.concat(['put', 'delete']);

export function isRetryableError(error) {
  return (
    error.code !== 'ECONNABORTED' &&
    (!error.response || (error.response.status >= 500 && error.response.status <= 599))
  );
}

export function isSafeRequestError(error) {
  if (!error.config) {
    // Cannot determine if the request can be retried
    return false;
  }

  return isRetryableError(error) && SAFE_HTTP_METHODS.indexOf(error.config.method) !== -1;
}

export function isIdempotentRequestError(error) {
  if (!error.config) {
    // Cannot determine if the request can be retried
    return false;
  }

  return isRetryableError(error) && IDEMPOTENT_HTTP_METHODS.indexOf(error.config.method) !== -1;
}

```

`isRetriableError` does not check the method where you can see only the `IDEMPOTENT_HTTP_METHODS` are retried with `isIdempotentRequestError`

cjwooo (approved):

PR #1007, opened 2023-05-01
FAI-6073 - Octopus source improvements

willmarks:
## Description

Improvements to the Octopus source:
- In certain situations, a nonretryable 500 is responded from the octopus API when attempting to request a source controlled resource without a source controlled url. We should not retry these requests.
- Sometimes the project deployment process cannot be retrieved but the deployment specific process can be. We should fall back to the deployment specific process.
- Do not pull deployment process by default

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/octopus-source/src/octopusClient.ts
+++ b/sources/octopus-source/src/octopusClient.ts
@@ -1,4 +1,4 @@
-import axios, {AxiosInstance, AxiosResponse} from 'axios';
+import axios, {AxiosError, AxiosInstance, AxiosResponse} from 'axios';
 import axiosRetry, {
   IAxiosRetryConfig,
   isIdempotentRequestError,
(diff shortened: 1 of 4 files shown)


tovbinm (approved):

sources/octopus-source/src/octopusClient.ts
```diff
@@ -69,16 +68,30 @@ export class OctopusClient {
       );
     };
 
+    const isNonRetryable500 = (error: AxiosError): boolean => {
+      return (
+        error.response &&
+        error.response.status >= 500 &&
+        error.response.status <= 599 &&
+        error.response.data?.ErrorMessage?.includes(
+          'We received a request for a version-controlled resource'
```
tovbinm, line 77: this quite fragile

willmarks: Its only for a very specific case that I don't want to retry a 500.

PR #1012, opened 2023-05-04
FAI-6076 Honor full model deletion records coming from feeds

ypc-faros:
## Description

For Community Edition, we need to make sure the deletions are processed in topological order.

For Cloud V2, the deletion records coming from feeds are not valid (they look like `{"vcs_TeamMembership__Deletion":{"where":"my-source"}` with `my-source` being the origin)

We instead accumulate all the model _**names**_ for which we get a full model deletion record and reuse `resetData` which takes care of both problems for us.

No changes needed for Cloud V1.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
+++ b/destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
@@ -16,6 +16,8 @@ export class FarosFeed extends Converter {
   readonly destinationModels: ReadonlyArray<DestinationModel> = ALL_MODEL_NAMES;
 
   private schema: FarosGraphSchema = undefined;
+  private deleteModelEntries: string[] = [];
+  private resetModels: Set<string> = new Set();
 
   id(): any {
     return undefined;
@@ -42,12 +44,33 @@ export class FarosFeed extends Converter {
       ctx.config.edition_configs.edition === Edition.COMMUNITY ||
       ctx.config.edition_configs.graphql_api === 'v2'
     ) {
-      // Ignore full model deletion records.
+      // Full model deletion records.
       // E.g., {"vcs_TeamMembership__Deletion":{"where":"my-source"}}
       // These are issued by the feed and are only applicable to the V1 API
+      // We accumulate the model names in 'deleteModelEntries' to reset them in topological order
       if (model.endsWith('__Deletion') && Object.entries(rec).length == 1) {
         const [key, value] = Object.entries(rec).pop();
-        if (key === 'where' && typeof value == 'string') return [];
+        if (key === 'where' && typeof value == 'string') {
+          const [baseModel] = model.split('__', 1);
+          // Only reset once per model. Some feeds emit these multiple times (from root and workers)
+          if (!this.resetModels.has(baseModel)) {
+            this.resetModels.add(baseModel);
+            this.deleteModelEntries.push(baseModel);
+          }
+          return [];
+        }
+      } else {
+        if (this.deleteModelEntries.length > 0) {
+          if (ctx.resetData) {
+            await ctx.resetData(this.deleteModelEntries);
+          } else {
+            // This is okay when dry run is enabled
+            ctx.logger?.warn(
+              `Ignored full model deletion records for ${this.deleteModelEntries.join()}`
+            );
+          }
+          this.deleteModelEntries = [];
+        }
       }
     }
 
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
```diff
@@ -16,6 +16,8 @@ export class FarosFeed extends Converter {
   readonly destinationModels: ReadonlyArray<DestinationModel> = ALL_MODEL_NAMES;
 
   private schema: FarosGraphSchema = undefined;
+  private deleteModelEntries: string[] = [];
+  private resetModels: Set<string> = new Set();
```
ted-faros, line 20: I don't think you need 2 data structures.  Just add everything to the set and extract the array with `Array.from` when needed.

destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
```diff
@@ -42,12 +44,33 @@ export class FarosFeed extends Converter {
...
       if (model.endsWith('__Deletion') && Object.entries(rec).length == 1) {
         const [key, value] = Object.entries(rec).pop();
-        if (key === 'where' && typeof value == 'string') return [];
+        if (key === 'where' && typeof value == 'string') {
+          const [baseModel] = model.split('__', 1);
+          // Only reset once per model. Some feeds emit these multiple times (from root and workers)
+          if (!this.resetModels.has(baseModel)) {
+            this.resetModels.add(baseModel);
+            this.deleteModelEntries.push(baseModel);
+          }
```
ted-faros, line 59:
```suggestion
        this.resetModels.add(baseModel);
```

destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
```diff
@@ -42,12 +44,33 @@ export class FarosFeed extends Converter {
...
+          return [];
+        }
+      } else {
+        if (this.deleteModelEntries.length > 0) {
+          if (ctx.resetData) {
+            await ctx.resetData(this.deleteModelEntries);
```
ted-faros, line 65:
```suggestion
            await ctx.resetData(Array.from(this.resetModels));
```

destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
ypc-faros: I did it this way because I didn't want to assume too much on what models the root and workers feed reset. It's probably okay to assume they do reset the same set of models (that is the case for the handful of feeds I'm familiar with) in which case I'd be happy to incorporate your suggestion. 

@cjwooo any thoughts on this?

cjwooo: I agree with Ted's suggestion.

ypc-faros: Thank you @cjwooo @ted-faros 
I'll incorporate this feedback 👍

ted-faros: Kinda confused here @ypc-faros.  The only thing that changes when we go from array and set to set is the order.  And I thought the reset logic sorts topologically so the input order doesn't matter.  What am I missing?

ted-faros: Oh, I see.  The set guards further additions and since array is reset we only call resetData once.  If that is accurate, I like (2) since it is clear (to me 😄) what is going on.

destinations/airbyte-faros-destination/src/converters/faros-feeds/faros_feed.ts
```diff
@@ -42,12 +44,31 @@ export class FarosFeed extends Converter {
       ctx.config.edition_configs.edition === Edition.COMMUNITY ||
       ctx.config.edition_configs.graphql_api === 'v2'
     ) {
-      // Ignore full model deletion records.
+      // Full model deletion records.
       // E.g., {"vcs_TeamMembership__Deletion":{"where":"my-source"}}
       // These are issued by the feed and are only applicable to the V1 API
+      // We accumulate the model names in 'deleteModelEntries' to reset them in topological order
```
ted-faros, line 50:
```suggestion
      // We accumulate the model names in 'resetModels' to reset them in topological order
```

ted-faros (approved):
lgtm

PR #1015, opened 2023-05-15
FAI-6239 Flush writer and yield on every state message (V2 only)

ypc-faros:
Flush writer and yield on every state message (V2 only)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -488,7 +491,13 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
               origin,
               Array.from(streamContext.resetModels)
             )
-        );
+        )) {
+          streamContext.resetModels = new Set(deleteModelEntries);
+          await graphQLClient.flush();
+          // Allow some time for the flush to complete
+          await Utils.sleep(15000);
```
ypc-faros, line 498: Move this to the source

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -488,7 +490,11 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
               origin,
               Array.from(streamContext.resetModels)
             )
-        );
+        )) {
+          streamContext.resetModels = new Set(deleteModelEntries);
```
tovbinm, line 494: what is this for?

tovbinm (approved):

PR #1031, opened 2023-06-01
Azure DevOps Pipeline improvements

ypc-faros:


sources/azurepipeline-source/resources/spec.json
```diff
@@ -7,41 +7,59 @@
     "required": [
       "access_token",
       "organization",
-      "project_names",
-      "cutoff_days"
+      "project_names"
```
cjwooo, line 10: You renamed this config to `projects`. Also, this rename will require an Alastor update as well, as its corresponding catalog entry currently refers to `project_names`.

destinations/airbyte-faros-destination/src/converters/azurepipeline/common.ts
```diff
@@ -68,6 +79,29 @@ export abstract class AzurePipelineConverter extends Converter {
...
+
+    const parts = repo.id.split('/');
+    // Expecting repo.id to be in the format of <org>/<repo>
+    // E.g., faros-ai/airbyte-connectors
+    if (parts.length < 2) {
+      return undefined;
+    }
+
+    return {
+      name: parts[1],
```
cjwooo, line 96: Lowercase the `vcs_Repository.name` and `vcs_Organization.uid` values to match behavior of other connectors. Also, `name` is not part of `vcs_Organization`'s key.

sources/azurepipeline-source/resources/spec.json
ypc-faros: Yes I have other changes in mind for the catalog. I'll include this one as well, thank you for the heads up 🙏

cjwooo (approved):
Just a couple comments

PR #1035, opened 2023-06-07
Datadog source: handle case where metrics unit is not provided

tovbinm:
## Description

Datadog source: handle case where metrics unit is not provided

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/datadog-source/src/datadog.ts
+++ b/sources/datadog-source/src/datadog.ts
@@ -28,7 +28,7 @@ export interface MetricPoint {
   metric: string;
   timestamp: number;
   value: number;
-  primaryUnit: v1.MetricsQueryUnit;
+  primaryUnit?: v1.MetricsQueryUnit;
   perUnit?: v1.MetricsQueryUnit;
   scope: string;
   tagSet: Array<string>;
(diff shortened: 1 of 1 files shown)


sources/datadog-source/src/datadog.ts
```diff
@@ -157,8 +157,14 @@ export class Datadog {
...
-            primaryUnit: metadata.unit[0],
-            perUnit: metadata.unit[1],
+            primaryUnit:
+              Array.isArray(metadata.unit) && metadata.unit.length > 0
+                ? metadata.unit[0]
+                : undefined,
+            perUnit:
+              Array.isArray(metadata.unit) && metadata.unit.length > 1
+                ? metadata.unit[1]
+                : undefined,
```
willmarks, line 167: Is `metadata.unit` only an array of 1 to 2 things? Or do we just not care about anything beyond primaryUnit and perUnit

willmarks (approved):

sources/datadog-source/src/datadog.ts
tovbinm: In my tests `unit` was always null, but in the test file we have this example - https://github.com/faros-ai/airbyte-connectors/blob/main/sources/datadog-source/test_files/metrics.json#L24

it's not really clear what is the primary nor secondary. I might be better to return `unit` as is (or add it as well)

tovbinm: from their docs
```
/**
     * Detailed information about the metric unit.
     * First element describes the "primary unit" (for example, `bytes` in `bytes per second`),
     * second describes the "per unit" (for example, `second` in `bytes per second`).
     */
    "unit"?: [MetricsQueryUnit, MetricsQueryUnit];
```

PR #1038, opened 2023-06-07
Faros Destination: added Datadog metrics stream converter

tovbinm:
## Description

Faros Destination: added Datadog metrics stream converter

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/datadog/metrics.ts
```diff
@@ -0,0 +1,67 @@
...
+    res.push(metricValue);
+
+    return res;
+  }
+
+  metricValueType(unit: Dictionary<any>): {category: string; detail: string} {
+    if (!unit || !unit.family) return {category: 'Custom', detail: 'unknown'};
+    const detail = toLower(unit.family);
+
+    // For more details see - https://docs.datadoghq.com/metrics/units/
```
tovbinm, line 54: @cjwooo @willmarks @nat-casey  please validate if there there are any other types we should support here?

PR #1047, opened 2023-06-09
Expand AirbyteStateMessage schema to include optional Source status info

cjwooo:
## Description

The goal here was to provide a way for an Airbyte Source to pass information to the Destination besides data records and incremental states, something that neither the Airbyte protocol nor Airbyte Server supports. Airbyte Server also only sends records and state messages from the Source to the Destination -- all other Airbyte message types, such as log messages, are not passed to the Destination. In addition, Airbyte Server does not stop or notify the Destination when the Source has a fatal error. This is intentional, as the Destination is supposed to finish processing all records up until the Source's error and then exit.

By expanding the AirbyteStateMessage schema, we enable Sources to signal our Faros Destination when a fatal error occurs. The Faros Destination will now skip the reset non-incremental models when such a signal is detected.

Scenario:
An Airbyte Source unexpectedly fails during a sync due to a breaking change in the system APIs that it fetches data from. The Source will signal to the Faros Destination that an error has occurred, and the Faros Destination will not reset any non-incremental models, so that the user's existing graph data does not suddenly disappear.

Ideally, this would be a completely new type of AirbyteMessage, but that would also require updates to Airbyte Server logic to handle this new message type, and we need a solution now. I spoke with one of the lead Airbyte devs and he seemed receptive to adding this kind of functionality upstream, so once we have time we can work towards that.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


ypc-faros (approved):

faros-airbyte-cdk/src/protocol.ts
```diff
@@ -257,11 +257,15 @@ export interface AirbyteState {
   [stream: string]: any;
 }
 
+export interface AirbyteSourceStatus {
+  status: 'RUNNING' | 'ERROR';
```
tovbinm, line 261: 'RUNNING' | 'ERRORED'

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -141,6 +141,10 @@ export abstract class AirbyteSourceBase<
           }`,
           e.stack
         );
+        yield new AirbyteStateMessage(
+          {data: connectorState},
+          {status: 'ERROR'}
```
tovbinm, line 146: let's do `{status: 'ERRORED', error: e.message ?? JSON.stringify(e)}`

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -707,7 +714,13 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
         );
       }
 
-      await resetData?.();
+      if (sourceFailed) {
+        this.logger.warn(
+          'Skipping reset of non-incremental models due to Airbyte Source failure'
```
tovbinm, line 719: assuming we propagate the error, we can then do: 
```
 this.logger.error(
     'Skipping reset of non-incremental models due to Airbyte Source failure: ${sourceFailed.error}'
)
```

faros-airbyte-cdk/src/protocol.ts
tovbinm: where do we yield `RUNNING` from? it's not used yet, right?

cjwooo: Yea not used yet.

PR #1049, opened 2023-06-09
FAI-6560: handle nil/empty array literals properly

ted-faros:
## Description

Fix handling of empty string by quoting strings and translating `nil` to `NULL`.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -210,6 +210,20 @@ function addLevel(u: Upsert): [Upsert, number][] {
   return result;
 }
 
+export function toPostgresArrayLiteral(value: any[]): string {
+  return `{${value
+    .map((s) => {
+      if (typeof s === 'string') {
+        return `"${s}"`;
+      }
+      if (isNil(s)) {
+        return 'NULL';
+      }
+      return s;
+    })
+    .join(',')}}`;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


ypc-faros (approved):

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -210,6 +210,20 @@ function addLevel(u: Upsert): [Upsert, number][] {
...
+export function toPostgresArrayLiteral(value: any[]): string {
+  return `{${value
+    .map((s) => {
+      if (typeof s === 'string') {
+        return `"${s}"`;
+      }
+      if (isNil(s)) {
+        return 'NULL';
+      }
+      return s;
```
willmarks, line 222: Do we have any object arrays that we would need to worry about?

ted-faros: object arrays would be represented by `jsonb` or `jsonb[]`

willmarks: What happens with something like this? https://github.com/faros-ai/canonical-models/blob/main/resources/models/qa.gql#LL26C33-L26C33

ted-faros: it is jsonb: https://github.com/faros-ai/hasura/blob/main/migrations/default/1674589115214_schema_v4/up.sql#L2544

PR #1055, opened 2023-06-21
FAI-6714: ensure upserts update nil values and relations

ted-faros:
## Description

Allow upserts of `nil` (i.e. null and undefined) values and relations.  Also, validate the `uid` fields are non-null while processing records.  This prevents a later db constraint violation that results in a full batch being discarded.

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -714,17 +714,32 @@ export class GraphQLClient {
     const foreignKeys: Dictionary<Upsert> = {};
     for (const [field, value] of Object.entries(record)) {
       const nestedModel = this.schema.references[model][field];
-      if (nestedModel && value) {
-        foreignKeys[field] = this.addUpsert(nestedModel.model, value);
-      } else {
+      if (nestedModel) {
+        if (isNil(value)) {
+          // for explicit nil relation, set the FK of this record to null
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -714,18 +714,33 @@ export class GraphQLClient {
...
       object['origin'] = origin;
       // add root flag to top-level objects
       object[ROOT_FLAG] = true;
     }
+    // since all our uids are non-null, check for nil uids early
+    // to prevent losing an entire batch later with db constraint
+    // error on flush
+    if (has(object, 'uid') && isNil(object['uid'])) {
+      throw new VError(
+        'cannot upsert null uid for model %s with keys %s',
```
tovbinm, line 739: `cannot upsert null or undefined uid for model %s with keys %s`

destinations/airbyte-faros-destination/test/graphql-client.test.ts
```diff
@@ -641,6 +641,103 @@ describe('graphql-client write batch upsert', () => {
     await client.flush();
     expect(queries).toEqual(responses.length);
   });
+  test('null fields bug', async () => {
```
tovbinm, line 644: `null fields bug` -> `allow upsert null values`

tovbinm (approved):

PR #1056, opened 2023-06-22
FAI-6714 - Convert undefined source record property values to nulls

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/faros-airbyte-cdk/src/logger.ts
+++ b/faros-airbyte-cdk/src/logger.ts
@@ -1,3 +1,4 @@
+import {isNil} from 'lodash';
 import pino, {DestinationStream, Level, Logger} from 'pino';
 import stream from 'stream';
 
@@ -86,6 +87,13 @@ export class AirbyteLogger {
       const msgLevelOrder = AirbyteLogLevelOrder((msg as AirbyteLog).log.level);
       if (levelOrder > msgLevelOrder) return;
     }
-    console.log(JSON.stringify(msg));
+
+    // convert undefined record values to nulls so they are stringified
+    const replacer =
+      msg.type === AirbyteMessageType.RECORD
+        ? (key, value) => (isNil(value) ? null : value)
+        : null;
+
+    console.log(JSON.stringify(msg, replacer));
   }
 }
--- a/faros-airbyte-cdk/test/logger.test.ts
+++ b/faros-airbyte-cdk/test/logger.test.ts
@@ -31,15 +31,14 @@ describe('logger', () => {
       number: 1,
       string: 'string',
       array: [1, 'string', {a: 1}],
-      object: {a: 1, b: 'string'},
       null: null,
       undefined: undefined,
-      objectWithUndefined: {a: 1, null: null, undefined: undefined},
+      object: {a: 1, b: 'string', null: null, undefined: undefined},
     };
     const expectedData = {
       ...data,
       undefined: null,
-      objectWithUndefined: {a: 1, null: null, undefined: null},
+      object: {a: 1, b: 'string', null: null, undefined: null},
     };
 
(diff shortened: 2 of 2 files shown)


ted-faros (approved):
💥

faros-airbyte-cdk/test/logger.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  test('converts undefined record fields to nulls', async () => {
+    const data = {
+      boolean: true,
+      number: 1,
+      string: 'string',
+      array: [1, 'string', {a: 1}],
+      null: null,
+      undefined: undefined,
+      object: {a: 1, b: 'string', null: null, undefined: undefined},
```
tovbinm, line 36: ehh, `null: null, undefined: undefined` is weird! we really only care about values for non-empty keys

faros-airbyte-cdk/src/logger.ts
```diff
@@ -86,6 +87,13 @@ export class AirbyteLogger {
       const msgLevelOrder = AirbyteLogLevelOrder((msg as AirbyteLog).log.level);
       if (levelOrder > msgLevelOrder) return;
     }
-    console.log(JSON.stringify(msg));
+
+    // convert undefined record values to nulls so they are stringified
+    const replacer =
+      msg.type === AirbyteMessageType.RECORD
```
tovbinm, line 93: we can also skip **keys**  which are empty, null or undefined

tovbinm (approved):
lgtm, see comments

faros-airbyte-cdk/test/logger.test.ts
cjwooo: So it's ok to have a record field value be an object with undefined values? They will not be passed to the destination.

faros-airbyte-cdk/src/logger.ts
cjwooo: From ChatGPT:

No, it is not possible for a JavaScript object to have a null or undefined key name. In JavaScript, object keys must be of type string or symbol.

If you attempt to use null or undefined as a key name when creating an object, they will be implicitly converted to strings.

Here's an example:

```
const obj = {};

obj[null] = 'Value for null key';
obj[undefined] = 'Value for undefined key';

console.log(obj);
```

In this example, when assigning values to null and undefined key names, JavaScript implicitly converts them to strings. The resulting object will have string keys named "null" and "undefined".

The output of console.log(obj) will be:
```
{ null: 'Value for null key', undefined: 'Value for undefined key' }
```
As you can see, the null and undefined keys are treated as strings in the resulting object.

ted-faros: no, please don't filter the key.  I like the PR as-is.

faros-airbyte-cdk/test/logger.test.ts
cjwooo: Actually I think you're misinterpreting. The **keys** in `{null: null, undefined: undefined}`, are the strings `"null"` and `"undefined"` respectively. The **values** are `null` and `undefined`.

cjwooo: I've updated the test to make this clearer.

faros-airbyte-cdk/src/logger.ts
tovbinm: Let's just fix the tests at least. Null / undefined keys are confusing.

tovbinm: You did it already. Thanks.

tovbinm (approved):

PR #1065, opened 2023-06-26
FAI-6617 - Enhancing Octopus Deployments with Variable Sets

omreego:
## Description
This PR adds the Deployment Variable Set to the Streamed Deployment Object

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/octopus-source/resources/spec.json
```diff
@@ -25,6 +25,14 @@
           "type": "string"
         }
       },
+      "variable_names": {
+        "type": "array",
+        "title": "Octopus Deployment Variable to Sync",
```
tovbinm, line 30: `Variables` -> `Variables`

sources/octopus-source/resources/spec.json
```diff
@@ -25,6 +25,14 @@
...
+        "description": "The Octopus Deployment Variables to sync.",
```
tovbinm, line 31: Any additional details you can provide for the users?

tovbinm: Does this get the variable values? Can these be sensitive? We should warn about this.

sources/octopus-source/src/octopusClient.ts
```diff
@@ -145,6 +147,20 @@ export class OctopusClient {
...
+    spaceId: string,
+    variableSetId: string
+  ): Promise<OctopusDeploymentVariable[] | undefined> {
+    try {
+      const res: VariableSetResponse = await this.get<VariableSetResponse>(
+        `/${spaceId}/variables/${variableSetId}`
+      );
+      return res.Variables;
+    } catch (err: any) {
+      return undefined;
```
tovbinm, line 160: Should we at least log the error?

tovbinm (approved):

sources/octopus-source/src/octopus.ts
```diff
@@ -151,15 +157,20 @@ export class Octopus {
...
+        const [project, environment, task, process, variables] =
+          await Promise.all([
+            this.client.getProject(deployment.ProjectId),
+            this.client.getEnvironment(deployment.EnvironmentId),
+            this.client.getTask(deployment.TaskId),
+            this.getDeploymentProcess(
+              deployment.ProjectId,
+              deployment.DeploymentProcessId
+            ),
+            this.getDeploymentVariables(
```
tovbinm, line 169: Let's avoid calling this API completely if no env deployment variable names were provided

tovbinm (approved):

sources/octopus-source/resources/schemas/deployments.json
```diff
@@ -123,6 +123,43 @@
             }
           }
         }
+      },
+      "Variables": {
+        "type": "array",
+        "items": [
```
willmarks, line 129: I believe this should just be an object, not an array of objects.

PR #1092, opened 2023-07-25
 FAI-6951: GHSA-g644-9gfx-q4q4 - update vm2 dependency

ted-faros:
## Description

Upgrade `jest` and `eslint` to address moderate vulns in `word-wrap` and `tough-cookie`

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/package.json
```diff
@@ -69,6 +69,10 @@
       "ts-jest": {
         "tsconfig": "test/tsconfig.json"
       }
+    },
+    "snapshotFormat": {
+      "printBasicPrototype": true,
+      "escapeString": true
```
alexisschwartz, line 75: ooc, why is `"escapeString": true` changed only in two of these?

alexisschwartz (approved):

destinations/airbyte-faros-destination/package.json
ted-faros: v28 of jest made breaking change to format of snapshots.  This overrides [new default](https://jestjs.io/docs/configuration#snapshotformat-object) of `false` so that our snapshots are grandfathered in.

PR #1098, opened 2023-08-04
FAI-6879: Enable option to disable syncing uptimes for components

chalenge:
## Description

- Enable the option to disable syncing uptimes for components
- Ensure we only try to fetch a max of 90 days
- Fix uptime UID

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/.github/workflows/skip-test-sources.txt
+++ b/.github/workflows/skip-test-sources.txt
@@ -2,6 +2,7 @@ agileaccelerator-source
 azure-repos-source
 azure-workitems-source
 azurepipeline-source
+azureactivedirectory-source
 bamboohr-source
 bitbucket-server-source
 datadog-source
--- a/sources/statuspage-source/src/statuspage.ts
+++ b/sources/statuspage-source/src/statuspage.ts
@@ -24,6 +24,7 @@ export interface StatuspageConfig {
   readonly page_ids?: ReadonlyArray<string>;
   readonly max_retries?: number;
   readonly page_size?: number;
+  readonly fetch_component_uptime?: boolean;
 }
 
 export class Statuspage {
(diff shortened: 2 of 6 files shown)


sources/statuspage-source/src/statuspage.ts
```diff
@@ -219,8 +220,13 @@ export class Statuspage {
       return `${year}-${month}-${day}`;
     };
 
-    const rangeStart =
-      rangeEndDate > this.startDate ? rangeEndDate : this.startDate;
+    // Can only get uptimes for the last 90 days.
+    const maxRangeStart = new Date();
+    maxRangeStart.setDate(maxRangeStart.getDate() - 90);
```
tovbinm, line 225: I think you can do `maxRangeStart.setDate(-90)` see examples - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate#description

sources/statuspage-source/src/streams/component_uptime.ts
```diff
@@ -25,6 +25,12 @@ export class ComponentUptimes extends StatuspageStreamBase {
   }
 
   async *streamSlices(): AsyncGenerator<StreamSlice> {
+    if (!this.cfg.fetch_component_uptime) {
```
tovbinm, line 28: let remove this log line. I am not sure why is this useful, or make it a debug

tovbinm (approved):

sources/statuspage-source/src/statuspage.ts
chalenge: No, this will be incorrect. setDate(0) gives you the last day of the previous month and then starts decreasing from there. e.g.
```typescript
const date = new Date()
date.setDate(-1)
console.log(`Date ${date}`) // "Date 1 Sun Jul 30 2023 12:37:19 GMT-0400 (Eastern Daylight Time)

const date2 = new Date()
date2.setDate(date2.getDate() - 1);
console.log(`Date 2 ${date2}`) // Date 2 Thu Aug 03 2023 12:37:19 GMT-0400 (Eastern Daylight Time)
```

tovbinm: perhaps just use `moment` package?

chalenge: Normally but not worth adding another package for a two lines of code.

.github/workflows/skip-test-sources.txt
```diff
@@ -2,6 +2,7 @@ agileaccelerator-source
 azure-repos-source
 azure-workitems-source
 azurepipeline-source
+azureactivedirectory-source
```
chalenge, line 5: https://faros-ai.atlassian.net/browse/FAI-7227

PR #1103, opened 2023-08-09
FAI-6171 Adds support to skip fields marked as deprecated from the wizard

ypc-faros:


tovbinm (approved):

ypc-faros:
Thank you @tovbinm 🙏 !

PR #1127, opened 2023-08-30
Workday Source: add support for basic authentication using username/password

omreego:
## Description
Workday Source: add support for basic authentication using username/password

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/workday-source/resources/spec.json
```diff
@@ -55,7 +55,17 @@
         "minimum": 500,
         "default": 60000,
         "description": "The time allowed for a request to timeout (in milliseconds)."
-      }
+      },
```
tovbinm, line 58: @cjwooo how can we make this conditional - either u/p auth OR clientId/clientSecret/refreshToken?

sources/workday-source/src/workday.ts
```diff
@@ -54,6 +56,18 @@ export class Workday {
     if (!cfg.baseUrl) {
       throw new VError('baseUrl must not be an empty string');
     }
+    if (cfg.customReportName) {
```
tovbinm, line 59: this is invalid. we should allow supporting either auth method.

tovbinm: this applies for all APIs.

tovbinm (approved):

PR #1144, opened 2023-09-08
removed failed check connection test

omreego:
## Description
Connection check for Workday connector not necessary for custom reports


omreego:
Not off correct base

PR #1145, opened 2023-09-08
Allow skipping connection check for Workday source

omreego:
We don't want to check for workers, so we have a skipWorkerCheck boolean.

### DIFF
--- a/sources/workday-source/resources/spec.json
+++ b/sources/workday-source/resources/spec.json
@@ -108,6 +108,11 @@
                 "title": "Request Timeout",
                 "default": 60000,
                 "description": "The time allowed for a request to timeout (in milliseconds)."
+            },
+            "skipWorkerCheck": {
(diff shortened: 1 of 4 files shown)


sources/workday-source/resources/spec.json
```diff
@@ -108,6 +108,11 @@
                 "title": "Request Timeout",
                 "default": 60000,
                 "description": "The time allowed for a request to timeout (in milliseconds)."
+            },
+            "skipWorkerCheck": {
```
tovbinm, line 112: rename to `skipConnectionCheck` and update the description accordingly, .i.e `Skip Connection Check`

tovbinm:
Superseded by https://github.com/faros-ai/airbyte-connectors/commit/bb53fe6185b731cb764d6aa3610a2d01184aca21

PR #1146, opened 2023-09-11
skip Worker Check

omreego:
## Description

> Provide description here

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/workday-source/resources/spec.json
+++ b/sources/workday-source/resources/spec.json
@@ -108,6 +108,11 @@
                 "title": "Request Timeout",
                 "default": 60000,
                 "description": "The time allowed for a request to timeout (in milliseconds)."
+            },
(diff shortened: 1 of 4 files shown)


sources/workday-source/resources/spec.json
```diff
@@ -109,9 +109,11 @@
                 "default": 60000,
                 "description": "The time allowed for a request to timeout (in milliseconds)."
             },
-            "skipConnectionCheck": {
+            "skipWorkerCheck": {
```
tovbinm, line 112: Let's keep the original `skipConnectionCheck` and `Skip Connection Check`

PR #1149, opened 2023-09-20
FAI-7388 Support airbyte_hidden

ypc-faros:
## Description

Adds support for `airbyte_hidden` https://docs.airbyte.com/connector-development/connector-specification-reference/#hiding-inputs-in-the-ui

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/faros-airbyte-cdk/src/destinations/destination-runner.ts
+++ b/faros-airbyte-cdk/src/destinations/destination-runner.ts
@@ -140,9 +140,15 @@ export class AirbyteDestinationRunner<
       .option(
         '--spec-file <path to spec>',
         'Path to the spec file. If not provided, the spec will be fetched from the destination'
-      ).option(
+      )
+      .option(
         '--include-deprecated-fields',
         'Include fields marked as deprecated in the spec',
+        false
+      )
+      .option(
+        '--include-hidden-fields',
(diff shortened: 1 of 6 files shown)


faros-airbyte-cdk/src/destinations/destination-runner.ts
```diff
@@ -140,11 +140,17 @@ export class AirbyteDestinationRunner<
...
+      )
+      .option(
         '--include-deprecated-fields',
         'Include fields marked as deprecated in the spec',
         false
       )
+      .option(
+        '--include-hidden-fields',
+        'Include fields marked as hidden in the spec',
+        false
```
tovbinm, line 152: why not make it default to true?

tovbinm (approved):

faros-airbyte-cdk/src/destinations/destination-runner.ts
ypc-faros: We want to have the same behavior as the Airbyte UI. These fields should be accepted by the connector but not necessarily exposed to the user (unless they explicitly ask for it via `--include-hidden-fields`)

PR #1166, opened 2023-10-02
FAI-7679 - Dedup jobs shared by workflows

willmarks:
## Description

Figured it out the root cause finally. Jobs are at the pipeline level and workflows can share jobs if a workflow is executed from the failed step. It will point to the same jobs up to the point of the failure and then create new jobs for the remained.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/circleci-source/src/streams/tests.ts
+++ b/sources/circleci-source/src/streams/tests.ts
@@ -40,23 +40,23 @@ export class Tests extends CircleCIStreamBase {
       streamSlice.projectName,
       since
     )) {
-      const seenJobs = new Set<string>();
+      const seenJobs = new Set<number>();
       for (const workflow of pipeline.workflows ?? []) {
         for (const job of workflow.jobs ?? []) {
-          if (job.job_number === undefined) {
+          const jobNum = job.job_number;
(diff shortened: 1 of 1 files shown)


sources/circleci-source/src/streams/tests.ts
```diff
@@ -40,24 +40,23 @@ export class Tests extends CircleCIStreamBase {
       streamSlice.projectName,
       since
     )) {
+      const seenJobs = new Set<string>();
```
willmarks, line 43: This is the key. It has to be per pipeline not per workflow.

sources/circleci-source/src/streams/tests.ts
```diff
@@ -40,9 +40,8 @@ export class Tests extends CircleCIStreamBase {
       streamSlice.projectName,
       since
     )) {
+      const seenJobs = new Set<number>();
```
willmarks, line 43: This is the key. It has to be per pipeline not per workflow.

ypc-faros (approved):

tovbinm:
Major kudos.

PR #1168, opened 2023-10-04
FAI-7888 Honor `replace_origin_map` default + change `accept_input_records_origin` default to `true`

ypc-faros:


destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -37,7 +37,7 @@
         "type": "boolean",
         "title": "Accept input records origin",
         "description": "Whether to accept the origin field (if present) of input records. If enabled, the input records origin will only be changed according to the replacements specified by 'Replace Records Origin'. Otherwise, the input records origin are ignored and overwritten.",
-        "default": false
+        "default": true
```
tovbinm, line 40: why are we making this change?

ypc-faros: Discussed internally. We agreed this is a more appropriate default.

tovbinm (approved):

PR #1178, opened 2023-10-16
FAI-7999: Support passing component group when fetching component uptimes

chalenge:
## Description

- Support passing component group when fetching component uptimes
- Fixes an issue where how we are writing application platform name in uptimes doesn't match how we are writing application from components as [we prepend the group name](https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/src/converters/statuspage/common.ts#L188)

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


matiaslcoulougian (approved):
LGTM

ypc-faros (approved):

sources/statuspage-source/test/index.test.ts
```diff
@@ -493,7 +509,51 @@ describe('index', () => {
         pageId: 'n3wb7hf336hn',
         componentId: 'component2',
         startDate: '2023-05-22',
+        componentGroupId: 'string',
```
ypc-faros, line 512: nit: Maybe use a better value than `string`

chalenge: Sadly this is coming from the test file JSON, so would have to change multiple things :-(

ypc-faros: No worries 👍

PR #1183, opened 2023-10-17
[FAI-8053] GraphDoctor Destination

omreego:
## Description
Creating GraphDoctor Destination.
Handles inputs from GraphDoctor Source.
Currently writes out to two Models.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/faros-graphdoctor/models.ts
```diff
@@ -0,0 +1,13 @@
+export interface SummaryKey {
+  uid: string;
+  source: string;
+}
+
+export interface DataQualityIssue {
+  uid: string;
+  description?: string;
+  created_at?: string;
+  recordIds?: [string];
```
cjwooo, line 10: `[string]` is the key type for an object, an array of strings is `string[]`

cjwooo (approved):

destinations/airbyte-faros-destination/src/converters/faros-graphdoctor/models.ts
omreego: Good catch, that was a mistake

PR #1184, opened 2023-10-18
[FAI-7803] Workday Destination

omreego:
## Description

Preparing workday destination converter.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


omreego:
The random number generators are in a file that creates a random test input. They aren't a concern because they aren't called during run time, only independently if one were to want to add a test

omreego:
Link to design spec:
https://docs.google.com/document/d/1765DPVG5nOv-TLBcx2FJud1MVxW6323ji6sNo0p3rPs/edit

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
+{
+    "title": "workday",
```
tovbinm, line 2: Workday

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+    "description": "Configuration options that apply to records generated by the workday Source.",
```
tovbinm, line 3: capitalize workday as Workday

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+    "oneOf": [
+        {
+            "type": "object",
+            "title": "Configuration",
+            "properties": {
+                "source_type": {
+                    "type": "string",
+                    "const": "workday",
+                    "order": 0
+                },
```
tovbinm, line 14: why is this needed?

omreego: I don't know, I saw this on all the other configs I checked so I copied their format

destinations/airbyte-faros-destination/src/converters/workday/models.ts
```diff
@@ -0,0 +1,16 @@
+export declare type EmployeeRecord = {
+  Start_Date: Date;
+  Full_Name: string;
+  Employee_ID: string;
+  Manager_Name: string;
```
ypc-faros, line 5: Should `Manager_Name` and `Manager_ID` be optional?

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+  private cycleChains: ReadonlyArray<string>[] = [];
+  FAROS_TEAM_ROOT = 'all_teams';
+
+  private replacedParentTeams: string[] = [];
+  private seenLocations: Set<string> = new Set<string>();
+  private generalLogCollection: string[] = [];
+
+  /** Almost every SquadCast record have id property */
+  id(record: AirbyteRecord): any {
+    return record?.record?.data?.Empoyee_Id;
```
ypc-faros, line 36: typo `Employee_ID`

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    const res: DestinationRecord[] = [];
+    const rec = record.record.data as EmployeeRecord;
+    if (!this.checkRecordValidity(rec)) {
+      this.recordCount.skippedRecords += 1;
+    } else {
+      this.recordCount.storedRecords += 1;
+      this.extractRecordInfo(rec);
+    }
+    return res;
```
ypc-faros, line 50: We can always `return []`. No need to declare `res`

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+    this.employeeIDtoRecord[rec.Employee_ID] = rec;
+    this.updateTeamToManagerRecord(rec);
+  }
+  private updateTeamToManagerRecord(rec: EmployeeRecord): void {
+    // We store all the possible Manager IDs for a Team,
+    // The last one in the list will be the most recent time.
+
+    // We check if the Team Name happens to be FAROS_TEAM_ROOT.
+    // If this is the case, then the rest of the processing won't work.
+    if (rec.Team_Name == this.FAROS_TEAM_ROOT) {
```
ypc-faros, line 65: Use `===` for string comparison

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+
+    // We have already stored info for this team. This should be a list
+    const recs_list = this.teamNameToManagerIDs[rec.Team_Name];
+
+    const crt_last_rec = recs_list[recs_list.length - 1];
+    if (crt_last_rec.Manager_ID === rec.Manager_ID) {
+      return;
+    }
+
+    if (
```
ypc-faros, line 87: If I'm understanding this logic correctly, I think we're assuming that the manager of a team is the manager of the last known employee to join a team, correct?

If so, why do we need to keep a list of managers+timestamp for each team. Doesn't it suffice to only keep the one with the freshest start date timestamp?

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+    {
+      "type": "object",
+      "title": "Configuration",
+      "properties": {
+        "source_type": {
+          "type": "string",
+          "const": "Workday",
+          "order": 0
+        },
+        "Orgs_To_Keep": {
```
ypc-faros, line 15: Please lowercase

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "title": "Orgs To Keep",
+          "description": "The teams underneath these orgs will be kept",
+          "default": []
+        },
+        "Orgs_To_Ignore": {
```
ypc-faros, line 24: Please lowercase

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
ypc-faros, line 1: You need to reference this file from https://github.com/faros-ai/airbyte-connectors/blob/main/destinations/airbyte-faros-destination/resources/source-specific-configs/spec.json

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+    const orgs_to_keep = ctx.config.source_specific_configs.Orgs_To_Keep;
+    const orgs_to_ignore = ctx.config.source_specific_configs.Orgs_To_Ignore;
+    if (!orgs_to_keep || !orgs_to_ignore) {
+      throw new Error(
+        'Orgs_To_Keep or Orgs_To_Ignore missing from source specific configs'
+      );
+    }
+    for (const org of orgs_to_keep) {
+      if (orgs_to_ignore.includes(org)) {
+        throw new Error('Overlap between Orgs_To_Keep and Orgs_To_Ignore');
```
ypc-faros, line 207: We need to perform the same check for `orgs_to_ignore`

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
ypc-faros, line 1: I might be oversimplifying but I think the logic to determine which teams to keep can be improved.

Basically for a given team, we go up until we hit a `org_to_keep` or `org_to_remove` node. Depending on what we hit we remove or keep. I'd just do that for every single team and store it somewhere, let's call that `nodesKept`.

Later on, if we need to re-connect the nodes to the ancestor that was kept, we just repeat the same process: go up the original tree until we hit a node that was kept (exists in `nodesKept`).

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Done

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Done

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Done

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: (For future reference - this is kept for debugging purposes. Since this is a fragile connector, we need to store information to potentially debug).

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,548 @@
...
+      );
+    }
+
+    return teamNameToParentTeamName;
+  }
+  private computeOwnershipChain(
+    elementId: string,
+    allOrgId: string,
+    teamToParent: Record<string, string>
+  ): {cycle: boolean; ownershipChain: ReadonlyArray<string>} {
```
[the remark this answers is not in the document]
omreego: Throwing an error at the end of processing

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Good point - this will really simplify the logic.

destinations/airbyte-faros-destination/test/converters/workday.test.ts
```diff
@@ -0,0 +1,288 @@
...
+  for (const fieldName of fieldNames) {
+    if (!(fieldName in fieldNameToValue)) {
+      throw new Error(`Field name ${fieldName} missing from fieldValues`);
+    }
+    crDest.setField(fieldName, fieldNameToValue[fieldName]);
+  }
+  return crDest;
+}
+
+function getCustomReportandCtxGivenKey(
```
omreego, line 48: @ypc-faros here we get the connector object and initialize it with config and create Stream Context. This is what allows us to call functions from the class and test intermediate steps in the connector

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+    }
+  }
+
+  private checkIfTime1GreaterThanTime2(time1: Date, time2: Date): boolean {
+    if (time1.getTime() > time2.getTime()) {
+      return true;
+    } else {
+      return false;
+    }
+  }
```
willmarks, line 107: dates should just be directly comparable. eg:
```ts
if (time1 > time2) {
    ...
}
```

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+    // We're only keeping active records
+    if (rec.Termination_Date) {
+      return false;
+    }
+    return true;
+  }
+
+  private getManagerIDFromList(recs: ManagerTimeRecord[]): string {
+    if (!recs) {
+      throw new Error('Missing recs');
```
willmarks, line 141: instead of throwing, is there a way for us to gracefully handle this scenario and log? We may need to return 
```ts
string | undefined
```
and handle the undefined case downstream.

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+    do {
+      ownershipChain.push(id);
+      if (visited.has(id)) {
+        return {cycle: true, ownershipChain};
+      }
+      visited.add(id);
+
+      const element = teamToParent[id];
+      if (!element) break;
+      id = element;
```
willmarks, line 202: can all be replaced with:
```ts
id = teamToParent[id];
```

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+      // we keep all teams
+      orgs_to_keep.push(this.FAROS_TEAM_ROOT);
+    }
+    // Setting the values
+    this.orgs_to_keep = orgs_to_keep;
+    this.orgs_to_ignore = orgs_to_ignore;
+    return [orgs_to_keep, orgs_to_ignore];
+  }
+
+  private KeepTeamLogicNew(ownershipChain): boolean {
```
willmarks, line 235: Is there a `KeepTeamLogicOld`? can we call this `shouldKeepTeam` or something?

nit: lets give `ownershipChain` a type.

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+    ctx: StreamContext
+  ): [ReadonlyArray<DestinationRecord>, Record<string, string>] {
+    // Class fields required to be filled (reference for testing):
+    // recordCount, teamNameToManagerIDs, employeeIDtoRecord
+    // FAROS_TEAM_ROOT, cycleChains, generalLogCollection
+    const res: DestinationRecord[] = [];
+    ctx.logger.info(
+      `Starting 'onProcessingComplete'. Records skipped: '${this.recordCount.skippedRecords}',` +
+        ` records kept: '${this.recordCount.storedRecords}'.`
+    );
```
willmarks, line 439: nit: This might make more sense in the onProcessComplete function (just in case logic moves around)

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
```diff
@@ -0,0 +1,484 @@
...
+
+  private getAcceptableTeams(
+    teamToParent: Record<string, string>,
+    ctx: StreamContext
+  ): Set<string> {
+    // This is the entry point for the logic which defines which teams to keep
+    // (ctx is included for potential debugging)
+    ctx.logger.info('Getting Acceptable teams');
+    const acceptableTeams = new Set<string>();
+    this.setOrgsToKeepAndIgnore(ctx);
```
willmarks, line 285: This seems a little dangerous that the timing of this call is important. I would either call this more obviously in some setup step right at the beginning of your on process complete or instead use the returned lists and pass them down to the methods that need them.

destinations/airbyte-faros-destination/test/resources/workday/generate_records.txt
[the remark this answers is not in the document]
omreego: This was raising a SonarQube error because it has a random number generator within it. 
I made it a .txt file to avoid the SonarQube parsing

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: That's true, I think I wrote the function in case there was more to check, but the check is simple

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Good point, I'll work on that

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: That's a good point - I can move this to the beginning of the process

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: true

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Ah, I deleted KeepTeamLogic - I'll rename it

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Nice catch! I borrowed this function so I didn't look too closely at this

destinations/airbyte-faros-destination/src/converters/workday/customreports.ts
omreego: Yep, doing that

willmarks (approved):
seems fine to me. Unless @ypc-faros had anything else I think its fine to merge. We can always iterate if anything comes out of testing with real data.

destinations/airbyte-faros-destination/src/converters/workday/models.ts
omreego: They can't be - they are essential for creating the org structure - at least Manager_ID

ypc-faros:
Thank you @willmarks for the review. I'll do another pass but it's looking pretty good 🎉 
Thank you for the tests @omreego !

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+          "const": "Workday",
+          "order": 0
+        },
+        "orgs_to_keep": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "title": "Orgs To Keep",
+          "description": "The teams underneath these orgs will be kept",
```
ypc-faros, line 21: nit: maybe "Teams belonging to these orgs will be kept"

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
```diff
@@ -0,0 +1,36 @@
...
+          "default": []
+        },
+        "orgs_to_ignore": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "title": "Orgs To Ignore",
+          "description": "The teams underneath these orgs will be ignored. If there is a keep org as a child of ignored, child will be kept.",
```
ypc-faros, line 30: I think we could remove the "If there is a keep org as a child of ignored, child will be kept."
IMO the logic to keep teams based on these configs deserves a doc page and we can just link from here once it's ready.

destinations/airbyte-faros-destination/test/converters/workday.test.ts
```diff
@@ -0,0 +1,305 @@
...
+  test('check resulting org structure from "empty" input', () => {
+    const [customReportDestination, ctx] = getCustomReportandCtxGivenKey(
+      mockttp,
+      'empty'
+    );
+    const [res, finalTeamToParent] = runCustomReportDestination(
+      customReportDestination,
+      ctx
+    );
+    expect(JSON.stringify(finalTeamToParent)).toMatch(
```
ypc-faros, line 274: nit: No need to stringify the `actual` in these expects. You can just do something like:
```
expect(finalTeamToParent).toMatchSnapshot();
or [toMatchInlineSnapshot](https://jestjs.io/docs/snapshot-testing#inline-snapshots)
```

ypc-faros (approved):
Agree with @willmarks. Let's take it for a spin and iterate if needed. Great work @omreego 👏 🥳 !

destinations/airbyte-faros-destination/resources/source-specific-configs/workday.json
omreego: That's true - we may need to show an example in a doc page

destinations/airbyte-faros-destination/test/converters/workday.test.ts
omreego: Ah, that's cool!

PR #1186, opened 2023-10-20
FAI-8128 - Consider already processed streams when choosing whether to process a record or push to process last

cjwooo:
## Description

This change will prevent the destination from needlessly saving records in memory when a stream's dependencies have already been processed. The order that an Airbyte source reads streams is determined by the configured catalog, so we can order the streams in that catalog to minimize the number of streams that need to be saved in memory.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/test/resources/statuspage/all-streams.log
```diff
@@ -1,16 +1,11 @@
...
-{"log":{"level":"INFO","message":"Finished syncing pages stream. Read 1 records"},"type":"LOG"}
+{"log":{"level":"INFO","message":"Syncing component_groups stream in full mode"},"type":"LOG"}
```
cjwooo, line 2: The records here are reordered to reflect the statuspage test catalog.

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -688,12 +689,22 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
                 this.logger.info(`Stream context stats: ${ctx.stats()}`);
               }
             }
-            // Process the record immediately if converter has no dependencies,
-            // otherwise process it later once all streams are processed.
-            if (converter.dependencies.length === 0) {
-              await writeRecord(ctx);
-            } else {
+            // Process the record immediately if converter has pending stream
```
ypc-faros, line 692: if converter has pending stream... => if converter has _no_ pending stream...

ypc-faros (approved):

willmarks (approved):

PR #1192, opened 2023-10-21
Factor out abstract surveys converter

ypc-faros:
## Description

All we need to do to support surveys from another source system is implement a converter extending `AbstractSurveys` and implement these methods:

```
abstract getSurveyId(record: AirbyteRecord): string | undefined;
abstract getTableName(record: AirbyteRecord): string | undefined;
abstract getSubmittedAt(record: AirbyteRecord): string | undefined;
```

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


matiaslcoulougian:
Nit: In getTableName, table name is kind of specific to Airtable. Don't know if there is a more generic name to refer to table/sheet/tab name for all different converters to implement. Maybe could add a comment on the function to clarify

ypc-faros:
> Nit: In getTableName, table name is kind of specific to Airtable. Don't know if there is a more generic name to refer to table/sheet/tab name for all different converters to implement. Maybe could add a comment on the function to clarify

Discussed with @matiaslcoulougian offline. We'll keep it as is and revisit if we can think of a better name when we add more converters

PR #1193, opened 2023-10-23
FAI-8128 - Some minor fixes 

tovbinm:


tovbinm:
@cjwooo we could also clear our accumulated records from memory once all the dependent streams are completed. WDYT?

leahmcguire (approved):

PR #1194, opened 2023-10-24
Asana spec nits + add cutoff

ypc-faros:
Adds `cutoff_days` for consistency with other sources

Leaves `start/end` date as hidden

### DIFF
--- a/sources/asana-source/src/asana.ts
+++ b/sources/asana-source/src/asana.ts
@@ -20,6 +20,7 @@ export const MIN_DATE = new Date(0).toISOString();
 // January 1, 2200
 export const MAX_DATE = new Date(7258118400000).toISOString();
 
+const DEFAULT_CUTOFF_DAYS = 90;
 const DEFAULT_PAGE_SIZE = 100;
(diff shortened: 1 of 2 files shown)


sources/asana-source/src/asana.ts
```diff
@@ -52,8 +54,19 @@ export class Asana {
...
+    let startDate: string;
+    let endDate: string;
+
+    if (config.start_date || config.end_date) {
+      startDate = config.start_date ?? MIN_DATE;
+      endDate = config.end_date ?? MAX_DATE;
+    } else {
+      const cutoffDays = config.cutoff_days ?? DEFAULT_CUTOFF_DAYS;
+      const cutoffDate = new Date();
+      cutoffDate.setDate(cutoffDate.getDate() - cutoffDays);
```
tovbinm, line 66: Please make sure it does what you expect

tovbinm (approved):

sources/asana-source/src/asana.ts
ypc-faros: 🙏

PR #1197, opened 2023-10-27
[FAI-8170] Azure Repos principalName Fix

omreego:
## Description

If principalName is missing, we write User Records with empty uid.
This fix ensures principalName is included in order to properly write records.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/azure-repos/users.ts
+++ b/destinations/airbyte-faros-destination/src/converters/azure-repos/users.ts
@@ -11,10 +11,8 @@ export class Users extends AzureReposConverter {
   ];
 
   private checkUserItemValidity(userItem: User): boolean {
-    if (!userItem.principalName) {
-      return false;
-    }
-    return true;
+    // Add checks to record in this function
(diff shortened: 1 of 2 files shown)


destinations/airbyte-faros-destination/src/converters/azure-repos/users.ts
```diff
@@ -10,12 +10,22 @@ export class Users extends AzureReposConverter {
     'vcs_User',
   ];
 
+  private checkUserItemValidity(userItem: User): boolean {
+    if (!userItem.principalName) {
```
cjwooo, line 14: `return !!userItem.principalName`. `!!` == true if variable is truthy

cjwooo (approved):

destinations/airbyte-faros-destination/src/converters/azure-repos/users.ts
omreego: Makes sense

PR #1199, opened 2023-10-30
FAI-8191 - Bitbucket Server: pull all projects if none are specified

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/bitbucket-server-source/resources/spec.json
+++ b/sources/bitbucket-server-source/resources/spec.json
@@ -4,14 +4,19 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "Bitbucket Server Spec",
     "type": "object",
-    "required": ["cutoff_days", "projects", "server_url"],
+    "required": [
+      "cutoff_days",
+      "server_url"
+    ],
     "additionalProperties": true,
     "properties": {
       "server_url": {
         "type": "string",
         "title": "Server URL",
         "description": "Bitbucket Server URL",
-        "examples": ["https://my.bitbucketserver.com"]
+        "examples": [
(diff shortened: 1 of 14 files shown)


sources/bitbucket-server-source/resources/spec.json
```diff
@@ -4,14 +4,19 @@
     "$schema": "http://json-schema.org/draft-07/schema#",
     "title": "Bitbucket Server Spec",
     "type": "object",
-    "required": ["cutoff_days", "projects", "server_url"],
+    "required": [
+      "cutoff_days",
```
ypc-faros, line 8: Since you're here `cutoff_days` shouldn't be required, since there's a default

willmarks (approved):

ypc-faros (approved):

sources/bitbucket-server-source/resources/spec.json
```diff
@@ -35,7 +39,9 @@
         },
         "title": "Projects",
         "description": "Keys of your Bitbucket Server projects",
-        "examples": ["PROJ"]
```
tovbinm, line 38: Perhaps the dafult should be `"*"` ? I think we did that for other sources

cjwooo: The default values in a spec.json are only used in the Airbyte UI setup screen. The actual value must still be provided in the connector config, i.e. when using the airbyte-local-cli. Therefore I prefer empty default to keep the config and the cli command small.

PR #1202, opened 2023-10-31
Fix Asana project stream

ypc-faros:
## Description

Because of the order of the spread and the model name, I ended up writing two identical boards instead of a board and a project with the same uid+name.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


thomas-gerber (approved):
LGTM

PR #1210, opened 2023-11-03
Add support for number fields to the wizard

ypc-faros:


faros-airbyte-cdk/src/help.ts
```diff
@@ -274,6 +278,7 @@ function choiceAsType(row: TableRow, choice: string) {
     case 'boolean':
       return choice === 'true';
     case 'integer':
+    case 'number':
```
leahmcguire, line 281: do you also need a return statement for the integer case?

leahmcguire (approved):

leahmcguire:
Thank you!  here is the ticket https://faros-ai.atlassian.net/browse/FAI-8320

faros-airbyte-cdk/src/help.ts
ypc-faros: The `return +choice;` applies to both (this is pretty much a casting of string to number)

PR #1211, opened 2023-11-03
FAI-8315  - Buildkite job env

willmarks:
## Description

Support retrieving job env when specified.

WAITING FOR CANONICAL MODEL UPDATE!!

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/buildkite-source/resources/gql/pipeline-builds-query.gql
```diff
@@ -38,6 +38,7 @@ query pipelineBuildsQuery($slug: ID!, $pageSize: Int!, $createdAtFrom: DateTime,
                     startedAt
                     url
                     command
+                    env @include(if: $shouldFetchEnv)
```
willmarks, line 41: My thought here is that env is probably only useful in a job of the command type

tovbinm: why not always fetch envs?

tovbinm: Got it. If it's a lot of data, then it makes sense to make it conditional.

PR #1212, opened 2023-11-03
FAI-8191 - Add global source option to compress state

cjwooo:
## Description

I made this optional instead of enforced because this is a public cdk and usually people want uncompressed state for easier inspection/debugging/modifying.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/faros-airbyte-cdk/src/protocol.ts
+++ b/faros-airbyte-cdk/src/protocol.ts
@@ -50,7 +50,10 @@ export enum AirbyteTraceFailureType {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface AirbyteConfig {}
+export interface AirbyteConfig {
+  compress_state?: boolean;
+  [k: string]: any;
+}
 
 export function parseAirbyteMessage(s: string): AirbyteMessage {
   try {
--- a/faros-airbyte-cdk/src/sources/source-base.ts
+++ b/faros-airbyte-cdk/src/sources/source-base.ts
@@ -17,6 +17,7 @@ import {
   SyncMode,
 } from '../protocol';
 import {AirbyteSource} from './source';
+import {State} from './state';
 import {AirbyteStreamBase} from './streams/stream-base';
 
 /**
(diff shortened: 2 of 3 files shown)


ypc-faros (approved):

faros-airbyte-cdk/src/protocol.ts
```diff
@@ -50,7 +50,10 @@ export enum AirbyteTraceFailureType {
 }
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface AirbyteConfig {}
+export interface AirbyteConfig {
+  compress_state?: boolean;
```
tovbinm, line 54: where will you be setting it to `true`?

cjwooo: The user sets it to true, e.g. with cli: `--src.compress_state true`

tovbinm: This is a breaking change for our connectors. Please make sure to reflect it in the airbyte-connectors-configs

cjwooo: This is not a breaking change. The param doesn't need to be set, and you can switch between compressed and uncompressed whenever you want.

tovbinm: currently we compress the state by default, and a lot of our connectors have a very large state which they should be compressing.

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -131,8 +132,15 @@ export abstract class AirbyteSourceBase<
           configuredStream,
           connectorState
         );
+
         for await (const message of generator) {
-          yield message;
+          if (isStateMessage(message) && config.compress_state) {
+            yield new AirbyteStateMessage({
+              data: State.compress(message.state.data),
```
tovbinm, line 139: dont forget the `message.state.sourceStatus`

PR #1213, opened 2023-11-03
FAI-8329: Fix serialization of primary keys that contain objects

ted-faros:
## Description

Fix serialization of primary keys that contain objects

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -12,6 +12,7 @@ import {
   intersection,
   isEmpty,
   isNil,
+  isObject,
   keys,
   max,
   orderBy,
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -115,13 +116,17 @@ export class UpsertBuffer {
   }
 }
 
-export function serialize(obj: Dictionary<number | string>): string {
+export function serialize(obj: Dictionary<any>): string {
   return keys(obj)
     .sort()
-    .map((k) => `${k}:${obj[k]}`)
+    .map((k) => `${k}:${serializeValue(obj[k])}`)
```
ted-faros, line 122: Using a custom function here instead of JSON.stringify because existing unit tests of this function break otherwise.  And, output of stringify is ugly 😜

ypc-faros (approved):

destinations/airbyte-faros-destination/src/common/graphql-client.ts
tovbinm: `JSON.stringify` is much more reliable than this recursive solution.

destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -115,13 +116,17 @@ export class UpsertBuffer {
...
     .join('|');
 }
 
+export function serializeValue(obj: any): string {
+  return isObject(obj) ? serialize(obj) : String(obj);
```
tovbinm, line 127: perhaps do `JSON.stringify(obj)`?

tovbinm, line 127: perhaps?
```suggestion
  return isObject(obj) ? serialize(obj) : JSON.stringify(obj);
```

PR #1221, opened 2023-11-07
Graph doctor better logging

omreego:
## Description

Improving logging and descriptions for tests.

## Type of change
- [X] Bug fix


sources/faros-graphdoctor-source/src/graphdoctor/z_scores.ts
```diff
@@ -55,8 +52,6 @@ export const runAllZScoreTests: GraphDoctorTestFunction = async function* (
     data_issues.push(...new_data_issues);
   }
 
-  // For testing:
-  console.log(data_issues);
```
tovbinm, line 59: ouch, please avoid using console.log, and use logger instead

willmarks (approved):

PR #1229, opened 2023-11-20
[FAI-8458] - Adding `onBeforeRead` method to source interface

omreego:
## Description
There are cases in which the config needs to be updated before running the connector and an API call needs to be made to update the config. The function "onBeforeRead" allows for this.

This is done to enable us to compute all the projects we need in the config before running the connector 
(Assuming a wildcard / blocklist is provided).

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


faros-airbyte-cdk/src/sources/source.ts
```diff
@@ -22,6 +22,18 @@ export abstract class AirbyteSource<
    */
   abstract discover(config: Config): Promise<AirbyteCatalogMessage>;
 
+  async onBeforeRead(
```
tovbinm, line 25: please add docs

tovbinm (approved):

faros-airbyte-cdk/src/sources/source.ts
omreego: Added :)

PR #1230, opened 2023-11-20
[FAI-8458] Circle CI - adding block list capabilities

omreego:
## Description
Adding a "block list" to Circle CI.
This also allows for the inputs to only be repo names as opposed to full project slugs:
e.g. instead of "A/B/C" for a project slug, users only need to input "C". The rest can be obtained by an API call to get the "org slug"

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/circleci-source/resources/spec.json
```diff
@@ -22,7 +22,15 @@
...
-        "description": "Names should be in the format \"vcs-slug/org-name/repo-name\""
+        "description": "\"repo-name\" or \"vcs-slug/org-name/repo-name\". \"*\" for all projects"
+      },
+      "project_block_list": {
+        "type": "array",
+        "items": {
+          "type": "string"
+        },
+        "title": "Project Name Blocklist",
+        "description": "If project_names is \"*\", then list of \"repo-name\" or \"vcs-slug/org-name/repo-name\""
```
cjwooo, line 33: `then enter a list of ....`

sources/circleci-source/src/index.ts
```diff
@@ -28,13 +30,41 @@ export class CircleCISource extends AirbyteSourceBase<CircleCIConfig> {
...
+        config,
+        this.logger
+      );
+    } else if (config.slugs_as_repos === false) {
+      filtered_project_names = await CircleCI.getFilteredProjects(
+        config,
+        this.logger
+      );
+    } else {
+      throw new Error('Variables slugs_as_repos not set');
```
cjwooo, line 61: You can drop the `Variables` from the message. Or maybe `slugs_as_repos param not set`

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -90,6 +99,179 @@ export class CircleCI {
...
+      }
+    } catch (error: any) {
+      throw new Error(
+        `Failed to get all project "repo names" or "project ids" from '/projects' endpoint`
+      );
+    }
+    if (repo_names.length == 0) {
+      throw new Error('No reponames found for this user');
+    }
+    const repoNamesToProjectIds: Map<string, string> = new Map<
```
cjwooo, line 265: I think you can drop the `<string, string>` from the Map constructor call. The type safety should still be enforced by `const <var>: Map<..> = `

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -90,6 +99,179 @@ export class CircleCI {
...
+    logger: AirbyteLogger,
+    api_version: string = 'v2'
+  ): AxiosInstance {
+    // In this function we rely on the fact that the  API URL contains 'v{X}' in it,
+    // where X is the version number, e.g. "2" or "1.1". We replace the
+    // version number with the one we want to use, stored in the param 'api_version'
+    let url = config.url ?? DEFAULT_API_URL;
+    const versionRegex = /v\d+(\.\d+)?/g;
+    url = url.replace(versionRegex, api_version);
+    logger.info(`New URL for axios: "${url}"`);
```
cjwooo, line 218: Don't mention axios in the log message. Put something like `Using API Url: `

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -90,6 +99,179 @@ export class CircleCI {
...
+  static async getWildCardProjectNames(
+    config: CircleCIConfig,
+    logger: AirbyteLogger
+  ): Promise<string[]> {
+    // We return list of updated project names
+    // In this case we know project names has a wildcard
+    // Always returns list of complete project names (not just repo names)
+    if (!config.project_names.includes('*')) {
+      throw new Error(
+        `We expect project names to just be the wildcard, instead: ${JSON.stringify(
```
cjwooo, line 190: `Expected project_names to just ..., instead got: ${JSON...`

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -90,6 +99,179 @@ export class CircleCI {
...
+    logger.info('Getting Org Slug');
+    let slug: string = '';
+    try {
+      const resp = await circleCIV2Instance.get('/me/collaborations');
+      if (resp.status != 200) {
+        throw new Error(
+          `Failed response from endpoint 'me/collaborations' for getting slug.`
+        );
+      }
+      const resp_data = resp.data[0];
```
cjwooo, line 115: Any chance this endpoint returns an empty result?

cjwooo (approved):

sources/circleci-source/src/circleci/circleci.ts
omreego: It's possible, but from what I've seen it doesn't. I'm wrapping the entire thing in a try-catch statement in case it returns something unexpected

PR #1240, opened 2023-11-30
FAI-8248 Axios + Faros js client upgrade

ypc-faros:


tovbinm (approved):
Thanks! We should bump minor version when releasing

ypc-faros:
> Thanks! We should bump minor version when releasing

Thank you @tovbinm 🙏

PR #1251, opened 2023-12-05
FAI-8704 Fixing spec to have airbyte secret for token

omreego:
## Description

Faros API Key should be an airbyte secret

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


tovbinm:
Is this still relevant?

PR #1256, opened 2023-12-12
FAI-8817 | Add check to avoid fetching rows out of range from sheet

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/sheets-source/src/sheets-reader.ts
+++ b/sources/sheets-source/src/sheets-reader.ts
@@ -313,12 +313,13 @@ export class SheetsReader {
     const res: GoogleSpreadsheetRow[] = [];
     let total = 0;
     let fetchedRows = 0;
+    // Fetch rows in batches of pageSize until we reach the end of the sheet
     do {
       const rows = await sheet.getRows({limit: pageSize, offset: total});
       res.push(...rows);
       fetchedRows = rows.length;
       total += fetchedRows;
-    } while (fetchedRows > 0);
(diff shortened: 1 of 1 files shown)


sources/sheets-source/src/sheets-reader.ts
```diff
@@ -313,12 +313,13 @@ export class SheetsReader {
...
     let total = 0;
     let fetchedRows = 0;
+    // Fetch rows in batches of pageSize until we reach the end of the sheet
     do {
       const rows = await sheet.getRows({limit: pageSize, offset: total});
       res.push(...rows);
       fetchedRows = rows.length;
       total += fetchedRows;
-    } while (fetchedRows > 0);
+    } while (fetchedRows > 0 && total < sheet.rowCount - 1); // -1 because row count includes header row, but it's excluded from getRows
```
ypc-faros, line 322: nit: extract `sheet.rowCount - 1` into a const (`nonHeaderRowCount`) for readability

ypc-faros (approved):

sources/sheets-source/src/sheets-reader.ts
tovbinm: does the same problem exist for anything else besides Google Spreadsheet?

matiaslcoulougian: No, it should not, as the problem is directly related to our requests to the Google Sheets API. For S3 and Local File options we are directly reading and loading the file to an XLSX.WorkBook

PR #1265, opened 2023-12-15
FAI-8724 FAI-8855 Circle CI Handles Github- Gitlab formats when getting all projects with wildcard

omreego:
## Description
Reading at this URL, we see that Circle CI have two different formats for their project names.
This PR adds support for that case:
https://circleci.com/docs/api/v2/index.html#operation/getProjectBySlug

Also adding documentation and simplified logic for reading the "OnBeforeRead" function.
Documentation in this file:
sources/circleci-source/src/index.ts
under the function 'OnBeforeRead'

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/circleci-source/resources/spec.json
+++ b/sources/circleci-source/resources/spec.json
@@ -1,95 +1,100 @@
 {
-  "documentationUrl": "https://docs.faros.ai",
-  "connectionSpecification": {
-    "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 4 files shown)


cjwooo (approved):
In general, our naming convention for variables is to use camelCasing. The exception is for variables that are read directly from the Source/Destination config, i.e. `config.variable_name`. Please update your variables accordingly in this or a subsequent PR.

PR #1276, opened 2023-12-26
GraphDoctor notes incremental and checks for 4-day tms_Task/Pull Request delay

omreego:
## Description
Graph Doctor will sync incremental, not full refresh, because we want a history of counts, not just the present values
Also add a test that checks if for certain models, tms_Task, etc., any objects were added in the last X days, where X defaults to 4.

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_WITHIN_DAYS = 4;
+
+export const compute_number_min_threshold: number = 10;
```
tovbinm, line 3: please follow the camelCase convention for variable names

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_WITHIN_DAYS = 4;
```
tovbinm, line 1: why 4 days?

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
tovbinm, line 1: what are all these magic constants? add a line of docs for each

sources/faros-graphdoctor-source/src/graphdoctor/timeTests.ts
```diff
@@ -57,6 +59,57 @@ export const runAllZScoreTests: GraphDoctorTestFunction = async function* (
...
+    `Last updated time for ${modelName}: ${last_updated_time_str}`
+  );
+  const last_difference: number =
+    secondsSinceEpoch - last_updated_time.getTime() / 1000;
+  if (last_difference > 86400 * withinDays) {
+    const desc_str: string = `Recency issue: ${modelName} last updated time greater than ${withinDays} days in the past: "${last_updated_time}".`;
+    return {
+      faros_DataQualityIssue: {
+        uid: `RecencyIssue_${modelName}_${obj_data.refreshedAt}`,
+        title: 'recency',
```
tovbinm, line 103: not a very useful title to my taste. are you following some convention here?

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
tovbinm: I think this needs to be a param, with a default value of 2 (days).

tovbinm (approved):

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
omreego: Definitely open to change - this is essentially a dummy variable. I need to make this an input to the connector so people can decide around what timeframe the error will happen.

4 was chosen for now because I thought 2 days might alert after a single weekend. Three days might account for a delay between Friday and Monday, so if there have been no updates for 4 days we can assume there is an issue.

sources/faros-graphdoctor-source/src/graphdoctor/timeTests.ts
omreego: For the uid, I wanted it to be splittable by "_" potentially for easy processing.
For title, I thought of it as a category we can use to separate the different issues in the future. 
You're right that it doesn't have enough information. I'll update it

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
tovbinm: see https://github.com/faros-ai/airbyte-connectors/pull/1276#discussion_r1436669897

sources/faros-graphdoctor-source/src/graphdoctor/timeTests.ts
tovbinm: is there are convention you've following here?

omreego: No convention for the "title" field

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
omreego: Done

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
omreego: Done

sources/faros-graphdoctor-source/src/graphdoctor/constants.ts
omreego: Done

sources/faros-graphdoctor-source/src/graphdoctor/timeTests.ts
tovbinm: generally the title should be assumed human readable & presentable, e.g `title: 'Data Recency Issue``. `recency` is more of a category name, which we don't have a field for.

PR #1280, opened 2023-12-27
FAI-8919 Optionally associate tags with the metric definitions OR values

ypc-faros:
## Description

Accepts a tag key/value to associate with the written metric definitions OR values.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/converters/aws-cloudwatch-metrics/metrics.ts
```diff
@@ -42,11 +66,38 @@ export class Metrics extends Converter {
...
 
+    if (this.config.tag_uid) {
+      const tagUid = this.config.tag_uid;
+      if (!this.tags.has(tagUid)) {
+        res.push({
+          model: 'faros_Tag',
+          record: {
+            uid: tagUid,
+            key: this.config.tag_key ?? null,
+            value: this.config.tag_value ?? null,
```
tovbinm, line 83: if key is null, or both are null, why bother writing it into the graph at all?

destinations/airbyte-faros-destination/src/converters/aws-cloudwatch-metrics/metrics.ts
```diff
@@ -42,11 +66,38 @@ export class Metrics extends Converter {
...
+          },
+        });
+        this.tags.add(tagUid);
+      }
+
+      res.push({
+        model: 'faros_MetricValueTag',
```
tovbinm, line 90: since it's the same tag key/value, you might as well just associate it with the metric definition instead, i.e. `faros_MetricDefinitionTag`.

`faros_MetricValueTag` is for storing the value specific tag, e.g. host name, container id etc.

ypc-faros: The specific use case I'm trying to support here is the following:

A metric is computed for two different groups that we want to compare. In my view the metric itself is not associated (tagged) with neither group A or B. We tag the metric values instead so that we can tell if they are measures of group A or B.

I can see how your proposal is valid (and indeed a much better choice) for other use cases.

Wdyt about supporting both? I'm thinking adding another config field to indicate what to tag (metric or metric value).

destinations/airbyte-faros-destination/src/converters/aws-cloudwatch-metrics/metrics.ts
ypc-faros: Having just the uid worked well for the use case we're solving for. I figured I wouldn't make it required if it's going to be redundant/unused in some cases.

tovbinm:
@ypc-faros are going forward with it or should I close it?

ypc-faros:
> @ypc-faros are going forward with it or should I close it?

Going forward. Waiting for your thoughts here: https://github.com/faros-ai/airbyte-connectors/pull/1280#discussion_r1437718540

destinations/airbyte-faros-destination/src/converters/aws-cloudwatch-metrics/metrics.ts
tovbinm: I say support just one - the simpler one.

tovbinm (approved):

PR #1288, opened 2024-01-05
FAI-8950 - Optimize CircleCI source by removing stream slices

willmarks:
## Description

Stream slices have a large performance impact because of how model batching is done. Only models from the same stream slice are batched. If we remove stream slices, then models from across streams can be batched.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


tovbinm:
@willmarks  Can we close this one?

willmarks:
Yeah, I think we don't want to do this for now

PR #1293, opened 2024-01-09
FAI-9009 - Isolate stream slice failures

willmarks:
## Description

Allows for continuing processing of slices even if N (configurable maximum up to unlimited represented by -1) of them fail. 

## Motivation

We want to do as much processing as possible. If a slice fails, it seems reasonable that we would want there to be a way to attempt processing other stream slices. This way, if a slice has a transient failure, it doesn't affect the processing of other slices.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/faros-airbyte-cdk/src/protocol.ts
+++ b/faros-airbyte-cdk/src/protocol.ts
@@ -119,7 +119,7 @@ export interface AirbyteConfiguredStream {
   cursor_field?: string[];
   destination_sync_mode?: DestinationSyncMode;
   primary_key?: string[][];
-  maxSliceFailures?: number; // -1 means no limit
+  maxSliceFailures?: number; // -1 means unlimited
 }
 
 export interface AirbyteConfiguredCatalog {
--- a/faros-airbyte-cdk/src/protocol.ts
+++ b/faros-airbyte-cdk/src/protocol.ts
@@ -119,6 +119,7 @@ export interface AirbyteConfiguredStream {
   cursor_field?: string[];
   destination_sync_mode?: DestinationSyncMode;
   primary_key?: string[][];
+  maxSliceFailures?: number; // -1 means no limit
 }
 
 export interface AirbyteConfiguredCatalog {
--- a/faros-airbyte-cdk/src/protocol.ts
+++ b/faros-airbyte-cdk/src/protocol.ts
@@ -52,6 +52,7 @@ export enum AirbyteTraceFailureType {
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface AirbyteConfig {
   compress_state?: boolean;
+  max_slice_failures?: number; // -1 means unlimited
   [k: string]: any;
 }
 
@@ -119,7 +120,6 @@ export interface AirbyteConfiguredStream {
   cursor_field?: string[];
   destination_sync_mode?: DestinationSyncMode;
   primary_key?: string[][];
-  maxSliceFailures?: number; // -1 means unlimited
 }
 
 export interface AirbyteConfiguredCatalog {
--- a/faros-airbyte-cdk/src/sources/source-base.ts
+++ b/faros-airbyte-cdk/src/sources/source-base.ts
@@ -266,13 +266,8 @@ export abstract class AirbyteSourceBase<
         if (!slice || !configuredStream.maxSliceFailures) {
           throw e;
         }
+
         failedSlices.push(slice);
-        if (
-          configuredStream.maxSliceFailures !== -1 && // -1 means unlimited allowed slice failures
-          failedSlices.length > configuredStream.maxSliceFailures
-        ) {
-          break;
-        }
         this.logger.error(
           `Encountered an error while processing ${streamName} stream slice ${JSON.stringify(
             slice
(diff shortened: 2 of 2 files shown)


faros-airbyte-cdk/src/protocol.ts
```diff
@@ -119,6 +119,7 @@ export interface AirbyteConfiguredStream {
   cursor_field?: string[];
   destination_sync_mode?: DestinationSyncMode;
   primary_key?: string[][];
+  maxSliceFailures?: number; // -1 means unlimited
```
willmarks, line 122: Feels a little bad to be touching the protocol but I couldn't think of a better place to hold the maxSliceFailure config. Definitely let me know if you have any better ideas.

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -244,23 +245,55 @@ export abstract class AirbyteSourceBase<
...
+        yield this.checkpointState(streamName, streamState, connectorState);
+        if (slice) {
+          this.logger.info(
+            `Finished processing ${streamName} stream slice ${JSON.stringify(
+              slice
+            )}. Read ${recordCounter} records`
+          );
+        }
+      } catch (e: any) {
+        if (!slice || !configuredStream.maxSliceFailures) {
```
willmarks, line 266: I'm trying to decide on the exact behavior. Currently, if `maxSliceFailures` is `undefined` or `0`, I just want it to throw immediately.

faros-airbyte-cdk/src/protocol.ts
cjwooo: We've already extended the protocol for other stuff.

If you put the param here, then it has to be specified in the catalog. I think it's more user-friendly to have it as a source-agnostic config like our added state compression param https://github.com/faros-ai/airbyte-connectors/blob/main/faros-airbyte-cdk/src/protocol.ts#L53-L56

faros-airbyte-cdk/src/sources/source-base.ts
tovbinm: let's make is a non breaking change. check for `undefined`

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -263,25 +263,23 @@ export abstract class AirbyteSourceBase<
...
-          } else {
-            break;
-          }
-        } else {
+        if (!slice || !configuredStream.maxSliceFailures) {
           throw e;
         }
+        failedSlices.push(slice);
+        if (
+          configuredStream.maxSliceFailures !== -1 && // -1 means unlimited allowed slice failures
```
tovbinm: I would avoid `-1`. There is `Infinity` instead.

faros-airbyte-cdk/src/protocol.ts
tovbinm: +1

faros-airbyte-cdk/src/sources/source-base.ts
willmarks: Is that easy to pass in via airbyte config if it expects a number?

faros-airbyte-cdk/src/sources/source-base.ts
willmarks: I thought this was checking `undefined` and would just throw if it was, no?

faros-airbyte-cdk/src/sources/source-base.ts
tovbinm: I dont know

faros-airbyte-cdk/src/sources/source-base.ts
tovbinm: I did not review the code thoroughly, but my point still stands - let's make sure that the changes are backwards compatible

faros-airbyte-cdk/src/sources/source-base.ts
willmarks: According to [this](https://stackoverflow.com/questions/34976378/how-do-you-represent-infinity-in-a-json-api#:~:text=There%20is%20no%20way%20to,is%20not%20part%20of%20it) there is no way to represent Infinity in JSON.

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -244,23 +257,51 @@ export abstract class AirbyteSourceBase<
...
+        yield this.checkpointState(streamName, streamState, connectorState);
+        if (slice) {
+          this.logger.info(
+            `Finished processing ${streamName} stream slice ${JSON.stringify(
+              slice
+            )}. Read ${recordCounter} records`
+          );
+        }
+      } catch (e: any) {
+        if (!slice || typeof maxSliceFailures === 'undefined') {
```
tovbinm, line 278: typeof?

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -313,6 +383,19 @@ export abstract class AirbyteSourceBase<
...
     return new AirbyteStateMessage({data: connectorState});
   }
+
+  private errorState(
+    streamName: string,
+    streamState: any,
+    connectorState: AirbyteState,
+    error: Error
+  ): AirbyteStateMessage {
+    connectorState[streamName] = streamState;
```
tovbinm, line 393: what is this for?

tovbinm (approved):

tovbinm:
Please update the PR description

faros-airbyte-cdk/src/sources/source-base.ts
willmarks: This is a helper to build the AirbyteStateMessage when there is an error. I use it to send the error message to the dest immediately but if we have not reached the maxSliceFailures I delay throwing an error until the end of processing the remaining slices.

PR #1298, opened 2024-01-18
FAI-9154 | Workday Connector - handling of Team ID along with Team Names

omreego:
## Description
Since Team Names might not be unique in a company, we will use the Team ID (also known as Org  ID).

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


tovbinm:
@omreego do we still need this?

omreego:
@tovbinm yes I'd like to add it to this repo as well.
Because we added it for Box, I know we can get Team IDs from Workday. I think it's important to use those instead of Team Names here as well. I made the changes last night after talking to Autodesk.

cjwooo (approved):

PR #1304, opened 2024-01-24
FAI-9231 | Partial Vuln Fix - curl CVE-2023-38545 

omreego:
## Description
Updating Node version and Alpine version and fixes curl vuln
https://nvd.nist.gov/vuln/detail/CVE-2023-38545

## Type of change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,22 +14,20 @@ RUN apk add --no-cache --virtual .gyp python3 make g++ curl \
     && npm install -g lerna @lerna/legacy-package-management tsc
 RUN lerna bootstrap --hoist
 
-CMD ["vim", "--version"]
+ARG version
+RUN test -n "$version" || (echo "'version' argument is not set, e.g --build-arg version=x.y.z" && false)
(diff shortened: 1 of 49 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16-alpine
+FROM node:21-alpine
```
tovbinm, line 1: before we do this, we need to update the node version across the whole repo and make sure it compiles & runs - for example https://github.com/faros-ai/airbyte-connectors/pull/542

tovbinm (changes requested):
not as simple :( see - https://github.com/faros-ai/airbyte-connectors/pull/542

Dockerfile
omreego: Ah, good to know.
I'll try out the steps in #542

tovbinm:
I am not sure if nodejs is fully backward compatible. Folks, how can we test this?

PR #1309, opened 2024-01-30
FAI-9312 - Support non-fatal errors that are passed to destination via state message

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/faros-airbyte-cdk/src/errors.ts
+++ b/faros-airbyte-cdk/src/errors.ts
@@ -63,3 +63,13 @@ export function wrapApiError(error: Error, message?: string): Error {
   }
   return new VError({info}, msg);
 }
+
+// Used to notify the destination of an error without failing the sync
+export class NonFatalError extends VError {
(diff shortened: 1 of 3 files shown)


faros-airbyte-cdk/src/errors.ts
```diff
@@ -63,3 +63,13 @@ export function wrapApiError(error: Error, message?: string): Error {
   }
   return new VError({info}, msg);
 }
+
+// Used to notify the destination of an error without failing the sync
+export class NonFatalError extends VError {
```
cjwooo, line 68: These error extensions will be probably be expanded as we integrating sync statuses into the connectors.

willmarks (approved):

tovbinm (approved):

PR #1323, opened 2024-02-05
workday source spec fix

omreego:
## Description
Looks like default URL was displayed incorrectly (creates double backslash).

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


ypc-faros (commented):
The spec change looks good but can we also be a bit more robust on constructing the URLs in the client?

Using the URL constructor should help.

E.g.,

```
const url = new URL('/oauth2/token', 'https://example.com/');

This gets rid of the double slash ^
```

tovbinm (approved):

PR #1356, opened 2024-03-05
FAI-8901 - Create local Faros Sources + upload logs

cjwooo:
## Description

Creates local accounts on the Faros platform and upload sync logs for viewing via Faros UI.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/destination-logger.ts
```diff
@@ -0,0 +1,154 @@
...
+      const syncLog: AirbyteSourceLog = {
+        timestamp: Date.now(),
+        message: {
+          level: AirbyteLogLevelOrder(msg.log.level),
+          msg: msg.log.message,
+        },
+      };
+
+      if (this._logFiles) {
+        this._logFiles.writeDestinationLogs(syncLog);
```
tovbinm, line 148:
```suggestion
        await this._logFiles.writeDestinationLogs(syncLog);
```

tovbinm (commented):
I think we should be protective on all these new calls - if they fail, we should just catch and log a warning

faros-airbyte-cdk/src/logger.ts
```diff
@@ -71,35 +71,39 @@ export class AirbyteLogger {
...
-    if (msg.type === AirbyteMessageType.LOG) {
-      const levelOrder = AirbyteLogLevelOrder(level);
-      const msgLevelOrder = AirbyteLogLevelOrder((msg as AirbyteLog).log.level);
-      if (levelOrder > msgLevelOrder) return;
-    }
+export function shouldWriteLog(
+  msg: AirbyteLog,
+  level: AirbyteLogLevel
+): boolean {
+  return AirbyteLogLevelOrder(msg.log.level) >= AirbyteLogLevelOrder(level);
```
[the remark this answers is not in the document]
cjwooo: No it wasn't a bug. I'm just flipping the conditional in this PR.

destinations/airbyte-faros-destination/src/destination-logger.ts
```diff
@@ -0,0 +1,154 @@
...
+    rl.on('line', (line) => processLine(line));
+    rl.on('close', () => resolve());
+    fileStream.on('error', (error) => reject(error));
+  });
+}
+
+export class FarosDestinationLogger extends AirbyteLogger {
+  private pendingLogs: AirbyteSourceLog[] = [];
+  private _logFiles: LogFiles;
+  private _shouldSaveLogs: boolean = true;
```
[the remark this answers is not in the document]
cjwooo: There are logs written before the destination determines whether or not the current sync is for a local account.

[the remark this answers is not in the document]
tovbinm: It would be better to keep the original behaviour where do dont write any files by default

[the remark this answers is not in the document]
cjwooo: Let me rephrase. The destination outputs logs before it determines whether it is for a local account, so we need to store those first set of logs in memory. Then after check the account, we either write those logs to disk to be uploaded at the end of the sync if it is a local account, or we discard them if it is not a local account.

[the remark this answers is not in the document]
tovbinm: ok, let's put this in the comment near the variable so it would explain how it's used

destinations/airbyte-faros-destination/src/destination-logger.ts
```diff
@@ -0,0 +1,185 @@
...
+      if (stream.closed || !logs.length) {
+        return;
+      }
+      const logString =
+        logs.map((log) => JSON.stringify(log)).join('\n') + '\n';
+      if (!stream.write(logString)) {
+        // If the stream returns false, wait for the drain event before continuing
+        return new Promise((resolve) => {
+          // Update maxListeners to suppress nodejs warning
+          stream.setMaxListeners(stream.getMaxListeners() + 1);
```
tovbinm, line 63: do we have to use async? all we're doing here is collecting the log entries.

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -66,6 +70,18 @@ export abstract class AirbyteSourceBase<
     return this.constructor.name;
   }
 
+  /**
+   * Source type
+   */
+  abstract get type(): string;
```
tovbinm, line 76: This is annoying to be setting this everywhere, especially because we already have the class name available. why can't we figure this out but need to request from the author to specify this?

faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -66,6 +70,18 @@ export abstract class AirbyteSourceBase<
...
+
+  /**
+   * Source mode
```
tovbinm, line 79: what is this about? mode of what?

faros-airbyte-cdk/src/utils.ts
```diff
@@ -24,7 +23,27 @@ export function redactConfig(config: AirbyteConfig, spec: AirbyteSpec): string {
...
+
+/** Redact config of all secret values based on the provided specification */
+export function redactConfig(
+  config: AirbyteConfig,
+  spec: AirbyteSpec
+): AirbyteConfig {
+  const redact = fastRedact({
+    paths: pathsToRedact(spec),
+    censor: 'REDACTED',
+    serialize: false,
```
tovbinm, line 37: what does this do?

faros-airbyte-cdk/src/utils.ts
```diff
@@ -24,7 +23,27 @@ export function redactConfig(config: AirbyteConfig, spec: AirbyteSpec): string {
...
+  });
+  return redact(_.cloneDeep(config)) as AirbyteConfig;
```
tovbinm, line 39: why do we need to cloneDeep now but we did not need it before?

faros-airbyte-cdk/src/sources/source-runner.ts
```diff
@@ -98,8 +102,16 @@ export class AirbyteSourceRunner<Config extends AirbyteConfig> extends Runner {
           }
 
           try {
+            this.logger.state = state ?? {};
```
tovbinm, line 105: I am really not following the logic here. It's way too involved to use there loggers now. One needs to remember to create it correctly by passing ctor vars, maintain state, and flush it at the end. Can we perhaps do a quick hands on session and try to refactor it correctly before we introduce this into the codebase?

faros-airbyte-cdk/src/sources/source-base.ts
cjwooo: This type needs to match our UI catalog entries.

faros-airbyte-cdk/src/sources/source-base.ts
cjwooo: Faros-feeds mode. Required for our Sources UI page to show correct category and logo per source.

faros-airbyte-cdk/src/utils.ts
cjwooo: Redacts config properties marked as secret parameters.

faros-airbyte-cdk/src/utils.ts
cjwooo: The fastRedact mutates the given object. If serialize == true, it doesn't matter because the function reverts its mutations before returning a string value. If serialize == false, which we have here because we upload the redacted config to the Faros backend, the mutations are persisted.

faros-airbyte-cdk/src/sources/source-runner.ts
cjwooo: Source developers are not affected by any of these changes.

faros-airbyte-cdk/src/utils.ts
tovbinm: I referred to `serialize: false`

faros-airbyte-cdk/src/utils.ts
tovbinm: ah, ok

faros-airbyte-cdk/src/sources/source-runner.ts
tovbinm: how about anyone who uses https://www.npmjs.com/package/airbyte-faros-destination package?

destinations/airbyte-faros-destination/src/destination-logger.ts
cjwooo: Yes. Snippet from nodejs stream doc:
```
* @return `false` if the stream wishes for the calling code to wait for the `'drain'` event to be emitted before continuing to write additional data; otherwise `true`.
```

Most of the time it does returns `true`.

faros-airbyte-cdk/src/sources/source-runner.ts
cjwooo: Developers creating custom faros destination are only concerned about converters, which these changes do not affect.

faros-airbyte-cdk/src/utils.ts
cjwooo: https://github.com/faros-ai/airbyte-connectors/pull/1356#discussion_r1520456495

faros-airbyte-cdk/src/sources/source-runner.ts
tovbinm: Which logger is accessible from converters?

faros-airbyte-cdk/src/sources/source-base.ts
cjwooo: No it's part of our public account schema.

faros-airbyte-cdk/src/sources/source-runner.ts
cjwooo: The logger in the stream context, which is the same logger. But the spec of the actual logging methods is unchanged.

tovbinm (approved):

PR #1361, opened 2024-03-05
FAI-9758 - Jira Task Pull Request Source and Converter

matiaslcoulougian:
TODO: 

- [x] Add tests 

## Description
First stream for Jira Airbyte Connector. Source and destination pull_requests implemented.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/src/jira.ts
```diff
@@ -0,0 +1,1635 @@
...
+  toString,
+} from 'lodash';
+import pLimit from 'p-limit';
+import path from 'path';
+import {Memoize} from 'typescript-memoize';
+import {VError} from 'verror';
+
+import {JiraClient} from './client';
+
+// TODO - Remove our models and use models in jira.js package
```
chalenge, line 26: Can you only add the parts that are necessary for the task pull request? Do not want to just copy everything from the old source without re-thinking the code

chalenge (changes requested):
Let's only add what is needed for task pull request.

matiaslcoulougian:
> Let's only add what is needed for task pull request.

Noted. I left all that was used in the `getIssues`, as we use than function for getting pull requests. But there is bunch of things we are not using yet e.g: assignee, additional fields, changelogs, sprint, etc... Currently we only need the issues key, updated and pull requests. Do you suggest reducing and simplifying it so we only have what we are currently using in `pull_requests` stream?

chalenge:
> > Let's only add what is needed for task pull request.
> 
> Noted. Removed a bunch but I left all that was used in the `getIssues`, as we use than function for getting pull requests. But most of issue things we are not using yet e.g: assignee, additional fields, changelogs, sprint, etc... Currently we only need the issues `key`, `updated` and `pull requests`. Do you suggest reducing and simplifying `getIssues` so we only have what we are currently using in `pull_requests` stream? @chalenge

Yes, let's make it minimal and fetch only needed fields for this part. Also helps make the review easier

sources/jira-source/src/streams/pull_requests.ts
```diff
@@ -0,0 +1,121 @@
...
+  getJsonSchema(): Dictionary<any, string> {
+    return require('../../resources/schemas/taskPullrequests.json');
+  }
+
+  get primaryKey(): StreamKey | undefined {
+    return 'number';
+  }
+
+  get cursorField(): string | string[] {
+    return ['issue', 'updated'];
```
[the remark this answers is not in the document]
matiaslcoulougian: I think this is how we reference nested fields. [Documented](https://github.com/faros-ai/airbyte-connectors/blob/5fc8244e2dbb1b1125cefe1b0c0123821855bbcb/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L81-L89) in `stream-base` `cursonField` function. (same for `primaryKey`)

sources/jira-source/src/streams/pull_requests.ts
```diff
@@ -0,0 +1,121 @@
...
+    protected readonly logger: AirbyteLogger
+  ) {
+    super(logger);
+  }
+  getJsonSchema(): Dictionary<any, string> {
+    return require('../../resources/schemas/taskPullrequests.json');
+  }
+
+  get primaryKey(): StreamKey | undefined {
+    return 'number';
```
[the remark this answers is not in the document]
matiaslcoulougian: Yes, true. If more than one issue reference to same PR, will have same number, so it can't be the PK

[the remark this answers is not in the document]
matiaslcoulougian: Any suggestions on what we should use?

sources/jira-source/src/streams/pull_requests.ts
```diff
@@ -0,0 +1,121 @@
...
+        .duration(this.config.cutoffLagDays || DEFAULT_CUTOFF_LAG_DAYS, 'days')
+        .asMilliseconds();
+      const newState: ProjectState = {
+        issueCutoff: Math.max(
+          latestRecordCutoff.getTime(),
+          newCutoff.getTime() - cutoffLag
+        ),
+      };
+      return {
+        ...currentStreamState,
```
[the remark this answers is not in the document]
matiaslcoulougian: We are overriding project new state using project key and maintaining same state for the other projects in the current stream state

sources/jira-source/src/streams/pull_requests.ts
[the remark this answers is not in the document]
matiaslcoulougian: Maybe composite key with `number` and `issue.key`?

sources/jira-source/src/jira.ts
```diff
@@ -0,0 +1,562 @@
...
+    return this.iterate(
+      (startAt) =>
+        this.api.v2.issueSearch.searchForIssuesUsingJql({
+          jql,
+          startAt,
+          fields: [...fieldIds, ...additionalFieldIds],
+          expand: 'changelog',
+          maxResults: this.maxPageSize,
+        }),
+      async (item: any) => {
```
chalenge, line 524: this should be de-coupled from getIssues. 
1. Issues streams should run to completion without worrying about pull requests.
2. We want to be able to independently sync tasks PRs from a list of Issues without fetching all the things.

sources/jira-source/src/jira.ts
```diff
@@ -0,0 +1,562 @@
...
+  getIssues(
+    projectId: string,
+    syncPullRequests: boolean,
+    updateRange?: [Date, Date]
+  ): AsyncIterableIterator<Issue> {
+    let jql = `project = "${projectId}"`;
+    if (updateRange) {
+      jql += ` AND ${Jira.updatedBetweenJql(updateRange)}`;
+    }
+    const fieldIds = [
```
chalenge, line 490: For webhooks we should not need to fetch all these fields otherwise this defeats the purpose. I think for this pass we only need the issue keys/ids so we can fetch the PRs for those. We add some flag `fetchKeysOnly` if running for webhooks or minimal jira source or something of that sort.

destinations/airbyte-faros-destination/src/converters/jira/task_pull_requests.ts
```diff
@@ -0,0 +1,41 @@
+import {AirbyteRecord} from 'faros-airbyte-cdk';
+import {toLower} from 'lodash';
+
+import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
+import {JiraConverter} from './common';
+
+// Required as dependency by Issues converter
```
chalenge, line 7: what is the dependency / let's remove the dependency. These should run independently of each other

destinations/airbyte-faros-destination/src/converters/jira/task_pull_requests.ts
```diff
@@ -0,0 +1,41 @@
...
+export class TaskPullRequests extends JiraConverter {
```
chalenge, line 8: for consistency let's call this IssuePullRequests. We use `issue` everywhere else in the Jira source/converter

sources/jira-source/src/streams/task_pull_requests.ts
```diff
@@ -0,0 +1,121 @@
...
+    const jira = await Jira.instance(this.config, this.logger);
+    const projectKeys =
+      this.config.projectKeys ?? (await jira.getProjectsByKey()).keys();
+    for (const projectKey of projectKeys) {
+      const updateRange =
+        syncMode === SyncMode.INCREMENTAL
+          ? this.getUpdateRange(streamState?.[projectKey])
+          : undefined;
+      for await (const issue of jira.getIssues(projectKey, true, updateRange)) {
+        if (issue.pullRequests) {
```
chalenge, line 66: you can do to combine these two lines
```typescript
for (const pullRequest of issue.pullRequests || [])
```

chalenge (commented):
Syncing pull requests should not be dependent on getting all the issue info and vice-versa. The pull requests stream should be in charge of fetching that from the API.

chalenge (approved):
lgtm. Let's :shipit:

PR #1363, opened 2024-03-13
FAI-9747 | Added destination configs for redact PII on specific fields and custom regex support

dbruno21:
## Todo
- [x] Add tests

## Description

Adds 3 parameters in faros destination:
 - redact_fields_map: similar to exclude_fields_map but instead of excluding it uses [redact-pii](https://www.npmjs.com/package/redact-pii) library to redact the field value, stripping PII.
 - redact_custom_replace: specifies a custom global value to be used as replacement for any matching pattern.
 - redact_custom_regex: extends builtins regex expressions used by redact-pii to also strip out any other custom pattern passed as input.

By default all these builtin regex's are enabled for redact and replaced with CONSTANT_CASE name for each of them:
- credentials
- creditCardNumber
- emailAddress
- ipAddress
- names
- password
- phoneNumber
- streetAddress
- username
- usSocialSecurityNumber
- zipcode
- url
- digits

The redactor can also be configured to:
- disable any unwanted builtin regex's
- use a global replace value instead of replacing with the pattern detected name
- extend with more regex expressions (not necessarily given as destination parameter)
- can be optionally extended with [Google Data Loss Prevention](https://cloud.google.com/security/products/dlp) to identify broader PII patterns

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/resources/spec.json
+++ b/destinations/airbyte-faros-destination/resources/spec.json
@@ -73,7 +73,7 @@
         "order": 5,
         "type": "string",
         "title": "Redact Fields",
-        "description": "JSON map to redact specific records field values, same structure as 'Exclude Fields' but with redaction instead of removal",
+        "description": "JSON map to redact specific records field values",
         "multiline": true,
         "default": "{}"
       },
--- a/destinations/airbyte-faros-destination/resources/spec.json
+++ b/destinations/airbyte-faros-destination/resources/spec.json
@@ -73,28 +73,34 @@
         "order": 5,
         "type": "string",
         "title": "Redact Fields",
-        "description": "JSON map to redact specific records field values, where each entry (model, fields[]) indicates that the records' fields of specific model type should be redacted",
+        "description": "JSON map to redact specific records field values, same structure as 'Exclude Fields' but with redaction instead of removal",
         "multiline": true,
         "default": "{}",
         "examples": [
           "{ \"tms_Task\": [ \"description\" ], \"vcs_PullRequest\": [ \"description\", \"htmlUrl\" ] }"
         ]
       },
-      "redact_custom_regex": {
+      "redact_custom_replace": {
         "order": 6,
         "type": "string",
+        "title": "Redact Custom Replace",
+        "description": "Word used to replace redacted fields specified by 'Redact Fields'"
+      },
+      "redact_custom_regex": {
+        "order": 7,
+        "type": "string",
         "title": "Redact Custom Regex",
-        "description": "Custom regex to match custom patterns to be redacted on the records field values specified by 'Redact Fields'"
+        "description": "Extend redaction rules with custom regex on the fields specified by 'Redact Fields'"
       },
       "dry_run": {
-        "order": 7,
+        "order": 8,
         "type": "boolean",
         "title": "Dry run",
         "description": "Process all input records but avoid writing into Faros API",
         "default": false
       },
       "invalid_record_strategy": {
-        "order": 8,
+        "order": 9,
         "type": "string",
         "title": "Invalid record strategy",
         "description": "Strategy to follow to handle an invalid input record.",
@@ -105,14 +111,14 @@
         ]
       },
       "jsonata_expression": {
-        "order": 9,
+        "order": 10,
         "type": "string",
         "title": "JSONata expression",
         "description": "JSONata expression for converting input records. If provided applies the expression based on specified JSONata apply mode.",
         "multiline": true
       },
       "jsonata_mode": {
-        "order": 10,
+        "order": 11,
         "type": "string",
         "title": "JSONata apply mode",
         "description": "JSONata apply mode when converting input records.",
@@ -123,7 +129,7 @@
         ]
       },
       "jsonata_destination_models": {
-        "order": 11,
+        "order": 12,
         "type": "array",
         "items": {
           "type": "string"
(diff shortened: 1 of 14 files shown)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -332,9 +335,36 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
...
+        customRedactors: {
+          before: [
+            {
+              regexpPattern: new RegExp(config.redact_custom_regex, 'gi'),
+              replaceWith: 'REDACTED',
+            },
+          ],
+        },
+      }),
+    });
```
willmarks, line 367: Does this redactor setup also automatically redact PII or will it only be using the custom regex pattern for redaction? If it is also automatically redacting, does it replace with `REDACTED` so that it and the customRedactor have the same `replaceWith`?

dbruno21: By default, the configuration is that all the builtins patterns (see description) are enabled as well. If one is unwanted it should be explicitly disabled. The custom regex is for extending with any additional custom patterns.

As for the replacement string, the default behavior is replacing it with the name of the matching pattern (e.g. PHONE_NUMBER, PERSON_NAME). We can also override using a global replacement value by adding  `globalReplaceWith: 'REDACTED'` on the redactor params.

So, it's configurable, we should just decide what's best.

willmarks: Oh I like that it redacts with PHONE_NUMBER etc. that sounds good. Thanks for the explanation!

tovbinm: we should add a global replace with param as well, in case a customer would like to override the default behaviour.

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -69,15 +69,32 @@
           "{ \"tms_Task\": [ \"description\" ], \"vcs_PullRequest\": [ \"description\", \"htmlUrl\" ] }"
         ]
       },
-      "dry_run": {
+      "redact_fields_map": {
         "order": 5,
+        "type": "string",
+        "title": "Redact Fields",
+        "description": "JSON map to redact specific records field values, where each entry (model, fields[]) indicates that the records' fields of specific model type should be redacted",
```
tovbinm, line 76: let's shorten these descriptions. we dont want to explode our destination spec. we had issues with it in the past.

destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -332,9 +335,36 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
...
+    if (config.exclude_fields_map) {
+      this.excludeFieldsByModel =
+        typeof config.exclude_fields_map === 'object'
+          ? config.exclude_fields_map
+          : JSON.parse(config.exclude_fields_map);
+    } else {
+      this.excludeFieldsByModel = {};
+    }
+
+    if (config.redact_fields_map) {
```
chalenge, line 347: @tovbinm should we also have a default map of fields we want to strip PII so it is automatic?

ypc-faros:
LGTM overall 🥳 
For tests, I'd recommend to extract the whole thing into a `Redactor` class and test its behavior separately

destinations/airbyte-faros-destination/src/destination.ts
tovbinm: Lets leave it as is for now

destinations/airbyte-faros-destination/resources/spec.json
ypc-faros: @dbruno21 We need to shorten them even more :|

```
docker build . --build-arg path=destinations/airbyte-faros-destination --build-arg version=0.0.1 -t yandryperez/airbyte-faros-destination --no-cache

docker run yandryperez/airbyte-faros-destination spec | wc -c
   **32496**
```
We're beyond the 32,000 limit.

We're walking on eggshells until we upgrade to after https://github.com/airbytehq/airbyte-platform/releases/tag/v0.50.35

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -69,15 +69,33 @@
           "{ \"tms_Task\": [ \"description\" ], \"vcs_PullRequest\": [ \"description\", \"htmlUrl\" ] }"
         ]
       },
-      "dry_run": {
+      "redact_fields_map": {
         "order": 5,
+        "type": "string",
+        "title": "Redact Fields",
+        "description": "JSON map to redact specific records field values, same structure as 'Exclude Fields' but with redaction instead of removal",
```
tovbinm, line 76:
```suggestion
        "description": "JSON map to redact specific records field values",
```

destinations/airbyte-faros-destination/src/destination-redactor.ts
```diff
@@ -0,0 +1,38 @@
+import {mapValues} from 'lodash';
+import {SyncRedactor} from 'redact-pii';
+
+import {DestinationRecord} from './converters/converter';
+
+export class FarosDestinationRedactor {
```
tovbinm, line 6:
```suggestion
export class RegexRecordRedactor {
```

tovbinm: + rename file + test accordingly

tovbinm (approved):
lgtm, see comments

destinations/airbyte-faros-destination/src/destination-redactor.ts
dbruno21: Renamed to RecordRedactor. Sounds better in case it's extended with non-regex and / or Google DLP.

tovbinm (approved):

PR #1380, opened 2024-03-26
FAI-10335 - Fix workday source url handling

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/workday-source/src/workday.ts
+++ b/sources/workday-source/src/workday.ts
@@ -50,7 +50,6 @@ export class Workday {
       throw new VError('baseUrl must not be an empty string');
     }
 
-    const baseUrl = new URL(cfg.baseUrl);
     const timeout = cfg.timeout ?? 60000;
     const headers = {'content-type': 'application/json'};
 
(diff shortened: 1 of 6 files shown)


sources/workday-source/src/workday.ts
```diff
@@ -81,22 +80,22 @@ export class Workday {
...
-    baseURL: URL,
+    baseURL: string,
     cfg: any,
     logger: AirbyteLogger
   ): Promise<{
     refresh_token: string;
     token_type: string;
     access_token: string;
   }> {
-    const authUrl = new URL(`/oauth2/${cfg.tenant}/token`, baseURL).toString();
```
cjwooo, line 99: unfortunately `new URL(path, base)` trims everything after the hostname in `base`, so the `/ccx` is dropped from `https://blah.workday.com/ccx`

omreego (approved):

PR #1399, opened 2024-04-03
FAI-10279 - Sprint reports stream test and mock objects + config to JSON files

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/test/index.test.ts
```diff
@@ -115,114 +110,67 @@ describe('index', () => {
...
-          },
-        },
-      ],
-    };
+    const expectedPullRequests = readTestResourceFile('pull_requests.json');
+    expectedPullRequests.data = expectedPullRequests.data.map((pr: any) => ({
+      ...pr,
+      issue: {
+        ...pr.issue,
+        updated: new Date(pr.issue.updated),
```
chalenge, line 118: should we make these typescript files so we do need to do this?

sources/jira-source/test/index.test.ts
```diff
@@ -115,114 +110,67 @@ describe('index', () => {
...
+    const expectedSprintReports = readTestResourceFile('sprint_reports.json');
+    expectedSprintReports.data = expectedSprintReports.data.map((sr: any) => ({
+      ...sr,
+      completedAt: new Date(sr.completedAt),
+    }));
+    return expectedSprintReports;
+  }
+
+  test('streams - sprint_reports', async () => {
+    const expectedSprintReports = getExpectedSprintReports();
```
chalenge, line 155: lets use snapshots instead of loading from a file

sources/jira-source/test/index.test.ts
matiaslcoulougian: Might be better yes

chalenge (approved):

PR #1400, opened 2024-04-03
FAI-10277 - Use streamSlice for issue_pull_requests read records.

matiaslcoulougian:
## Description
Added streamSlice usage in the readRecords function to make use of the stream slices we are returning and parallelize runs per project.

## Type of change
- [ x ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/jira-source/src/streams/issue_pull_requests.ts
+++ b/sources/jira-source/src/streams/issue_pull_requests.ts
@@ -30,31 +30,28 @@ export class IssuePullRequests extends StreamWithProjectSlices {
     streamState?: StreamState
   ): AsyncGenerator<PullRequest> {
     const jira = await Jira.instance(this.config, this.logger);
-    const projectKeys =
-      this.config.projectKeys ?? (await jira.getProjectsByKey()).keys();
-    for (const projectKey of projectKeys) {
-      const updateRange =
-        syncMode === SyncMode.INCREMENTAL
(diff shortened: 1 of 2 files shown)


sources/jira-source/src/streams/issue_pull_requests.ts
```diff
@@ -30,31 +30,28 @@ export class IssuePullRequests extends StreamWithProjectSlices {
...
+        : undefined;
+    for await (const issue of jira.getIssues(
+      projectKey,
+      true,
+      updateRange,
+      true,
+      true,
+      [DEV_FIELD_NAME]
+    )) {
+      for (const pullRequest of issue.pullRequests || []) {
```
chalenge, line 46: I thought we de-coupled this and we have a method that only takes a list of issues and fetches pull requests?

chalenge (approved):

sources/jira-source/src/streams/issue_pull_requests.ts
matiaslcoulougian: No, we didn't. We just made the getIssues return only the fields necessary for writing issuePullRequests, making a pullRequest call for each issue inside

chalenge: This needs to be de-coupled. For webhooks we might wanna query the graph instead of Jira

PR #1407, opened 2024-04-05
FAI-10449: Support for wildcards blacklist/whitelist in Circle CI connector

chalenge:
## Description

- Allow pattern matching for wildcards in CircleCI connector to enable easier filtering of slugs

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/circleci-source/src/index.ts
+++ b/sources/circleci-source/src/index.ts
@@ -8,6 +8,7 @@ import {
   AirbyteState,
   AirbyteStreamBase,
 } from 'faros-airbyte-cdk';
+import GlobToRegExp from 'glob-to-regexp';
 import VError from 'verror';
 
 import {CircleCI, CircleCIConfig} from './circleci/circleci';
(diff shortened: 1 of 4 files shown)


sources/circleci-source/src/index.ts
```diff
@@ -103,4 +108,31 @@ export class CircleCISource extends AirbyteSourceBase<CircleCIConfig> {
       new Tests(circleCI, config, this.logger),
     ];
   }
+
+  static filterBlockList(
+    projectSlugs: ReadonlyArray<string>,
+    projectSlugBlocklist: Set<string>,
+    excludedRepoSlugs: Set<string>
+  ): ReadonlyArray<string> {
+    // Convert filter patterns to regular expressions
```
willmarks, line 117: I wonder if we should use something like [this library](https://www.npmjs.com/package/glob-to-regexp) in order to support all glob patterns?

willmarks (approved):
lgmt. Had one comment about using a library to have expanded pattern support but what you have is fine too.

sources/circleci-source/src/index.ts
chalenge: Yes, couldn't find any I liked. Let me try this.

PR #1409, opened 2024-04-08
FAI-10378: Add mapping when group is not in context

chalenge:
## Description

- Add mapping when group is not in context

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/zendesk/common.ts
+++ b/destinations/airbyte-faros-destination/src/converters/zendesk/common.ts
@@ -52,11 +52,17 @@ export abstract class ZendeskConverter extends Converter {
     );
   }
 
-  protected orgTeam(ctx: StreamContext, group: Dictionary<any>): OrgTeam {
+  protected orgTeam(
+    ctx: StreamContext,
+    group?: Dictionary<any>
+  ): OrgTeam | undefined {
     const teamMapping = this.teamMapping(ctx);
-    const mappedTeam = teamMapping[trim(group.name)] || teamMapping['*'];
+    const mappedTeam = teamMapping[trim(group?.name)] || teamMapping['*'];
     if (mappedTeam) {
       return {uid: mappedTeam};
+    }
+    if (!group) {
+      return undefined;
     }
     return {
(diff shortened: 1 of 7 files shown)


destinations/airbyte-faros-destination/test/converters/zendesk.test.ts
```diff
@@ -395,6 +395,26 @@ describe('satisfaction ratings', () => {
     expect(res).toHaveLength(3);
     expect(res).toMatchSnapshot();
   });
+
+  test('rating fallback to group when no group in context', async () => {
```
willmarks, line 399: nit: fallback to group when no group in context

cause it makes the test:
```
satisfaction ratings rating fallback to group when no group in context
```

destinations/airbyte-faros-destination/src/converters/zendesk/common.ts
```diff
@@ -52,12 +52,18 @@ export abstract class ZendeskConverter extends Converter {
...
   }
 
-  protected orgTeam(ctx: StreamContext, group: Dictionary<any>): OrgTeam {
+  protected orgTeam(
+    ctx: StreamContext,
+    group?: Dictionary<any>
+  ): OrgTeam | undefined {
     const teamMapping = this.teamMapping(ctx);
-    const mappedTeam = teamMapping[trim(group.name)] || teamMapping['*'];
+    const mappedTeam = teamMapping[trim(group?.name)] || teamMapping['*'];
```
willmarks, line 60: nit: 
```ts
teamMapping[trim(group?.name) || '*']
```

willmarks (approved):

destinations/airbyte-faros-destination/src/converters/zendesk/common.ts
chalenge: Not the same, `teamMapping[trim(group?.name) || '*']` returns `teamMapping['*']` only when `trim(group?.name)` is falsy. We want to return `teamMapping['*']` even that is not falsy but there's no mapping

PR #1411, opened 2024-04-09
FAI-10504 Delete old task-board and task-project relationships

ypc-faros:


destinations/airbyte-faros-destination/src/converters/asana/tasks.ts
```diff
@@ -239,4 +250,92 @@ export class Tasks extends AsanaConverter {
...
+
+    const res: DestinationRecord[] = [];
+
+    for (const project of projectMembershipsToDelete) {
+      res.push({
+        model: 'tms_TaskBoardRelationship__Deletion',
+        record: {
+          where: {
+            task: taskKey,
+            board: {uid: project, source: this.source},
```
dbruno21, line 324: What's the reason for boards and projects to delete using the same set (projectMembershipsToDelete) ? This would work since for each "project" it writes both the project and board deletion version, but wouldn't it try to delete records that don't even exist unnecessarily like this?

PR #1414, opened 2024-04-11
FAI-10447: CircleCI fix incremental sync

chalenge:
## Description

- CircleCI fix incremental sync, make project slugs consistent and incorrect lastUpdated
- Fix block_list param inconsistency

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/circleci-source/resources/spec.json
+++ b/sources/circleci-source/resources/spec.json
@@ -43,7 +43,13 @@
           "type": "string"
         },
         "title": "Project Slugs",
-        "description": "Project slugs look like \"vcs-slug/org-name/repo-name\". Enter \"*\" for all projects."
+        "description": "Project slugs to sync in the form \"vcs-slug/org-name/repo-name\". Enter \"*\" for all projects. For projects that use GitLab or GitHub App use \"circleci/org-id/project-id\".",
+        "examples": [
+          "gh/CircleCI-Public/api-preview-docs",
+          "bb/CircleCI-Public/api-preview-docs",
+          "circleci/CircleCI-Public/497f6eca-6276-4993-bfeb-53cbbbba6f08",
+          "*"
+        ]
       },
       "project_block_list": {
         "type": "array",
@@ -51,7 +57,13 @@
           "type": "string"
         },
         "title": "Project Slug Blocklist",
-        "description": "Project slugs look like \"vcs-slug/org-name/repo-name\" or with wildcards like \"vcs-slug/org-name/*\""
+        "description": "List of project slugs that should be excluded from syncing. Use exact project slugs or use wildcard characters to filter according to slug patterns, similar to regular expressions.",
+        "examples": [
+          "gh/CircleCI-Public/api-preview-docs",
+          "gh/CircleCI-Public/*",
+          "gh/CircleCI*",
+          "circleci/*/test-builds"
+        ]
       },
       "pull_blocklist_from_graph": {
         "type": "boolean",
--- a/sources/circleci-source/resources/spec.json
+++ b/sources/circleci-source/resources/spec.json
@@ -43,7 +43,7 @@
           "type": "string"
         },
         "title": "Project Slugs",
-        "description": "Project slugs should look like \"vcs-slug/org-name/repo-name\". Only enter \"*\" for all projects."
+        "description": "Project slugs look like \"vcs-slug/org-name/repo-name\". Enter \"*\" for all projects."
       },
       "project_block_list": {
         "type": "array",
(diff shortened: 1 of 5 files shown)


sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -21,7 +21,7 @@ export interface CircleCIConfig {
   readonly token: string;
   readonly url?: string;
   project_slugs: ReadonlyArray<string>;
-  readonly project_blocklist?: string[];
+  readonly project_block_list?: string[];
```
chalenge, line 24: Matches [here](https://github.com/faros-ai/airbyte-connectors/blob/main/sources/circleci-source/resources/spec.json#L48)

matiaslcoulougian (approved):

sources/circleci-source/resources/spec.json
```diff
@@ -43,15 +43,15 @@
...
+        "description": "Project slugs look like \"vcs-slug/org-name/repo-name\". Enter \"*\" for all projects."
       },
       "project_block_list": {
         "type": "array",
         "items": {
           "type": "string"
         },
         "title": "Project Slug Blocklist",
-        "description": "Project slugs should look like \"vcs-slug/org-name/repo-name\""
+        "description": "Project slugs look like \"vcs-slug/org-name/repo-name\" or with wildcards like \"vcs-slug/org-name/*\""
```
ypc-faros, line 54: Same here, add example(s)

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -151,17 +151,23 @@ export class CircleCI {
         // //circleci.com/organization_id/project_id
         // https://github.com/organization/repository
         const projectMatch = projectVCSUrl.match(
-          /^[^/]*\/\/([^./]+).*\/([^./]+)\/([^./]+)$/
+          /^[^/]*\/\/([^./]+).*\/([^./]+)\/([\w.-]+)$/
         );
         if (!projectMatch || projectMatch.length < 1) {
```
ypc-faros, line 156: Below we do `const projectId = projectMatch[3];` so should this assert for length >= 4?

chalenge: Oh yeah, I was curios about that. Forgot to ask @willmarks why we only check for < 1?

sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -280,7 +299,7 @@ export class CircleCI {
     startDate.setDate(startDate.getDate() - this.cutoffDays);
 
     const lastUpdatedAt = since ? new Date(since) : startDate;
-    const gracePeriod = lastUpdatedAt;
+    const gracePeriod = new Date(lastUpdatedAt);
```
ypc-faros, line 302: It was already a `Date` ...

sources/circleci-source/resources/spec.json
chalenge: @ypc-faros. I was worried about the size of the catalog or this doesn't apply here?

ypc-faros: No, it's only for the Faros destination spec 👍

sources/circleci-source/src/circleci/circleci.ts
chalenge: This was a gotcha. Not sure why we need the gracePeriod but if don't make it a new Date reference gracePeriod points to same reference as lastUpdatedAt. It doesn't get the value only, thus when we update [gracePeriod value](https://github.com/faros-ai/airbyte-connectors/pull/1414/files#diff-dcadf7d560989bc22d562a359cb75ca4284aee3713bb6f5166275f5a4c7c8beaL284) lastUpdated also points to this new updated value resulting in the [syncs always pulling data ](https://github.com/faros-ai/airbyte-connectors/pull/1414/files#diff-dcadf7d560989bc22d562a359cb75ca4284aee3713bb6f5166275f5a4c7c8beaL303)from wayback 🤦🏽 
image
what we want is two different values
image

Can put a comment

ypc-faros: Omg 🤯 
`Not sure why we need the gracePeriod` <= yeah good point, maybe we could get rid of it altogether? It doesn't seem like `lastUpdatedAt` is updated anywhere ... 

We should clean up this logic (maybe not now...) It's hard to read for me

ypc-faros (approved):

sources/circleci-source/src/circleci/circleci.ts
chalenge: Yeah, we should update it, but needs more understanding of how CircleCi works
We use `lastUpdatedAt` to filter in `fetchWorkflows` and use gracePeriod for `pipelines`

sources/circleci-source/src/circleci/circleci.ts
willmarks: The < 1 wasn't really necessary iirc as I think it is redundant with checking that projectMatch exists but I added per a review comment. Something like `!projectMatch?.length` probably would have been better.

PR #1415, opened 2024-04-12
FAI-10607 - Add run_mode config to support WebhookGaps mode 

matiaslcoulougian:
## Description

Check for  WebhookGaps mode to select streams before records read.

TODO: 
- [ ] Add test case

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/jira-source/resources/spec.json
+++ b/sources/jira-source/resources/spec.json
@@ -119,6 +119,17 @@
         "items": {
           "type": "string"
         }
+      },
+      "run_mode": {
+        "order": 18,
+        "title": "Run Mode",
+        "description": "Run mode for selecting the usage of the connector",
+        "type": "string",
+        "default": "Full",
+        "enum": [
+          "Full",
+          "WebhookGaps"
+        ]
       }
     }
   }
--- a/sources/jira-source/resources/spec.json
+++ b/sources/jira-source/resources/spec.json
@@ -18,20 +18,23 @@
       "username": {
         "order": 2,
         "title": "Username",
-        "description": "Username for authentication (optional)",
-        "type": "string"
+        "description": "Username for authentication",
+        "type": "string",
+        "airbyte_secret": true
       },
       "password": {
         "order": 3,
         "title": "Password",
-        "description": "Password for authentication (optional)",
-        "type": "string"
+        "description": "Password for authentication",
+        "type": "string",
+        "airbyte_secret": true
       },
       "token": {
         "order": 4,
         "title": "Token",
-        "description": "Token for authentication (optional)",
-        "type": "string"
+        "description": "Token for authentication",
+        "type": "string",
+        "airbyte_secret": true
       },
       "sync_additional_fields": {
         "order": 5,
@@ -128,7 +131,7 @@
         "default": "Full",
         "enum": [
           "Full",
-          "WebhookGaps"
+          "WebhookSupplement"
         ]
       }
     }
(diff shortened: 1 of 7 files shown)


sources/jira-source/src/jira.ts
```diff
@@ -31,6 +32,7 @@ export interface JiraConfig extends AirbyteConfig {
   readonly cutoff_days?: number;
   readonly cutoff_lag_days?: number;
   readonly board_ids?: ReadonlyArray<string>;
+  readonly run_mode?: RunMode;
```
ypc-faros, line 35: This needs to be added to the spec. See [here](https://github.com/faros-ai/airbyte-connectors/blob/3bd8555be7373f19f6b109776483442f93785c51/destinations/airbyte-faros-destination/resources/cloud.json#L31) for an example of an enum

sources/jira-source/src/streams/common.ts
```diff
@@ -21,6 +21,17 @@ export type StreamState = {
   readonly [projectOrBoard: string]: {cutoff: number};
 };
 
+export enum RunMode {
+  Full = 'Full',
+  WebhookGaps = 'WebhookGaps',
```
ypc-faros, line 26: nit: I don't love this name. Sounds too negative. What about `WebhookComplement`? cc/ @chalenge

ypc-faros:
Looking good! Let's add the test before merging...

sources/jira-source/src/streams/common.ts
matiaslcoulougian: Like it better, trying to think some other alternatives...

sources/jira-source/resources/spec.json
ypc-faros, line 1: Let's add `"airbyte_secret": true` to user/pass and token

ypc-faros, line 1: nit: We don't need the `(optional)` in descriptions like `Username for authentication (optional)`
Airbyte properly renders the required and optional fields in the UI based on what's in `required`

sources/jira-source/resources/spec.json
```diff
@@ -119,6 +119,17 @@
...
+      },
+      "run_mode": {
+        "order": 18,
+        "title": "Run Mode",
+        "description": "Run mode for selecting the usage of the connector",
+        "type": "string",
+        "default": "Full",
+        "enum": [
+          "Full",
+          "WebhookGaps"
```
ypc-faros, line 131: Also rename here => WebhookSupplement

sources/jira-source/test/index.test.ts
```diff
@@ -182,4 +183,24 @@ describe('index', () => {
...
     );
   });
+
+  test('onBeforeRead with run_mode WebhookSupplement should filter streams', async () => {
+    const source = new sut.JiraSource(logger);
+    const catalog = readTestResourceFile('catalog.json');
+    const config = readTestResourceFile('config.json');
+    config.run_mode = RunMode.WebhookSupplement;
+    const {catalog: newCatalog} = await source.onBeforeRead(config, catalog);
+    expect(newCatalog.streams).toHaveLength(3);
```
ypc-faros, line 193: ultra nit: Remove this length check. It's redundant given the `toMatchSnapshot` and it's going to be annoying to change it by hand every time we add a stream

ypc-faros (changes requested):
A couple of final comments. We're almost there 🥳

ypc-faros (approved):

PR #1417, opened 2024-04-15
FAI-10626 - Add bucketing to Jira Conector

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/jira-source/resources/spec.json
+++ b/sources/jira-source/resources/spec.json
@@ -133,6 +133,20 @@
           "Full",
           "WebhookSupplement"
         ]
+      },
+      "project_bucket_id": {
+        "order": 19,
+        "type": "integer",
+        "title": "Bucket Number",
+        "description": "Bucket number for this source to determine which portion of projects and boards to pull. Use it when distributing the load between multiple sources.",
+        "default": 1
+      },
+      "project_bucket_total": {
+        "order": 20,
+        "type": "integer",
+        "title": "Total Number of Buckets",
+        "description": "Total number of buckets to distribute projects and boards across. Use it when distributing the load between multiple sources",
+        "default": 1
       }
     }
   }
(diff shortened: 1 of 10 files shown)


sources/jira-source/src/jira.ts
```diff
@@ -716,4 +747,21 @@ export class Jira {
     });
     return filterJQL.jql;
   }
+
+  bucket(project: string): number {
```
matiaslcoulougian, line 751: We should move this function to a common place, as we have the same one for Bitbucket Server connector: https://github.com/faros-ai/airbyte-connectors/blob/f54bf1b70d6693e922c232fbc79e0113ef1170cb/sources/bitbucket-server-source/src/bitbucket-server/index.ts#L608C3-L613C4

I'm not sure where is the correct place to define it so it can be shared by both connectors and future ones.

matiaslcoulougian: @chalenge @ypc-faros

chalenge: Looks like we can put it ins [shared utilities here](https://github.com/faros-ai/airbyte-connectors/tree/main/faros-airbyte-common)

chalenge (commented):
Mostly fine, let's move to shared utilities

sources/jira-source/resources/spec.json
```diff
@@ -133,6 +133,20 @@
           "Full",
           "WebhookSupplement"
         ]
+      },
+      "project_bucket_id": {
```
ypc-faros, line 137: Maybe just `bucket_id`? Since we can bucketize boards or projects...

ypc-faros (approved):

PR #1419, opened 2024-04-19
FAI-10477 - Query Sprints from faros graph for connector WebhookSupplement mode

matiaslcoulougian:
## Description
If run mode is WebhookSupplement and faros client is available, query sprints-board relationships from Faros graph instead of using Jira API.

Also decoupled issues and issue pull requests into separate functions.
## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/src/streams/common.ts
```diff
@@ -81,6 +83,12 @@ export abstract class StreamBase extends AirbyteStreamBase {
     }
     return currentStreamState;
   }
+
+  protected supportsFarosClient(): boolean {
+    return (
+      this.config.run_mode === RunMode.WebhookSupplement && !!this.farosClient
```
chalenge, line 89:
```typescript
&& this.farosClient
```

chalenge (approved):
Minor nits and address SonarCloud issues

sources/jira-source/src/streams/common.ts
matiaslcoulougian: It should be a boolean for it to work, thats why I added the `!!`

PR #1424, opened 2024-04-25
FAI-10151 - Jira Sprints source stream and converter

matiaslcoulougian:
## Description
Jira Sprints source stream and converter and some refactor to sprint-reports stream to remove unnecessary checks and projectKey, and also support update range for faros gql query

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/jira-source/src/models.ts
+++ b/sources/jira-source/src/models.ts
@@ -36,11 +36,20 @@ export interface PullRequest {
 export interface SprintReport {
   readonly id: number;
   readonly boardId?: string;
-  readonly projectKey?: string;
-  readonly completedAt?: Date;
+  readonly closedAt?: Date;
   readonly completedPoints?: number;
   readonly completedInAnotherSprintPoints?: number;
   readonly notCompletedPoints?: number;
   readonly puntedPoints?: number;
   readonly plannedPoints?: number;
+}
+
+export interface Sprint {
+  readonly id: number;
+  readonly boardId?: string;
+  readonly name: string;
+  readonly state: string;
+  readonly startedAt?: Date;
+  readonly endedAt?: Date;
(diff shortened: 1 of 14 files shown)


sources/jira-source/test/__snapshots__/index.test.ts.snap
```diff
@@ -146,14 +145,27 @@ exports[`index streams - sprint_reports with run mode WebhookSupplement using Fa
...
 `;
+
+exports[`index streams - sprints 1`] = `
+[
+  {
+    "boardId": "1",
+    "closedAt": 2024-03-14T12:00:00.000Z,
+    "endedAt": undefined,
+    "id": 1,
+    "name": undefined,
```
ypc-faros, line 166: Can we change the tests setup so that these `undefined` fields are populated?

ypc-faros (approved):
lgtm

Small nit in the tests

sources/jira-source/src/models.ts
```diff
@@ -36,11 +36,20 @@ export interface PullRequest {
...
-  readonly completedAt?: Date;
+  readonly closedAt?: Date;
   readonly completedPoints?: number;
   readonly completedInAnotherSprintPoints?: number;
   readonly notCompletedPoints?: number;
   readonly puntedPoints?: number;
   readonly plannedPoints?: number;
 }
+
+export interface Sprint {
```
chalenge, line 47: @matiaslcoulougian let's use the models from the client. It has all of this already

matiaslcoulougian: 👍

PR #1427, opened 2024-04-29
FAI-10150: Add faros users stream to jira source

chalenge:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/jira/users.ts
```diff
@@ -11,10 +11,10 @@ export class Users extends JiraConverter {
     ctx: StreamContext
   ): Promise<ReadonlyArray<DestinationRecord>> {
     const user = record.record.data;
-    const uid = user.accountId ?? user.name;
+    const uid = user.accountId ?? user.key;
```
matiaslcoulougian, line 14: Is it possible a case where user has neither accountId nor key?

matiaslcoulougian (approved):
LGTM

destinations/airbyte-faros-destination/src/converters/jira/users.ts
chalenge: Highly unlikely but better safe

PR #1430, opened 2024-04-30
FAI-10147 - Jira projects stream

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/resources/schemas/farosProjects.json
```diff
@@ -0,0 +1,18 @@
...
+      "type": "string"
+    },
+    "name": {
+      "type": "string"
+    },
+    "description": {
+      "type": "string"
+    }
+  }
+}
```
chalenge, line 18: nit: newline

destinations/airbyte-faros-destination/src/converters/jira/faros_projects.ts
```diff
@@ -0,0 +1,32 @@
+import {AirbyteRecord} from 'faros-airbyte-cdk';
+import {Utils} from 'faros-js-client';
+
+import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
+import {JiraConverter} from './common';
+
+export class FarosProjects extends JiraConverter {
+  readonly destinationModels: ReadonlyArray<DestinationModel> = ['tms_Project'];
+
+  async convert(
```
chalenge, line 10: why not re-use the Community converter?

sources/jira-source/src/streams/faros_projects.ts
```diff
@@ -0,0 +1,33 @@
...
+    syncMode: SyncMode,
+    cursorField?: string[],
+    streamSlice?: Dictionary<any>,
+    streamState?: Dictionary<any>
+  ): AsyncGenerator<Version2Models.Project> {
+    const jira = await Jira.instance(this.config, this.logger);
+    for await (const project of jira.getProjects()) {
+      // Skip projects that are not in the project_keys list
+      const projectKeys = this.config.project_keys;
+      if (projectKeys && !projectKeys.includes(project.key)) {
```
chalenge, line 27: think we can optimize this not to fetch all projects at least for Jira Cloud but only those specified in the [list using keys parameter](https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-projects/#api-rest-api-2-project-search-get). Will need handling for invalid projects I think but we can test it out

sources/jira-source/test_files/projects.json
```diff
@@ -0,0 +1,20 @@
...
+    "name": "Test 2",
+    "description": "This is a test"
+  },
+  {
+    "key": "TEST-3",
+    "id": 1236,
+    "name": "Test 3",
+    "description": "This is a test"
+  }
+]
```
chalenge, line 20: nit: newline

destinations/airbyte-faros-destination/src/converters/jira/faros_projects.ts
matiaslcoulougian: Because it includes some logic using boardOwnership flag, which we said we wanted to remove to simplify. But if the `useBoardOwnership` flag is not used and its always true by default it should not affect.

sources/jira-source/src/streams/faros_projects.ts
matiaslcoulougian: Tested it out. In the Jira.js library keys is defined as a [string[]](https://mrrefactoring.github.io/jira.js/interfaces/Version2.Version2Parameters.SearchProjects.html#keys) parameter. It works passing only one project in the array, but returns no projects if array contains more than one

matiaslcoulougian: Could not make it work, maybe its a bug in the library when trying formatting the array into the necessary string format required by the Jira API.

matiaslcoulougian: The only alternative I can think of is directly calling the API and passing the `keys=` parameters as described in Jira docs

chalenge (approved):

PR #1432, opened 2024-04-30
FAI-10727 | Vanta Destination Delay Requests

omreego:
## Description
This PR contains several updates.
1. Requests to Faros Graph which match UIDs to Objects are combined into single requests which check for many matches at once. Previously, the destination was failing because of too many consecutive requests the Faros Graph).
2. Vanta Source was updated to check for 504 errors and retry after sleeping.
3. Types fixed to include the various "severity" values (some numerical, some string).
4. Source GraphQL Queries have unused fields removed

## Type of change
- [X] Bug fix
- [x] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/vanta/vulnerabilities.ts
```diff
@@ -124,20 +147,24 @@ export abstract class Vulnerabilities extends Converter {
         this.vcsRepositoryQuery,
         {repoName: vcsRepoName}
       );
+      this.sleep(this.sleepMs, ctx.logger);
```
tovbinm, line 150: Is there a way to detect that you are overwhelming the API? I think you should only sleep when rate limited? or perhaps consider using https://www.npmjs.com/package/p-limit

tovbinm: same below

omreego: Good point - I'll try out that lib

omreego: I found a way to reduce all these queries - we can perform batch searches

omreego: We take a list of say 100 repo names, and insert that as a variable, and search for all vcs Repos where "repo Name in repoNames", that way we'll convert 100 separate queries into 1

destinations/airbyte-faros-destination/src/converters/vanta/README
```diff
@@ -0,0 +1,19 @@
+
```
tovbinm, line 1: rename to README.md

destinations/airbyte-faros-destination/src/converters/vanta/utils.ts
```diff
@@ -37,3 +39,40 @@ export function getQueryFromName(name: string): string {
...
+  variables: Record<string, any>,
+  object_nm: string,
+  inputs: string[]
+): Promise<any> {
+  ctx.logger.info('Running query with backoff');
+  let attempts = 0;
+  const delays = [5, 10, 20, 40]; // Sleep times in seconds
+  while (attempts < delays.length) {
+    try {
+      const resp = await ctx.farosClient.gql(ctx.graph, query, variables);
```
tovbinm, line 59: I dont think you should be reimplementing the retry logic, but use what's available in the underlying library, ie. axios

when creating faros client, you can specify axios config and allow retry behaviour - https://github.com/faros-ai/faros-js-client/blob/main/src/client.ts#L46

tovbinm: or do you not have access to the point of creation of faros client?

omreego: I think I can recreate the faros client with retry behaviour

omreego: Recreated faros Client with retry

destinations/airbyte-faros-destination/src/converters/vanta/vulnerabilities.ts
```diff
@@ -440,53 +586,65 @@ export abstract class Vulnerabilities extends Converter {
...
     ctx: StreamContext
   ): Promise<ReadonlyArray<DestinationRecord>> {
+    const newFarosClientConfig: FarosClientConfig = {
+      url: ctx.config.edition_configs.api_url,
+      apiKey: ctx.config.edition_configs.api_key,
+    };
+    const newAxiosRequestConfig: AxiosRequestConfig = {
+      'axios-retry': {
+        retryCount: 3,
+        lastRequestTime: 5000,
```
tovbinm, line 633: what is this value? `lastRequestTime`

tovbinm: @willmarks / @cjwooo can you please check if this is the best way of creating the Faros Client with retry?

omreego: @tovbinm good question - I think this isn't the correct variable

willmarks: I thought the Faros client already was configured to retry by default? (by default it retries 3 times)

omreego: I see it is configured that way. I think the gateway timeouts must have extended through the retries. 
I think the new update where the requests are reduced by a factor of 100 will fix the issues. So far I've been trying it and it works fine

PR #1445, opened 2024-05-07
FAI-11093 Jira tests -- client

ypc-faros:
Was able to port these unchanged from the feed


ypc-faros:
Only difference is that our client doesn't implement `getStats`. Should I port that functionality over since I'm here @chalenge @matiaslcoulougian ?

ypc-faros:
> Only difference is that our client doesn't implement `getStats`. Should I port that functionality over since I'm here @chalenge @matiaslcoulougian ?

Ported ✅

chalenge (approved):
thank you 🙏🏽

PR #1447, opened 2024-05-08
Refactor getIssues in jira source. Part 1

ypc-faros:
In a separate PR, I'll try to untangle the rest of the params and the corresponding logic (`fetchKeysOnly`, `includeAdditionalFields`, `additionalFields`)

### DIFF
--- a/sources/jira-source/src/streams/faros_issue_pull_requests.ts
+++ b/sources/jira-source/src/streams/faros_issue_pull_requests.ts
@@ -4,6 +4,7 @@ import {Utils} from 'faros-js-client';
 import {Dictionary} from 'ts-essentials';
 
 import {DEV_FIELD_NAME, Jira} from '../jira';
+import {JqlBuilder} from '../jql-builder';
 import {
   ProjectStreamSlice,
   StreamState,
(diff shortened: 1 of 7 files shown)


sources/jira-source/src/streams/faros_issue_pull_requests.ts
```diff
@@ -36,10 +37,11 @@ export class FarosIssuePullRequests extends StreamWithProjectSlices {
         ? this.getUpdateRange(streamState?.[projectKey]?.cutoff)
         : undefined;
     for await (const issue of jira.getIssues(
-      projectKey,
-      updateRange,
+      new JqlBuilder()
+        .withProject(projectKey)
+        .withDateRange(updateRange)
```
chalenge, line 42: Thinking out loud, since this stream will run before or after the issues stream, we are prolly not fully utilizing `@Memoize` on getIssues since the date ranges will likely slight differ forcing us to do a fetch all the issue again. May need to re-think these stream dependencies

chalenge (approved):
🎉

sources/jira-source/src/streams/faros_issue_pull_requests.ts
ypc-faros: 🤔 
I'll look into it 👍

PR #1452, opened 2024-05-10
FAI-11182: Re-use jira instance if exists

chalenge:
## Description

- Prevent re-instantiating the Jira client for every stream slice

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/jira-source/src/jira.ts
+++ b/sources/jira-source/src/jira.ts
@@ -118,6 +118,7 @@ export const DEFAULT_API_URL = 'https://prod.api.faros.ai';
 export const DEFAULT_GRAPH = 'default';
 
 export class Jira {
+  private static jira: Jira;
   private readonly fieldIdsByName: Map<string, string[]>;
   // Counts the number of failed calls to fetch sprint history
   private sprintHistoryFetchFailures = 0;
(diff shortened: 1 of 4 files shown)


sources/jira-source/src/jira.ts
```diff
@@ -267,17 +271,6 @@ export class Jira {
     }
   }
 
-  private static updatedBetweenJql(range: [Date, Date]): string {
```
chalenge, line 270: moved to jqlBuilder

sources/jira-source/test/index.test.ts
```diff
@@ -362,22 +362,23 @@ describe('index', () => {
...
     await testStreamSlices({
       ...config,
       bucket_total: 2,
       bucket_id: 1,
-      projects,
+      projects: ['TEST', 'TEST2', 'TEST3'],
     });
+  });
 
+  test('stream with project slices using bucket 2', async () => {
```
chalenge, line 375: @ypc-faros had to break up the test due to the singleton being re-used. Any ideas on how to force recreation?

cjwooo (approved):

PR #1456, opened 2024-05-13
Fuse the calls to `searchForIssuesUsingJql` when possible + memoization + cleanups

ypc-faros:
Splits the former `getIssues` into 3 functions:

- `getIssuesKeys`: only fetches issue `key`(s) => used only by the `faros_board_issues` stream
- `getIssues`: fetches all the fields needed for the `faros_issues` stream (if requested) and all the fields needed for the `faros_issue_pull_requests` (if requested). This function also stores a projection of the issue (only the fields needed for `faros_issue_pull_requests`) in memory (`seenIssues`). I decided to use my own data structure vs `@Memoized` so that the memory usage was minimized.
- `getIssuesCompact`: fetches the fields needed for `faros_issue_pull_requests` (from memory or from the api)

Aside from the above, a couple of minor changes:
- Made sure that the `jql`(s) for `getIssues` and `getIssuesCompact` match, by storing `start_date` and `end_date` in the "global" config.
- Major cleanup in how we calculate which fields should be fetched, depending on the streams requested.
- Bug fix in the toposort (the reverse was not needed)


sources/jira-source/src/jira.ts
```diff
@@ -624,24 +647,71 @@ export class Jira {
...
       async (item: any) => {
+        this.memoizeIssue(item, jql);
+
         return issueTransformer.toIssue(item);
       },
       'issues'
     );
   }
 
+  private memoizeIssue(item: any, jql: string) {
```
chalenge, line 669: is there a way we can generalize this? We kinda have the same need for Sprint and Sprint Reports so wondering if we would also need to do the same for that

ypc-faros: Maybe..., let's chat offline

sources/jira-source/src/jira.ts
```diff
@@ -52,6 +54,9 @@ export interface JiraConfig extends AirbyteConfig {
   readonly api_url?: string;
   readonly api_key?: string;
   readonly graph?: string;
+  readonly requestedStreams?: Set<string>;
```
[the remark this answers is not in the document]
matiaslcoulougian: I like more the option of using the `run_mode`. We also avoid adding yet another config

chalenge (approved):

PR #1458, opened 2024-05-13
FAI-11182: Jira source when project list is supplied only fetch those

chalenge:
## Description

- Optimize calls to only fetch the listed projects

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/jira-source/src/jira.ts
+++ b/sources/jira-source/src/jira.ts
@@ -17,7 +17,7 @@ import * as fs from 'fs';
 import parseGitUrl from 'git-url-parse';
 import https from 'https';
 import jira, {AgileModels, Version2Models} from 'jira.js';
-import {concat, isNil, pick, toInteger, toLower, toString} from 'lodash';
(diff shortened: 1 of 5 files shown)


sources/jira-source/src/jira.ts
```diff
@@ -184,6 +184,7 @@ export class Jira {
           'X-Force-Accept-Language': true,
         },
         timeout: cfg.timeout ?? DEFAULT_TIMEOUT,
+        paramsSerializer: {indexes: null},
```
matiaslcoulougian, line 187: Is this config added for making keys param work in searchProjects?

matiaslcoulougian (approved):
LGTM 🚀

sources/jira-source/src/jira.ts
chalenge: Yes, the default behaviour results in incorrect query params. See https://github.com/axios/axios/issues/5058#issuecomment-1272229926

PR #1459, opened 2024-05-13
FAI-11383: CircleCI Memory fix

willmarks:
## Description

Currently, the CircleCI converter stores all test executions it sees so that it can aggregate all of the test case result stats based off of test records it converts. This has caused issues where the destination will run out of memory. However, we can instead only worry about on test execution at a time because the source emits records related to a single test execution in order. Since we can assume that all test case results related to a test execution will be processed in order before moving onto the next test execution, we can aggregate one TestExecution at a time until we  see the first record that is from a new TestExecution at which point we send in the TestExecution and start aggregating the next. The final TestExecution needs to be sent in the onProcessComplete logic

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/circleci/tests.ts
+++ b/destinations/airbyte-faros-destination/src/converters/circleci/tests.ts
@@ -1,6 +1,5 @@
 import {AirbyteRecord} from 'faros-airbyte-cdk';
 import {Utils} from 'faros-js-client';
-import {Dictionary} from 'ts-essentials';
 
 import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
 import {CircleCICommon, CircleCIConverter} from './common';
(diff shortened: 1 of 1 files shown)


tovbinm (approved):

destinations/airbyte-faros-destination/src/converters/circleci/tests.ts
```diff
@@ -141,22 +141,30 @@ export class Tests extends CircleCIConverter {
...
+          test.pipeline_vcs,
+          test.project_slug
+        ),
+      };
+      // Reset test case results set
+      this.testCaseResults.clear();
     } else {
-      te.testCaseResultsStats.unknown += 1;
+      if (testCaseResultStatus.category === 'Failure') {
+        this.currentTestExecution.status = {category: 'Failure', detail: null};
```
tovbinm, line 155: why is this necessary?

willmarks: Its cause as soon as we see one failure the whole test execution should be marked as failed. Also, I am just copying what we had already.

PR #1466, opened 2024-05-17
FAI-11182: Fix sprint state to be for in context board

chalenge:
## Description

- Sprints can be associated with multiple boards other than the origin board. State should not be associated with origin board as we may not have synced that. Use the board in context for the state
- https://support.atlassian.com/jira-software-cloud/docs/view-and-understand-the-sprint-report/

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


sources/jira-source/src/streams/faros_sprints.ts
```diff
@@ -38,13 +38,14 @@ export class FarosSprints extends StreamWithBoardSlices {
...
         id: sprint.id,
-        originBoardId: sprint.originBoardId ?? toInteger(boardId),
+        originBoardId: sprint.originBoardId,
         name: sprint.name,
         state: sprint.state,
         startDate: sprint.startDate,
         endDate: sprint.endDate,
         completeDate: sprint.completeDate,
         activatedDate: sprint['activatedDate'],
+        boardId: toInteger(sprint.originBoardId),
```
matiaslcoulougian, line 48: shouldn't this be `boardId: toInteger(boardId)` ?

matiaslcoulougian (changes requested):

matiaslcoulougian (approved):

PR #1472, opened 2024-05-22
FAI-11490: Add option to sync closed reports starting with most recent

chalenge:
## Description

- For extremely large Jira instances where we can fetch sprints normally in the order they're are received, fetch sprints starting from the most recent in the backlog to make the fetch faster since the api does not support [time-based filtering](https://community.developer.atlassian.com/t/is-there-any-api-to-retrieve-the-list-of-sprints-in-a-month/14519) and only returns only [max of 50](https://community.developer.atlassian.com/t/is-it-possible-to-pull-a-list-of-sprints-in-a-project-via-the-rest-api/53336/3).

## Type of change
- [X] Bug fix
- [x] New feature
- [ ] Breaking change


matiaslcoulougian (approved):
LGTM

PR #1473, opened 2024-05-23
FAI-11519 | Refactor getBoards logic in project-board-filter

matiaslcoulougian:
## Description
Refactor getBoards to request boards by project key, so we avoid using board location which is not populated for Jira Server/Data Center. Also fixed some typos. 

Note: with this change we are increasing the amount of request for getting boards, from one requests (or more depending on the size) for getting all boards to one request (or more) per project to get its boards.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/src/project-board-filter.ts
```diff
@@ -48,45 +48,50 @@ export class ProjectBoardFilter {
...
-              this.farosClient,
-              this.config.graph ?? DEFAULT_GRAPH
-            )
-          : jira.getBoards();
+
+      // Ensure projects is populated
+      await this.getProjects();
+
+      if (this.supportsFarosClient()) {
+        const boards = jira.getBoardsFromGraph(
```
chalenge, line 56: `const boards = await jira.getBoardsFromGraph(`

sources/jira-source/src/project-board-filter.ts
```diff
@@ -48,45 +48,50 @@ export class ProjectBoardFilter {
...
+          ) {
+            this.boards.add(board.id.toString());
+          }
         }
       } else {
-        for (const board of this.config.boards) {
-          if (await jira.isBoardInBucket(board)) {
-            await this.maybeIncludeBoard(board);
+        for (const project of this.projects) {
+          for await (const board of await jira.getBoards(project)) {
```
chalenge, line 79: remove first await

sources/jira-source/src/project-board-filter.ts
```diff
@@ -48,45 +48,50 @@ export class ProjectBoardFilter {
...
-          ? jira.getBoardsFromGraph(
-              this.farosClient,
-              this.config.graph ?? DEFAULT_GRAPH
-            )
-          : jira.getBoards();
+
+      // Ensure projects is populated
+      await this.getProjects();
+
+      if (this.supportsFarosClient()) {
```
chalenge, line 55: extract this into two functions one handling the `supportsFarosClient`. Hard to follow

faros-airbyte-common/src/jira/types.ts
```diff
@@ -130,3 +130,8 @@ export interface Board extends AgileModels.Board {
   uid: string;
   projectKey: string;
 }
+
+export interface FarosBoard {
+  id: number;
```
chalenge, line 135: this is a string if it is coming from faros graph

sources/jira-source/src/project-board-filter.ts
```diff
@@ -48,45 +48,50 @@ export class ProjectBoardFilter {
...
-          await this.maybeIncludeBoard(board.id.toString());
+          // If boards are specified, only include those
+          if (
+            this.config.boards &&
+            !this.config.boards.includes(board.id.toString())
+          ) {
+            continue;
+          }
+          // Check if at least one board project key is in the projects list
+          if (
```
chalenge, line 69: if we using `getBoardsFromGraph` only here to get boards then this logic will be much simpler if we invert the query and query for projects and the boards under it
```gql
tms_Project {
  uid
  boards
  ...
}

File: sources/jira-source/src/streams/faros_boards.ts [hunk 6]
```diff
@@ -33,7 +33,7 @@ export class FarosBoards extends StreamWithProjectSlices {
       projectKey: this.projectKey,
     };
 
-    for await (const board of jira.getBoards(this.projectKey)) {
+    for await (const board of await jira.getBoards(this.projectKey)) {
```
chalenge, line 36: remove extra await, what am I missing?

sources/jira-source/src/jira.ts
```diff
@@ -809,52 +824,58 @@ export class Jira {
...
 
-  async *getBoardsFromGraph(
+  // Fetch project with boards nested from Faros
+  async *getProjectBoardsFromGraph(
     farosClient: FarosClient,
-    graph: string
-  ): AsyncIterableIterator<AgileModels.Board> {
-    const boards = this.iterate(
+    graph: string,
+    projectIds: Array<string>
```
chalenge, line 856: projectKeys

faros-airbyte-common/src/jira/types.ts
```diff
@@ -139,3 +139,8 @@ export interface IssueProjectVersion {
   readonly key: string;
   readonly projectVersionId: string;
 }
+
+export interface FarosProject {
+  key: string;
+  boardIds: string[];
```
chalenge, line 145: boardUids

sources/jira-source/src/streams/faros_projects.ts
```diff
@@ -25,7 +25,7 @@ export class FarosProjects extends StreamBase {
     const projectKeys = this.config.projects?.length
       ? new Set(this.config.projects.map((key) => toUpper(key)))
       : undefined;
-    for await (const project of jira.getProjects(projectKeys)) {
+    for await (const project of await jira.getProjects(projectKeys)) {
```
chalenge, line 28: remove extra await

sources/jira-source/src/jira.ts
```diff
@@ -809,52 +824,58 @@ export class Jira {
...
           source: 'Jira',
           offset: startAt,
           pageSize: this.maxPageSize,
+          projects: projectIds,
         });
-        return data?.tms_TaskBoard;
+        return data?.tms_Project;
       },
-      async (item: any) => {
+      async (item: any): Promise<FarosProject> => {
```
chalenge, line 868: this should just be `(item: any): FarosProject`

sources/jira-source/src/streams/faros_board_issues.ts
```diff
@@ -27,7 +27,7 @@ export class FarosBoardIssues extends StreamWithBoardSlices {
     const boardJql = await jira.getBoardJQL(boardConfig.filter.id);
     // Jira Agile API GetConfiguration response type does not include key property
     // but, in practice, it is actually present
-    const projectKey = boardConfig.location['key'];
+    const projectKey = boardConfig.location?.['key'];
```
chalenge, line 30: this then doesn't work for Jira Server right?

matiaslcoulougian: Yes, projectKey will be null always. I'm wondering, if we are already using board's JQL, why do we need to add the project key condition?

matiaslcoulougian: I think if the board's JQL does not include a project, board might contain issues from more than one project, so we should not restrict the JQL to only one project (which would be board's location)

chalenge (approved):
LGTM, after fixing the board JQL

PR #1474, opened 2024-05-23
More test utilities refactor + `Readable` `read` bug fix

ypc-faros:
Refactored `destinationWriteTest` so that we no longer need to pass the expected stats. Instead, we extract the relevant lines from the logs and compare to a snapshot.

Also added `generateBasicTestSuite` to generate a test suite containing the common basic test that runs the destination and checks the stats in the logs only.

Fixes a bug where we were not always fully reading the process stdout (I was experiencing flakiness due to this). Found `consumers.text` which does the right thing and as a bonus runs way faster.


destinations/airbyte-faros-destination/test/converters/abstract-surveys.test.ts
```diff
@@ -54,27 +50,10 @@ describe('abstract-surveys', () => {
...
-
-    await destinationWriteTest({
-      configPath,
-      catalogPath,
-      streamsLog: airtableSurveysAllStreamsLog,
-      streamNamePrefix,
-      expectedProcessedByStream,
-      expectedWrittenByModel,
-    });
+  generateBasicTestSuite({
```
chalenge, line 53: ❤️

chalenge (approved):
NIce ⭐

PR #1483, opened 2024-05-29
FAI-11445 | Support start_date and end_date parameters for backfilling

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/jira-source/src/jira.ts
+++ b/sources/jira-source/src/jira.ts
@@ -24,7 +24,6 @@ import moment from 'moment';
 import pLimit from 'p-limit';
 import path from 'path';
 import {Memoize} from 'typescript-memoize';
-import * as util from 'util';
 import {VError} from 'verror';
 
 import {JiraClient} from './client';
--- a/sources/jira-source/src/jira.ts
+++ b/sources/jira-source/src/jira.ts
@@ -24,6 +24,7 @@ import moment from 'moment';
 import pLimit from 'p-limit';
 import path from 'path';
 import {Memoize} from 'typescript-memoize';
+import * as util from 'util';
 import {VError} from 'verror';
 
 import {JiraClient} from './client';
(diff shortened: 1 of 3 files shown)


sources/jira-source/src/jira.ts
```diff
@@ -24,6 +24,7 @@ import moment from 'moment';
 import pLimit from 'p-limit';
 import path from 'path';
 import {Memoize} from 'typescript-memoize';
+import * as util from 'util';
```
chalenge, line 27: remove unused import

sources/jira-source/src/jira.ts
```diff
@@ -244,11 +245,29 @@ export class Jira {
       }
     }
 
-    cfg.end_date = moment().utc().toDate();
-    cfg.start_date = moment()
-      .utc()
-      .subtract(cfg.cutoff_days || DEFAULT_CUTOFF_DAYS, 'days')
-      .toDate();
+    if ((cfg.start_date || cfg.end_date) && cfg.cutoff_days) {
```
chalenge, line 248: How are we ensuring the backfill does not overwrite the current state of the source? @cjwooo / @ypc-faros do we have a way to ignore saving state?

chalenge (commented):
Need to handle state so we don't overwrite it

sources/jira-source/src/jira.ts
ypc-faros: We should implement [supportsIncremental](https://github.com/faros-ai/airbyte-connectors/blob/ac28bee95dfd76a01f97c65eb00c5800fcb3c692/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L77) returning `false` when we're in backfill mode

matiaslcoulougian: Hm, I think it will still override it with an empty state: [1](https://github.com/faros-ai/airbyte-connectors/blob/ac28bee95dfd76a01f97c65eb00c5800fcb3c692/faros-airbyte-cdk/src/sources/source-base.ts#L300),[2](https://github.com/faros-ai/airbyte-connectors/blob/ac28bee95dfd76a01f97c65eb00c5800fcb3c692/faros-airbyte-cdk/src/sources/source-base.ts#L339)

ypc-faros: You're right, it's not so simple 😬 
We'd also need to make sure that the destination doesn't reset the models for the origin

I think we should first get alignment on what backfill means and what we want to happen.

We have 2 options:

1) **Quick and dirty**: We assume that a backfill sync will run independently of the regular sync (that's what I've been doing). In this case we don't need to worry about anything: there's no state to ignore and the destination reset doesn't delete anything since the connections (should) have a different origin (we'd probably need to document this)
2) **Slow and clean** We add support for backfill syncs to the cdk (maybe a flag in the source config?). When in backfill mode we a) make sure that the source ignores any state. b) make sure that the destination DOES NOT reset any models. c) make sure that the destination DOES NOT yield any state.

What do you think @cjwooo @chalenge @matiaslcoulougian ?

ypc-faros: Chatted offline with @cjwooo who also likes option 2

matiaslcoulougian: Like option 2 as well

matiaslcoulougian:
Moved to draft as we will work on a more generic solution and probably refactor this logic.

PR #1490, opened 2024-05-31
Support backfills

ypc-faros:
## Description

Adds support for backfills. When a source is configured with `backfill`, the streams state will be ignored and unmodified. The destination will NOT reset the models. This way, one can temporarily reuse a configured sync for backfill, by setting `backfill` to `true` and providing `start_date` and `end_date`. Once the backfill is completed, one can turn `backfill` off and the configured sync will continue normally from where it left off.

Also adds some utility function to deal with calculating the date range to sync based on `start_date`, `end_date` and `cutoff_days`

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/faros-airbyte-cdk/src/sources/source-base.ts
+++ b/faros-airbyte-cdk/src/sources/source-base.ts
@@ -1,4 +1,4 @@
-import {keyBy} from 'lodash';
+import {keyBy, pick} from 'lodash';
 import toposort from 'toposort';
 import VError from 'verror';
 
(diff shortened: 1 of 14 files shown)


faros-airbyte-cdk/src/sources/source-base.ts
```diff
@@ -26,6 +26,11 @@ import {AirbyteSource} from './source';
 import {State} from './state';
 import {AirbyteStreamBase} from './streams/stream-base';
 
+type PartialAirbyteConfig = Pick<
+  AirbyteConfig,
+  'backfill' | 'max_slice_failures'
```
chalenge, line 31: can't we have both or we don't have want backfills to be always work?

chalenge (approved):
🚀

faros-airbyte-cdk/src/sources/source-base.ts
ypc-faros: Both `backfill` and `max_slice_failures` are included in `PartialAirbyteConfig`
This also confuses me 😬 
https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

PR #1491, opened 2024-05-31
FAI-11632 - Scale heap size with available memory

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,8 @@ RUN apk add --no-cache --virtual .gyp python3 make g++ \
     && npm install -g npm lerna @lerna/legacy-package-management tsc
 RUN lerna bootstrap --hoist
 
+COPY ./docker ./docker
+
 ARG version
 RUN test -n "$version" || (echo "'version' argument is not set, e.g --build-arg version=x.y.z" && false)
 ENV CONNECTOR_VERSION $version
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -14,6 +14,8 @@ RUN apk add --no-cache --virtual .gyp python3 make g++ \
     && npm install -g npm lerna @lerna/legacy-package-management tsc
 RUN lerna bootstrap --hoist
 
+COPY ./docker ./docker
```
tovbinm, line 17: what's in this folder?

docker/entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Assign 75% of available memory to nodejs heap
+TOTAL_MEM=$(grep MemTotal /proc/meminfo | awk '{print $2}')
+TOTAL_MEM_MB=$(($TOTAL_MEM / 1024))
+MAX_OLD_SPACE_SIZE=$(($TOTAL_MEM_MB * 75 / 100))
```
tovbinm, line 6: perhaps we should make `75` to be passed as an arg into the docker?

tovbinm (approved):

PR #1502, opened 2024-06-03
FAI-10892 | Jira Additional Fields stream

matiaslcoulougian:
## Description
Stream to populate tasks additional fields when they are updated. We get Jira issues only fetching key and additional fields and push them to update faros task additional fields. This way we support doing a backfill only of additional fields for historical tasks. Works under `RunMode=AdditionalFields`

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


ypc-faros:
Hmmm I took a quick look at the additional fields stream and I think it looks conceptually correct but I think we can simplify. My understanding of the proposed solution here:
We query the Faros graph for the potential tasks that need to be updated with additional fields (these are kept in memory -- potentially pretty expensive).
We get the jira issues and for issues in the intersection with the tasks above, determine whether they should be updated (if the set of additional fields differs)

My proposed simplification:
In the source: we DO NOT query the Faros graph. We only get the issues additional fields from jira for the time range and output an object with the jira issue key + additional fields.
In the destination: we ALWAYS write the `tms_Task__Update` . I think the majority of the updates will fall in the bucket of tasks who actually needed to be updated. There'll be a small bucket of tasks who might not need the update (i.e., they already had the new set of additional fields) => nothing happens (this will probably be offset in terms of cost by the query that we're NOT making to the graph in the source vs the previous solution).

sources/jira-source/src/jira.ts
```diff
@@ -698,9 +698,12 @@ export class Jira {
       this.fieldIdsByName,
       this.statusByName,
       additionalFieldIds,
-      this.additionalFieldsArrayLimit
+      this.additionalFieldsArrayLimit,
+      this.logger
     );
 
+    this.logger.info(`ADDITIONAL FIELDS IN GET ISSUES: ${additionalFieldIds}`);
```
ypc-faros, line 705: Remove

sources/jira-source/src/jira.ts
```diff
@@ -763,6 +766,27 @@ export class Jira {
...
+      this.fieldNameById,
+      this.fieldIdsByName,
+      this.statusByName,
+      additionalFieldIds,
+      this.additionalFieldsArrayLimit
+    );
+    for await (const issue of issues) {
+      yield {
+        key: issue.key,
+        additionalFields: issueTransformer.extractAdditionalFields(issue),
```
ypc-faros, line 785: Nice!

sources/jira-source/src/jira.ts
```diff
@@ -815,14 +839,19 @@ export class Jira {
         'updated',
       ].forEach((field) => fieldIds.add(field));
     }
+    const additionalFieldIds: string[] = this.getAdditionalFieldIds(fieldIds);
+    return {fieldIds: Array.from(fieldIds), additionalFieldIds};
+  }
+
+  private getAdditionalFieldIds(fieldIds: Set<string>): string[] {
```
ypc-faros, line 846: nit: Do not extract this method if we're not gonna reuse it. Or rename `fieldIds` to `skipFieldIds`

sources/jira-source/src/streams/faros_issue_additional_fields.ts
```diff
@@ -0,0 +1,42 @@
...
+    const projectKey = streamSlice?.project;
+    const updateRange = this.getUpdateRange();
+    const issues = jira.getIssueCompactWithAdditionalFields(
+      new JqlBuilder()
+        .withProject(projectKey)
+        .withDateRange(updateRange)
+        .build()
+    );
+    for await (const issue of issues) {
+      yield omit(issue, 'fields');
```
ypc-faros, line 39: No need to omit `fields` since you're yielding an object with `key` and `additionalFields` in `getIssueCompactWithAdditionalFields`

sources/jira-source/src/index.ts
```diff
@@ -85,12 +87,10 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
     catalog: AirbyteConfiguredCatalog;
     state?: AirbyteState;
   }> {
-    let streams = catalog.streams;
-    if (config.run_mode === RunMode.WebhookSupplement) {
-      streams = streams.filter((stream) =>
-        WebhookSupplementStreamNames.includes(stream.stream.name)
-      );
-    }
+    const streamNames = RunModeStreams[config.run_mode];
```
ypc-faros, line 90: Do we need to fallback to the `Full` here? 
`const streamNames = RunModeStreams[config.run_mode];` => `const streamNames = RunModeStreams[config.run_mode ?? RunMode.Full];`

destinations/airbyte-faros-destination/src/converters/jira/faros_issue_additional_fields.ts
```diff
@@ -0,0 +1,32 @@
+import {AirbyteRecord} from 'faros-airbyte-cdk';
+
+import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
+import {JiraConverter} from './common';
+
+export class FarosIssueAdditionalFields extends JiraConverter {
+  readonly destinationModels: ReadonlyArray<DestinationModel> = [
+    'tms_Task__Update',
```
ypc-faros, line 8: Remove. Not needed.

We actually don't want to reset any models under any circumstance when this stream is synced.

sources/jira-source/src/index.ts
matiaslcoulougian: Yes, good catch

sources/jira-source/src/jira.ts
```diff
@@ -832,7 +854,7 @@ export class Jira {
     }
     const additionalFieldIds: string[] = [];
     for (const fieldId of this.fieldNameById.keys()) {
-      // Skip fields that are already included in the fields above
+      // Skip fields that are already included in the fieldIds set
```
ypc-faros, line 857: nit: Revert comment

ypc-faros (approved):
Niceeee 🥳 !

PR #1503, opened 2024-06-03
Removing unused dataquality tests

omreego:
## Description
It seems that the data generated by the queries which are computing Z Scores and more sophisticated data aren't being used, so let's save the computing space for now.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


tovbinm (commented):
just remove them completely or put them behind an argument(s)

PR #1506, opened 2024-06-05
FAI-11734: Asana source fetch workspace users with /users endpoint

chalenge:


sources/asana-source/src/asana.ts
```diff
@@ -286,14 +286,16 @@ export class Asana {
 
   async *getUsers(workspace: string): AsyncGenerator<User> {
     const opt_fields = ['email', 'name'];
-    yield* this.fetchData<User>(`workspaces/${workspace}/users`, opt_fields);
+    yield* this.fetchData<User>(`/users`, opt_fields, {workspace});
```
ypc-faros, line 289: Do we need the forward slash? It works without, maybe it works as well but no need to risk it 😄

ypc-faros (approved):

sources/asana-source/src/asana.ts
chalenge: Removed, better safe

PR #1509, opened 2024-06-05
FAI-11476: Xray source - test executions

chalenge:
## Description

- Add test executions stream
- Add test runs stream
- Add steps to tests

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/xray-source/resources/queries/get-tests.gql
```diff
@@ -22,6 +22,10 @@ query getTests(
       status {
         name
       }
+      steps {
```
matiaslcoulougian, line 25: Add steps in `getTests()` yield record and `Test` type

matiaslcoulougian (approved):
LGTM after addressing comment

PR #1518, opened 2024-06-10
FAI-11030: Re-fetch additional fields when config is updated

matiaslcoulougian:
## Description

We augmented the state of the `faros_issues` stream with (per project) 1) the earliest update timestamp of issue pulled (in order to know how far back we need to re-fetch) and 2) the set of additional fields that we've pulled.

When the current set of additional fields differs from what's in the state, it means we need to re-fetch the additional fields for the project. We pull the additional fields from the timestamp described in (1) to the cutoff, since the stream itself is already taking care of [cutoff - end date (or $now) )

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/src/index.ts
```diff
@@ -95,6 +101,17 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
     if (config.fetch_teams) {
       streamNames.push(...TeamStreamNames);
     }
+    // Check if the additional fields changed from last sync. If so, add the additional fields stream to update tasks
+    if (config.additional_fields && config.sync_additional_fields) {
+      const additionalFieldsFromState =
+        state?.['faros_issue_additional_fields']?.additionalFields;
+      if (
+        !additionalFieldsFromState ||
```
chalenge, line 109: Not sure, I am too keen on this. In the initial run this will cause the source to run this stream even though we will be pulling these additional fields in the Issues stream. I am thinking this should be on the issues_stream instead and only run it if we have state. @ypc-faros thoughts

Also how is state passed for full mode?

sources/jira-source/src/streams/faros_issue_additional_fields.ts
```diff
@@ -23,7 +23,7 @@ export class FarosIssueAdditionalFields extends StreamWithProjectSlices {
     syncMode: SyncMode,
     cursorField?: string[],
     streamSlice?: ProjectStreamSlice,
-    streamState?: StreamState
+    streamState?: AdditionalFieldsStreamState
   ): AsyncGenerator<IssueCompact> {
     const jira = await Jira.instance(this.config, this.logger);
     const projectKey = streamSlice?.project;
```
chalenge, line 29: For line below. `this.getUpdateRange()`. Ideally we should be able to sync the updateRange not for the current 90-day window but for the whole timeRange of issues we have, e.g. if we have pulled data for 1-year. More of a nice-to-have but something to think of. Maybe store the now timestamp of when we initially sync 😬

chalenge (commented):
Let's make it work without having to run the additional fields stream initially

sources/jira-source/src/streams/faros_issue_additional_fields.ts
ypc-faros: +1 but I'd say this is a hard requirement, not a nice to have

ypc-faros:
This is the simplest approach I can think of:

1. We leave the `faros_issue_additional_fields` as is (only used when run mode is `AdditionalFields`).
2. We augment the `faros_issues` stream state with a) the oldest date pulled and b) the set of additional fields pulled (per project)
3. In the `faros_issues` stream itself, besides pulling issues, we reconcile the additional fields per project (for each project, if the set of additional fields pull differs from the set in the current config, we pull additional fields from `oldest date` to the start of the current range and update the state for that project)

It might feel kind of weird to emit records of the same shape as `faros_issue_additional_fields` from `faros_issues` (we might need to include a boolean or something indicating that the record is just an additional fields update for the task, and reuse some code in the converters). However, I think the alternative would be even more complicated (it'd involve sharing/updating state between streams)

ypc-faros:
Made some minor changes. PTAL 
I think this is good to go since you've already done some testing 🥳

dbruno21 (approved):

PR #1521, opened 2024-06-10
FAI-11476: Update Xray source spec 

chalenge:
## Description

- Update Xray source spec 
- Add support for failed requests 

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


sources/xray-source/resources/spec.json
```diff
@@ -2,23 +2,50 @@
...
+      "projects": {
+        "order": 3,
+        "type": "array",
+        "title": "Project Keys or IDs",
+        "description": "List of Jira project Keys or IDs to sync.",
+        "items": {
+          "type": "string"
+        }
+      },
+      "cutoff_days": {
```
ypc-faros, line 37: Is `cutoff_days` used?

sources/xray-source/src/xray.ts
```diff
@@ -32,18 +33,13 @@ export class Xray {
...
-
-    if (!config.client_id || !config.client_secret) {
-      throw new VError(
-        'Please provide Xray Cloud authentication details, ' +
-          'Client Id and a Client Secret'
-      );
-    }
-
+    const timeout = config.timeout
+      ? Math.min(config.timeout, XRAY_DEFAULT_TIMEOUT)
```
ypc-faros, line 37: Why do we have to do this? If this is a limitation enforced by their api, let's document it in the spec

ypc-faros (approved):

sources/xray-source/resources/spec.json
chalenge: Removed, have another PR where I will re-add it

sources/xray-source/src/xray.ts
chalenge: Not supposed to be here 🤦🏽

PR #1524, opened 2024-06-11
FAI-11735 - Upload logs without hash verification if needed

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/sync.ts
+++ b/destinations/airbyte-faros-destination/src/sync.ts
@@ -200,7 +200,7 @@ class FarosSyncClient extends FarosClient {
       }),
       'Failed to upload sync logs'
     );
-    if (!uploadResp) {
+    if (!uploadResp && !skipVerify) {
       await this.uploadLogs(accountId, syncId, logs, true);
     }
     this.airbyteLogger?.debug('Finished uploading sync logs');
--- a/destinations/airbyte-faros-destination/src/sync.ts
+++ b/destinations/airbyte-faros-destination/src/sync.ts
@@ -9,6 +9,8 @@ import {DEFAULT_AXIOS_CONFIG} from 'faros-js-client/lib/client';
 import {Dictionary} from 'ts-essentials';
 import {VError} from 'verror';
 
+import {LogContent} from './destination-logger';
+
 export interface Account {
   accountId: string;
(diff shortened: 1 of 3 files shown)


destinations/airbyte-faros-destination/src/sync.ts
```diff
@@ -165,23 +167,43 @@ class FarosSyncClient extends FarosClient {
...
+    this.airbyteLogger?.debug(
+      'Uploading sync logs' +
+        (skipVerify ? ' (skipping hash verification)' : '')
+    );
+    const urlResp: {uploadUrl: string} = await this.attemptRequest(
+      this.request('POST', `/accounts/${accountId}/syncs/${syncId}/logs`, {
+        hash: skipVerify ? undefined : logs.hash,
+      }),
+      `Failed to generate log file url for sync ${syncId} for account ${accountId}`
+    );
```
willmarks, line 185: maybe it would make sense to make this call only once rather than a second time when skipVerify? We could allow the url to optionally be passed in and we only make this call when it is not passed?

destinations/airbyte-faros-destination/src/sync.ts
```diff
@@ -165,23 +167,43 @@ class FarosSyncClient extends FarosClient {
...
-          'content-md5': hash,
+          'content-length': logs.content.length,
+          'content-md5': skipVerify ? undefined : logs.hash,
           'content-type': 'text/plain; charset=UTF-8',
         },
       }),
       'Failed to upload sync logs'
     );
-    this.logger.debug('Finished uploading sync logs');
+    if (!uploadResp) {
```
willmarks, line 203: Can we get stuck in a weird infinite loop? Maybe we should add `&& !skipVerify` to make sure we only try the recursion once?

destinations/airbyte-faros-destination/src/sync.ts
cjwooo: The hash is consumed by AWS API to generate the S3 presigned url, so we have to generate a distinct one without using the hash here.

willmarks (approved):

PR #1531, opened 2024-06-12
FAI-11836: Resets fail w/ HTTP 504

ted-faros:
## Description

Break non-incremental model reset queries into paginated delete by id queries

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -377,23 +385,49 @@ export class GraphQLClient {
...
-        [`delete_${model}`]: {
+      const query = {
+        [model]: {
           __args: {
             where: {
               _and: {...deleteConditions},
             },
           },
-          affected_rows: true,
+          id: true,
```
cjwooo, line 395: How does this line get translated into "return the id field"? (confused about the `true` portion)

cjwooo (approved):

ypc-faros (approved):
Can we also add this setting to the [spec](https://github.com/faros-ai/airbyte-connectors/blob/9eb8ad6b43ced6442a12d186703698304d18b8a0/destinations/airbyte-faros-destination/resources/cloud.json#L41) and wire it through?

destinations/airbyte-faros-destination/src/common/graphql-client.ts
ted-faros: that is how the library works: https://www.npmjs.com/package/json-to-graphql-query#query-with-nested-objects

ted-faros:
> Can we also add this setting to the [spec](https://github.com/faros-ai/airbyte-connectors/blob/9eb8ad6b43ced6442a12d186703698304d18b8a0/destinations/airbyte-faros-destination/resources/cloud.json#L41) and wire it through?

Added.  Can you take another look @ypc-faros ?

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -175,6 +175,12 @@
         "title": "Fail on source error",
         "description": "Fail the sync if the source has an error, even if it's not a fatal error.",
         "default": false
+      },
+      "reset_page_size": {
```
ypc-faros, line 179: nit: Add `"order": 17,`

ypc-faros (approved):

PR #1543, opened 2024-06-21
FAI-12009 | GitHub auth app params cleanup

dbruno21:
## Description

- Removed installation_id from app auth params
  - Replaced with multiple octokit instances (one per installation, which makes sense since quota is per installation as well)
- Removed client_id + client_secret from app auth params
  - Not used

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/github-source/resources/spec.json
```diff
@@ -174,25 +116,24 @@
         "description": "GitHub API previews. See https://docs.github.com/en/enterprise-server@3.1/rest/overview/api-previews",
         "items": {
           "type": "string"
-        },
-        "order": 4
+        }
       },
       "api_url": {
-        "order": 5,
+        "order": 6,
```
chalenge, line 122: why do we need this in the source?

sources/github-source/src/github.ts
```diff
@@ -25,32 +26,57 @@ export const DEFAULT_GRAPH = 'default';
 export const PAGE_SIZE = 100;
 
 export class GitHub {
-  private static github: GitHub;
+  private static _instance: GitHub;
```
chalenge, line 29: let's keep this as github for consistency with other sources

sources/github-source/src/github.ts
```diff
@@ -25,32 +26,57 @@ export const DEFAULT_GRAPH = 'default';
...
   ) {}
 
   static async instance(
     cfg: GitHubConfig,
     logger: AirbyteLogger
   ): Promise<GitHub> {
-    if (GitHub.github) return GitHub.github;
-
-    const octokit = makeOctokitClient(cfg, logger);
+    let instance: GitHub = GitHub._instance;
```
chalenge, line 42: let's invert this and avoid lots of code in if statement and exit early

sources/github-source/src/github.ts
```diff
@@ -64,9 +90,9 @@ export class GitHub {
   }
 
   async *getOrganizationsIterator(): AsyncGenerator<Organization> {
-    if (this.authType === 'token') {
-      const iter = this.octokit.paginate.iterator(
-        this.octokit.orgs.listForAuthenticatedUser,
+    if (this.config.authentication.type === 'token') {
```
chalenge, line 93: really not loving this, not too keen on us having to constantly check the config. Think we can do some OOP for `apps` and one for `orgs` and have these methods specific to that and avoid having if statements?

sources/github-source/src/github.ts
```diff
@@ -25,32 +26,57 @@ export const DEFAULT_GRAPH = 'default';
...
-    if (this.authType === 'token') {
-      await this.octokit.users.getAuthenticated();
-    } else if (this.authType === 'app') {
-      await this.octokit.apps.getAuthenticated();
+    if (this.config.authentication.type === 'token') {
+      await this.authOctokit.users.getAuthenticated();
+    } else if (this.config.authentication.type === 'app') {
+      await this.authOctokit.apps.getAuthenticated();
+    } else {
+      throw new VError('Invalid authentication');
```
chalenge, line 79: `Unsupported authentication method: `${xxx}`

sources/github-source/src/github.ts
```diff
@@ -25,32 +26,57 @@ export const DEFAULT_GRAPH = 'default';
...
+    return instance;
+  }
 
-    GitHub.github = new GitHub(octokit, cfg.authentication.auth, logger);
-    await GitHub.github.checkConnection();
-    return GitHub.github;
+  private async octokit(org?: string): Promise<Octokit> {
+    if (this.config.authentication.type !== 'app' || !org)
+      return this.authOctokit;
+    if (!this.octokitByInstallationOrg.has(org)) {
```
chalenge, line 55: given we already know the org list, feel like we should generate these Octokits when we instantiate `GitHub` so we don't have to do create [everytime we need it like](https://github.com/faros-ai/airbyte-connectors/pull/1543/files#diff-c3ec475eeb0ddb699ccf89b07eb40f24741bc8974d505f77035e1a8d27f4e563R129), not sure what other places we gonna need that too

sources/github-source/src/github.ts
```diff
@@ -25,32 +26,57 @@ export const DEFAULT_GRAPH = 'default';
...
+      GitHub._instance = instance;
+    }
+    return instance;
+  }
 
-    GitHub.github = new GitHub(octokit, cfg.authentication.auth, logger);
-    await GitHub.github.checkConnection();
-    return GitHub.github;
+  private async octokit(org?: string): Promise<Octokit> {
+    if (this.config.authentication.type !== 'app' || !org)
```
chalenge, line 53: shouldn't this just be `if (this.config.authentication.type === 'token')`

sources/github-source/resources/spec.json
dbruno21: For the copilot seats logic to set inactive to true when the API doesn't return anymore the seat
https://github.com/faros-ai/airbyte-connectors/blob/a2f407c96a6b96edc336177e25214dee1d991640/sources/github-source/src/github.ts#L139-L165

Would it be doable in the converter instead?

dbruno21: I talked to @ypc-faros about this and will try to move to destination on another PR

sources/github-source/src/github.ts
dbruno21: removed this piece of code (handled by oop)

sources/github-source/src/github.ts
dbruno21: updated to generate Octokits on GitHub instantiation

sources/github-source/src/github.ts
dbruno21: removed optional from org param

sources/github-source/src/github.ts
dbruno21: PTAL

sources/github-source/src/github.ts
```diff
@@ -25,85 +26,53 @@ export const DEFAULT_GRAPH = 'default';
...
-      await this.octokit.apps.getAuthenticated();
-    }
-  }
+  abstract getAccessibleOrganizationsIterator(): AsyncGenerator<string>;
 
   @Memoize()
-  async getOrganizations(): Promise<ReadonlyArray<Organization>> {
-    const orgs: Organization[] = [];
-    for await (const org of this.getOrganizationsIterator()) {
+  async getAccessibleOrganizations(): Promise<ReadonlyArray<string>> {
```
chalenge: let's keep as `getOrganizations`

sources/github-source/src/github.ts
```diff
@@ -194,3 +163,97 @@ export class GitHub {
     }
   }
 }
+
+export class TokenGitHub extends GitHub {
```
chalenge, line 167: maybe better names are GitHubTokenClient and GitHubAppClient. 
also nit: no need to export these

sources/github-source/src/github.ts
```diff
@@ -25,85 +26,53 @@ export const DEFAULT_GRAPH = 'default';
...
+  }
 
-    const octokit = makeOctokitClient(cfg, logger);
+  abstract checkConnection(): Promise<void>;
 
-    GitHub.github = new GitHub(octokit, cfg.authentication.auth, logger);
-    await GitHub.github.checkConnection();
-    return GitHub.github;
-  }
+  abstract octokit(org: string): Octokit;
```
chalenge, line 55: this can be `octokit(org?: string)`

sources/github-source/src/github.ts
```diff
@@ -194,3 +163,97 @@ export class GitHub {
...
+    cfg: GitHubConfig,
+    logger: AirbyteLogger
+  ): Promise<GitHub> {
+    const baseOctokit = makeOctokitClient(cfg, undefined, logger);
+    const github = new TokenGitHub(cfg, baseOctokit, logger);
+    await github.checkConnection();
+    return github;
+  }
+
+  octokit(org: string): Octokit {
```
chalenge, line 178: `octokit()` - our linters suck. It should be complaining

chalenge (approved):
Few things to address, but lgtm

sources/github-source/src/github.ts
dbruno21: I thought about it and think it's better to force org to be passed for developers to be conscious about it. I think there is no case where we would like to access resources outside of an installation, but if that happens it's better to access baseOctokit directly

sources/github-source/src/github.ts
dbruno21: Had to export for mocking stuff in tests

PR #1545, opened 2024-06-25
FAI-11594 | Copilot Usage stream (source + converter)

dbruno21:
## Description

Copilot Usage stream (source + converter)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


faros-airbyte-common/src/github/types.ts
```diff
@@ -15,3 +15,31 @@ export interface CopilotSeat {
 export enum GitHubTool {
   Copilot = 'GitHubCopilot',
 }
+
+export interface CopilotUsageSummary {
```
chalenge, line 19: can't we use the octokit model or extend it with the few properties we need?

faros-airbyte-common/src/github/types.ts
```diff
@@ -15,3 +15,31 @@ export interface CopilotSeat {
...
+  total_lines_accepted?: number;
+  total_lines_discarded?: number;
+  total_active_users?: number;
+  total_chat_acceptances?: number;
+  total_chat_turns?: number;
+  total_active_chat_users?: number;
+  breakdown: LanguageEditorBreakdown[] | null;
+}
+
+export interface LanguageEditorBreakdown {
```
chalenge, line 35: same as above

chalenge (commented):
Mostly looks fine, see comments about models

ypc-faros:
Hmmm we need to deal with an interesting case here.
Since the GH api for this returns 28 days worth of data (from now), this stream cannot be made incremental. However we also don't want it to make it full refresh, otherwise we'd always wipe historical data older than 28 days.

I _think_ we could work around it by just declaring that we support incremental:

```
get supportsIncremental(): boolean {
    return true;
  }
```

If I'm correct, the airbyte local cli will run this stream in incremental/append and we'd need to make sure we also put incremental in the supported sync modes in the (airbyte connectors configs) catalog.

faros-airbyte-common/src/github/types.ts
dbruno21: Octokit types aren't as happy as Jira ones
We can install on faros-airbyte-common @octokit/rest package and make types like this:
`export type CopilotUsageSummary = {org: string} & RestEndpointMethodTypes["copilot"]["usageMetricsForOrg"]["response"]["data"][0];`
Is that ok?

chalenge: Yikes, had never looked at them. Maybe using [@octokit/types](https://www.npmjs.com/package/@octokit/types) would be slightly better but also looks ugly

dbruno21: That looks a little better. Updated.

faros-airbyte-common/src/github/types.ts
dbruno21: updated

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_usage.ts
```diff
@@ -230,7 +227,7 @@ export class FarosCopilotUsage extends GitHubConverter {
 
   private calculateDiscards(summary: CopilotUsageSummary): void {
     const calculateDifference = (
-      object: CopilotUsageSummary | LanguageEditorBreakdown,
+      object: CopilotUsageSummary | CopilotUsageSummary['breakdown'][0],
```
ypc-faros, line 230: Can we hide that thing 🙈 in the types file and actually call it `LanguageEditorBreakdown`?

ypc-faros (approved):

PR #1546, opened 2024-06-25
FAI-12022: Jira Projects and Boards inclusion and exclusion list

matiaslcoulougian:
## Description
As default behaviour is pull all project/boards, we are supporting exclusión lists so it's possible for the client to exclude specific projects or boards.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/resources/spec.json
```diff
@@ -101,44 +101,66 @@
         "title": "Use Users Prefix Search",
         "default": false
       },
-      "projects": {
+      "projects_included": {
```
ypc-faros, line 104: I'd prefer we keep this as `projects`

sources/jira-source/resources/spec.json
```diff
@@ -101,44 +101,66 @@
...
         },
         "examples": [
-          "PROJ1",
-          "PROJ2"
+          "AIR",
+          "DEV"
         ]
       },
-      "cutoff_days": {
+      "projects_excluded": {
```
ypc-faros, line 117: Maybe `exclude_projects` ?

sources/jira-source/resources/spec.json
```diff
@@ -101,44 +101,66 @@
...
         "order": 15,
+        "title": "Project Keys Excluded",
+        "description": "List of Jira project keys from which data wont be fetched. By default, no projects are excluded.",
```
ypc-faros, line 120: Let's also mention that this list is ignored if `projects` is specified

sources/jira-source/resources/spec.json
```diff
@@ -101,44 +101,66 @@
...
         "type": "integer",
         "title": "Cutoff Lag Days",
         "default": 0,
         "description": "Apply lag to the end cutoff saved in the state. Objects updated after this will be rewritten during the next sync."
       },
-      "boards": {
-        "order": 17,
-        "title": "Board IDs",
-        "description": "List of Jira board IDs from which to fetch data.",
+      "boards_included": {
```
ypc-faros, line 144: Same suggestion as above

sources/jira-source/resources/spec.json
```diff
@@ -101,44 +101,66 @@
...
+        "order": 18,
+        "title": "Board IDs Included",
+        "description": "List of Jira board IDs from which to fetch data. By default, all boards are included.",
+        "type": "array",
+        "items": {
+          "type": "string"
+        }
+      },
+      "boards_excluded": {
```
ypc-faros, line 153: Same as above

sources/jira-source/test/resources/issue_pull_requests/config.json
```diff
@@ -1,8 +1,8 @@
 {
-  "url": "https://jira.com",
+  "url": "https://attlassian.jira.com",
```
chalenge, line 2: why the change. Also typo `atlassian`

sources/jira-source/src/jira.ts
```diff
@@ -54,10 +54,12 @@ export interface JiraConfig extends AirbyteConfig {
   readonly page_size?: number;
   readonly timeout?: number;
   readonly use_users_prefix_search?: boolean;
-  readonly projects?: ReadonlyArray<string>;
+  readonly projects_included?: ReadonlyArray<string>;
```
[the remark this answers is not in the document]
chalenge: What @ypc-faros said. Keep as projects

sources/jira-source/src/streams/faros_projects.ts
```diff
@@ -17,11 +17,18 @@ export class FarosProjects extends StreamBase {
...
-      ? new Set(this.config.projects.map((key) => toUpper(key)))
+    const projectKeys = this.config.projects_included?.length
+      ? new Set(this.config.projects_included.map((key) => toUpper(key)))
       : undefined;
     for (const project of await jira.getProjects(projectKeys)) {
-      yield project;
+      // Include queried project if projectKeys are specified or
+      // if either there is no exclusion list or the project is not in the exclusion list
+      if (
+        projectKeys ||
```
ypc-faros, line 27: In here do we check for `projectKeys` to determine whether `projects_excluded` should be ignored?

If so, I don't like that we need to keep track of this convention everywhere.

I think we have two options:
1) We rewrite the config in the `onBeforeRead` method (if both the inclusion and exclusion lists are specified, we remove the exclusion list)
or
2) We throw in the connection check

I think that will make the code much more readable once you can code under the assumption that only one of these is populated.

sources/jira-source/src/project-board-filter.ts
```diff
@@ -1,6 +1,6 @@
 import {AirbyteLogger} from 'faros-airbyte-cdk';
 import {FarosClient} from 'faros-js-client';
-import {toUpper} from 'lodash';
+import {toString, toUpper} from 'lodash';
```
chalenge, line 3: is this being used?

chalenge (approved):

sources/jira-source/src/project-board-filter.ts
matiaslcoulougian: Yes, here: https://github.com/faros-ai/airbyte-connectors/blob/3af9a497b3fbb1cadb522badd8e8711074b9b82b/sources/jira-source/src/project-board-filter.ts#L80

PR #1552, opened 2024-06-28
FAI-11592 | GitHub users stream (source + converter)

dbruno21:
## Description

GitHub users stream (source + converter)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Extra info


destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -214,6 +215,65 @@ export abstract class GitHubConverter extends Converter {
   id(record: AirbyteRecord): any {
     return record?.record?.data?.id;
   }
+
+  protected convertUser(user: User): DestinationRecord[] {
```
ypc-faros, line 219: Why put this here and not in the converter itself? I'd move them unless you plan on reusing them somewhere else...

sources/github-source/src/types.ts
```diff
@@ -21,3 +21,43 @@ type GitHubApp = {
 };
 
 export type GitHubAuth = GitHubToken | GitHubApp;
+
+// Copied from @octokit/graphql/dist-types/types.d.ts
```
ypc-faros, line 25: Yikes, they don't export these types? :(

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -214,6 +215,65 @@ export abstract class GitHubConverter extends Converter {
...
+    ) {
+      res.push({
+        model: 'vcs_UserEmail',
+        record: {
+          user: {uid: user.login, source: this.streamName.source},
+          email: user.email,
+        },
+      });
+    }
+    const type = ((): {category: string; detail: string} => {
```
ypc-faros: You might be able to use [toCategoryDetail](https://github.com/faros-ai/faros-js-client/blob/17e550e9a8cad6c97d11dabbf44b06810c629fcd/src/utils.ts#L121)

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -214,6 +215,47 @@ export abstract class GitHubConverter extends Converter {
...
+        },
+      });
+    }
+    const type = GitHubCommon.vcs_UserType(user);
+    res.push({
+      model: 'vcs_User',
+      record: {
+        uid: user.login,
+        source: this.streamName.source,
+        ...(user.name && {name: user.name}),
```
ypc-faros, line 239: nit: Would `name: user.name ?? null` work?

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: https://github.com/faros-ai/feeds/blob/2359a0c6191f8293ad7fc3f032212a38e3f1e3b6/feeds/vcs/github-feed/src/feed.ts#L278
On the feeds, users records are written from multiple data sources/endpoints, I was already thinking we would have to reuse it on other streams.

sources/github-source/src/types.ts
dbruno21: No, what we can do is adding a codegen script to build those based on github schema + our queries. [example](https://benlimmer.com/2020/05/16/adding-typescript-types-github-graphql-api/)
But for now, I just copied the types from feed since they are already manually built there

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: Just updated to reuse the same fn already written for the community converter

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: Based on how it's done on [feeds](https://github.com/faros-ai/feeds/blob/2359a0c6191f8293ad7fc3f032212a38e3f1e3b6/feeds/vcs/github-feed/src/feed.ts#L297-L304) I thought it won't.
The way it's done in feeds that user data is collected from multiple places (e.g. pull requests, commits, org members) in a set to dedupe based on id/login and filling any null properties that may be present depending on the place where it was extracted. So we don't want to patch values that may be present with null, right?

destinations/airbyte-faros-destination/src/converters/github/common.ts
chalenge: Let's not follow the feeds pattern and do the right thing for streams. The source should be agnostic of how we want to write these data to Faros

sources/github-source/src/streams/faros_users.ts
```diff
@@ -0,0 +1,30 @@
...
+export class FarosUsers extends StreamWithOrgSlices {
+  getJsonSchema(): Dictionary<any, string> {
+    return require('../../resources/schemas/farosUsers.json');
+  }
+
+  get primaryKey(): StreamKey {
+    return ['login'];
+  }
+
+  get supportsIncremental(): boolean {
```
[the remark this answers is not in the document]
dbruno21: vcs_User is an incremental model in feeds (I think it is because we don't want to delete records from people that are not a member anymore or for example people that have made commits as well.

Removing this isn't going to delete old records?

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: What is the right way on streams to write a user e.g. for the author of a commit?
Or do we already know there won't be data parity between feeds and this source?

dbruno21: For example, [this is for commits](https://github.com/faros-ai/feeds/blob/2359a0c6191f8293ad7fc3f032212a38e3f1e3b6/feeds/vcs/github-feed/src/feed.ts#L571-L578).
When we jump into the commits stream, should we just emit uid + source for the user as reference and not include vcs_User on the destinationModels?

chalenge: Oh, this is the destination. Thought it was the source. Maybe fine. Ideally parity and/or better

destinations/airbyte-faros-destination/src/converters/github/common.ts
chalenge: Have you checked the community source how they handle authors/committers in PR and commits? Pushing that whole user object also will make the objects massive but not critical

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -214,6 +215,47 @@ export abstract class GitHubConverter extends Converter {
...
+        ...(user.email && {email: user.email}),
```
chalenge, line 240: maybe use lodash `omitBy(.. isNil)` to make it DRYer

sources/github-source/src/types.ts
dbruno21: Oh thought, you mean for the query responses.
For this specific one, we don't make use of the graphql package as we already make use the .graphql method from the core octokit instance.

sources/github-source/src/github.ts
```diff
@@ -1,26 +1,34 @@
...
+import {ExtendedOctokit, makeOctokitClient} from './octokit';
+import {GitHubConfig, GraphQLErrorResponse, OrgMembers} from './types';
+
+const ORG_MEMBERS_QUERY = fs.readFileSync(
+  path.join(__dirname, '..', 'resources', 'queries', 'list-members-query.gql'),
+  'utf8'
+);
 
 export const PAGE_SIZE = 100;
+const PROMISE_TIMEOUT_MS = 120_000;
```
chalenge, line 24: Can we document where this number comes from or came up with it?

sources/github-source/resources/queries/list-members-query.gql
```diff
@@ -0,0 +1,17 @@
+query listMembers($login: String!, $cursor: String, $page_size: Int = 100) {
```
chalenge, line 1: in next PR, can we add the test that reads these graphQL queries and tests it against different github versions. We have it in the old feed

sources/github-source/src/github.ts
```diff
@@ -66,6 +74,30 @@ export abstract class GitHub {
     ]);
   }
 
+  async *getOrganizationMembers(org: string): AsyncGenerator<User> {
+    const iter = this.octokit(org).graphql.paginate.iterator<OrgMembers>(
+      ORG_MEMBERS_QUERY,
+      {
+        login: org,
+        page_size: PAGE_SIZE,
```
chalenge, line 82: should we make this configurable

sources/github-source/src/types.ts
```diff
@@ -21,3 +21,43 @@ type GitHubApp = {
...
+export type OrgMembers = {
+  organization: {
+    membersWithRole: {
+      nodes: User[];
+      pageInfo: PageInfo;
+    };
+  };
+};
+
+export type User = {
```
chalenge, line 52: use the one from `airbyte-common`

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -214,6 +215,47 @@ export abstract class GitHubConverter extends Converter {
...
+        ...(user.html_url && {htmlUrl: user.html_url}),
+        type,
+      },
+    });
+    return res;
+  }
+
+  protected convertMembership(user: User, org: string): DestinationRecord[] {
```
chalenge, line 248: if we gonna re-use this for PRs etc we may need to make sure we don't associate users with orgs they do not belong to, e.g. a reviewer or author on an open-source project, so maybe make org optional

destinations/airbyte-faros-destination/src/converters/github/faros_users.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  async convert(
+    record: AirbyteRecord,
+    ctx: StreamContext
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    const res: DestinationRecord[] = [];
+    const user = record.record.data as User;
+    res.push(...this.convertUser(user));
+    res.push(...this.convertMembership(user, user.org));
+    return res;
```
chalenge, line 22:
```typescript
return [this.convertUser(user), this.convertMembership(user, user.org)];
```

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: > Have you checked the community source how they handle authors/committers in PR and commits? Pushing that whole user object also will make the objects massive but not critical

For PRs the converter is writing vcs_User for the author as expected.
https://github.com/faros-ai/airbyte-connectors/blob/7f73299e6e53a405f8cc89cf27bbcd922235e3f1/destinations/airbyte-faros-destination/src/converters/github/pull_requests.ts#L36-L42

As for commits, it seems it isn't handled
https://github.com/faros-ai/airbyte-connectors/blob/8e594ace3f7b6193850c95b3a79a6d81ff48f84a/destinations/airbyte-faros-destination/src/converters/github/commits.ts#L8-L11

sources/github-source/src/github.ts
dbruno21: It's all coming from feeds (both the value and the helper function). There is no context/docs there on why all graphql calls are wrapped with a timeout limit, but I thought it may be there for a reason. 
If you know why I can add it.

sources/github-source/src/github.ts
dbruno21: On feeds it was fixed, can add as improvement.
It's the same value for rest / graphql resources

sources/github-source/src/types.ts
dbruno21: This one is meant to be the typing for the graphql result member nodes. The one on airbyte-common is for the emitted record. Might actually rename this one.

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: Yea, makes sense to change if we end up doing so.

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: > Have you checked the community source how they handle authors/committers in PR and commits? Pushing that whole user object also will make the objects massive but not critical

For PRs the converter is writing vcs_User for the author as expected.
https://github.com/faros-ai/airbyte-connectors/blob/7f73299e6e53a405f8cc89cf27bbcd922235e3f1/destinations/airbyte-faros-destination/src/converters/github/pull_requests.ts#L36-L42

As for commits, it seems it isn't handled
https://github.com/faros-ai/airbyte-connectors/blob/8e594ace3f7b6193850c95b3a79a6d81ff48f84a/destinations/airbyte-faros-destination/src/converters/github/commits.ts#L8-L11

sources/github-source/resources/queries/list-members-query.gql
dbruno21: added on this one

sources/github-source/src/types.ts
chalenge: It's fine it will make this look better and more readable. Looks ugly right now

chalenge (approved):

PR #1553, opened 2024-06-28
FAI-12095 - Signal an error if there is evidence that the Source crashed unexpectedly

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/destination.ts
+++ b/destinations/airbyte-faros-destination/src/destination.ts
@@ -923,6 +923,21 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
           .every((m) => m === DestinationSyncMode.OVERWRITE);
 
       if (
+        sourceConfigReceived &&
+        !sourceSucceeded &&
+        !syncErrors.fatal.length &&
+        !syncErrors.nonFatal.length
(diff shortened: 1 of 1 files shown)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -922,6 +922,21 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
           .map((s) => s.destination_sync_mode)
           .every((m) => m === DestinationSyncMode.OVERWRITE);
 
+      if (
+        sourceConfigReceived &&
+        !sourceSucceeded &&
```
tovbinm, line 927: what if `sourceSucceeded = false` and no `syncErrors` are present?

cjwooo: That's the exact case this conditional is checking -- destination didn't get any success signal or any error signal from a Faros-managed source. It knows that the source is a Faros one due to it having gotten the source configuration message.

ypc-faros (approved):

PR #1562, opened 2024-07-02
FAI-11601 | Added run_mode param for GitHub source

dbruno21:
## Description

- Added run_mode param for GitHub source
  - CopilotEvaluationApp mode streams already finished

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/resources/spec.json
+++ b/sources/github-source/resources/spec.json
@@ -75,7 +75,7 @@
         "description": "GitHub API URL, e.g for GitHub Enterprise: https://my-ghe-host.com/api/v3",
         "default": "https://api.github.com"
       },
-      "orgs": {
+      "organizations": {
         "order": 2,
         "title": "Organizations list",
         "description": "List of GitHub organization from which to fetch data. If none passed, all visible organizations for the authenticated user/app will be used.",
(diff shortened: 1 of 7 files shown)


ypc-faros (approved):

sources/github-source/src/streams/common.ts
```diff
@@ -8,6 +8,48 @@ export type OrgStreamSlice = {
   org: string;
 };
 
+export enum RunMode {
+  MarketplaceApp = 'MarketplaceApp',
```
chalenge, line 12: this should be FarosMarketplaceApp

sources/github-source/resources/spec.json
```diff
@@ -117,6 +117,19 @@
...
         }
+      },
+      "run_mode": {
+        "order": 6,
+        "title": "Run Mode",
+        "description": "Run mode for selecting the usage of the connector",
+        "type": "string",
+        "default": "Standard",
+        "enum": [
+          "MarketplaceApp",
```
chalenge, line 128: FarosMarketplaceApp. Well actually what is the name of the app in the marketplace? We should use that

chalenge (commented):
Let's use correct name for `MarketplaceApp`

sources/github-source/resources/spec.json
dbruno21: updated

sources/github-source/src/streams/common.ts
dbruno21: updated

PR #1563, opened 2024-07-03
Make Jira board_issues and project_version_issues syncs more performant

chalenge:
## Description

- Make Jira board_issues and project_version_issues syncs more performant
- Instead of getting all issues in a board or for a project version, only get those since when we started syncing since everything else will be a phantom
- Fixes for Jira retry logic

## Type of change
- [X] Bug fix
- [x] New feature
- [ ] Breaking change


matiaslcoulougian (approved):
LGTM🔥

matiaslcoulougian:
Solve sonnarcloud issues

PR #1568, opened 2024-07-05
FAI-11596 | Pull requests stream (source + converter) - only full sync

dbruno21:
## Description

- Pull requests stream (source + converter) - only full sync (incremental to be done on another PR)
- Autogenerate graphql query response types and use them as base for common types (moved query/schemas to common)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
```diff
@@ -0,0 +1,90 @@
...
+          title: pr.title,
+          description: Utils.cleanAndTruncate(pr.body),
+          state,
+          htmlUrl: pr.url,
+          createdAt: Utils.toDate(pr.createdAt),
+          updatedAt: Utils.toDate(pr.updatedAt),
+          mergedAt: Utils.toDate(pr.mergedAt),
+          readyForReviewAt,
+          commitCount: pr.commits.totalCount,
+          commentCount: pr.comments.totalCount, // + reviewCommentCount ??? https://github.com/faros-ai/feeds/blob/2359a0c6191f8293ad7fc3f032212a38e3f1e3b6/feeds/vcs/github-feed/src/feed.ts#L1169
```
dbruno21, line 63: https://github.com/faros-ai/feeds/blob/2359a0c6191f8293ad7fc3f032212a38e3f1e3b6/feeds/vcs/github-feed/src/feed.ts#L1169

@chalenge @ypc-faros
The complexity here is that reviewCommentCount comes from the nodes fetched on PR reviews, which would be fetched on PR Reviews stream instead. How should we handle this here?

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
```diff
@@ -0,0 +1,90 @@
...
+            linesDeleted: pr.deletions,
+            filesChanged: pr.changedFiles,
+          },
+          author: pr.author
+            ? {uid: pr.author.login, source: this.streamName.source}
+            : null,
+          mergeCommit: pr.mergeCommit
+            ? {repository: pr.repo, sha: pr.mergeCommit.oid}
+            : null,
+          // ...branchInfo, https://github.com/faros-ai/feeds/commit/865536ae5bae6f9c1edb6a4ea826d91caf9a6136#diff-62dd1a5f8c5bafec2c89e9f598c58ee32673e3d2b8a6ef60a75fc612be45bc72
```
dbruno21, line 75: https://github.com/faros-ai/feeds/commit/865536ae5bae6f9c1edb6a4ea826d91caf9a6136#diff-62dd1a5f8c5bafec2c89e9f598c58ee32673e3d2b8a6ef60a75fc612be45bc72

@chalenge 
Think you missed vcs_Branch related stuff on the spike.
1. Should we write vcs_Branch and add to destinationModels on this stream (Pull Requests)?
2. Also saw the logic to write vcs_Repository and vcs_Organization to prevent phantoms? Should this stream also emit them to follow that logic?

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
```diff
@@ -0,0 +1,90 @@
...
+  ];
+
+  async convert(
+    record: AirbyteRecord,
+    ctx: StreamContext
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    const pr = record.record.data as PullRequest;
+
+    if (pr.author?.login && !this.collectedAuthors.has(pr.author.login)) {
+      this.collectedAuthors.set(pr.author.login, pr.author);
```
dbruno21, line 23: I'm hoping that we won't get pr.author records where for the same login handle we have different properties being filled, if not we need to do the logic to merge with non-null ones.

chalenge (commented):
Looks fine, question on the generated stuff?

faros-airbyte-common/src/github/generated/index.ts
```diff
@@ -0,0 +1,7954 @@
+export type Maybe<T> = T | null;
```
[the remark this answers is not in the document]
dbruno21: This are the auto-generated file for types based on queries (run with `npm run codegen:github`).
There are lot of types generated from the schema which aren't actually used, but unfortunatelly there is no setting to leave them out. https://stackoverflow.com/a/74531662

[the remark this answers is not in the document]
chalenge: Wondering if this is really worth the trouble. Seems like a lot of useless stuff. Okay with unknown types if this makes it easier to use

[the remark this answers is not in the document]
dbruno21: I think the query response types are really useful. I thought at first it wasn't when making the graphql query for users stream, but for the PRs one which is a big one, it really is.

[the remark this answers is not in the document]
dbruno21: Useless stuff doesn't make it any harder to use, it's just there.

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
dbruno21: So to be clear:
1. Emit vcs_Branch. Add to destination models as well?
2. Don't emit vcs_Repository and vcs_Organization.

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
dbruno21: updated to handle easily (works for other streams as well)

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
dbruno21: updated (added vcs_Branch to destination models for now)

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
```diff
@@ -0,0 +1,148 @@
+import {AirbyteRecord} from 'faros-airbyte-cdk';
+import {PullRequest} from 'faros-airbyte-common/github';
+import {Utils} from 'faros-js-client';
+import {camelCase, isNil, last, omitBy, toLower, upperFirst} from 'lodash';
+
+import {DestinationModel, DestinationRecord, StreamContext} from '../converter';
+import {GitHubConverter} from './common';
+
+type RepoKey = {
```
chalenge, line 9: also seeing the same duplicated code in the community converter. Can we share the code in the next PR

chalenge (approved):
Let's split PRs like this into smaller ones in the future.

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
dbruno21: I'm seeing more differences than things in common. It made sense for organizations stream as the conversion was exactly the same. Even if it looks similar, subtle differences on how each field is converted from the logic we are migrating from feeds to the one that's on the community converter, needs to be reviewed one by one so that we don't mess it up.

The fact that the stream records data structure is different makes it harder and involves refactoring to making it generic for both. Is it really worth?

PR #1569, opened 2024-07-05
FAI-11601 | Added full run mode (GitHub)

dbruno21:
## Description

Added full run mode (GitHub)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/src/streams/common.ts
+++ b/sources/github-source/src/streams/common.ts
@@ -13,6 +13,7 @@ export enum RunMode {
   CopilotEvaluation = 'CopilotEvaluation',
   Minimum = 'Minimum',
   Standard = 'Standard',
+  Full = 'Full',
 }
 
 export const CopilotEvaluationAppStreamNames = [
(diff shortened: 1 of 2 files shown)


sources/github-source/src/streams/common.ts
```diff
@@ -48,6 +64,7 @@ export const RunModeStreams = {
   [RunMode.CopilotEvaluation]: CopilotEvaluationStreamNames,
   [RunMode.Minimum]: MinimumStreamNames,
   [RunMode.Standard]: StandardStreamNames,
+  [RunMode.Full]: FullStreamNames,
```
chalenge, line 67: what will be the difference between Full and Standard?

dbruno21: I just saw on the spike docs, that Standard doesn't include ALL.
Full must include all except toggle ones.

ypc-faros: That `Standard` name feels kind of vague to me. Although we're going to have to document the streams included for each mode anyway, so not a big deal I guess.

ypc-faros (approved):

sources/github-source/src/streams/common.ts
chalenge: Fine for now. Need to think more

chalenge (approved):

PR #1578, opened 2024-07-10
FAI-11596 | Incremental sync support for pull requests stream

dbruno21:
## Description

Incremental sync support for pull requests stream + cutoff_days param

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/src/github.ts
+++ b/sources/github-source/src/github.ts
@@ -18,6 +18,7 @@ import {
   ORG_MEMBERS_QUERY,
   PULL_REQUESTS_QUERY,
 } from 'faros-airbyte-common/github/queries';
+import {Utils} from 'faros-js-client';
 import {isEmpty, isNil, pick} from 'lodash';
 import {Memoize} from 'typescript-memoize';
 import VError from 'verror';
@@ -27,6 +28,7 @@ import {GitHubConfig, GraphQLErrorResponse} from './types';
 
 export const PAGE_SIZE = 100;
 const PROMISE_TIMEOUT_MS = 120_000;
+export const DEFAULT_CUTOFF_DAYS = 90;
 
 export abstract class GitHub {
   private static github: GitHub;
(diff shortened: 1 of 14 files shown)


sources/github-source/src/streams/common.ts
```diff
@@ -80,11 +87,55 @@ export abstract class StreamBase extends AirbyteStreamBase {
...
+      currentStreamState,
+      orgRepoKey
+    );
+  }
+
+  protected getUpdateStartDate(cutoff?: number): Date | undefined {
+    return cutoff ? Utils.toDate(cutoff) : this.config.startDate;
+  }
+
+  static calculateUpdatedStreamState(
```
chalenge, line 112: this is duplicated from Jira. Can we put in cdk commons?

sources/github-source/src/streams/common.ts
```diff
@@ -80,11 +87,55 @@ export abstract class StreamBase extends AirbyteStreamBase {
...
+      return {
+        ...currentStreamState,
+        [orgRepoKey]: newState,
+      };
+    }
+    return currentStreamState;
+  }
+
+  static orgRepoKey(org: string, repo: string): string {
+    return `${org}/${repo}`;
```
chalenge, line 138: add `toLower`

chalenge (approved):

sources/github-source/src/streams/common.ts
ypc-faros: +1 (Ok to do in another PR)

sources/github-source/src/github.ts
```diff
@@ -129,14 +130,16 @@ export abstract class GitHub {
     );
     for await (const res of this.wrapIterable(iter, this.timeout)) {
       for (const pr of res.repository.pullRequests.nodes) {
-        pullRequests.push({
+        if (cutoffDate && Utils.toDate(pr.updatedAt) <= cutoffDate) {
```
ypc-faros, line 133: Not needed in this PR but we shouldn't wait too long to adapt all the methods that support incremental streams to process date ranges (start, end).
We'll need it for supporting backfills (see jira source)

sources/github-source/src/streams/common.ts
ypc-faros: 🦅 👀

sources/github-source/src/github.ts
dbruno21: Yes, main reason for not adding it is that GitHub API doesn't support filtering on dates. We can paginate from start and ignore if not included on the range. But should be a separate query so that we only fetch updatedAt + pageInfo.

PR #1581, opened 2024-07-10
FAI-12109 - Manage non-incremental model resets by stream

cjwooo:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -1079,11 +1148,27 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
         // Prepare destination models to delete if any
         if (destinationSyncMode === DestinationSyncMode.OVERWRITE) {
           deleteModelEntries.push(...converter.destinationModels);
+          const streamName = StreamName.fromString(stream).name;
+
+          if (resetModelsByStream[streamName]) {
```
ypc-faros, line 1153: Just curious... I don't understand how this could happen. The destination `write` is run with a single catalog.

Or are we thinking of a weird case where someone has a file with the output for multiple sources...?

ypc-faros (approved):
🔥

destinations/airbyte-faros-destination/src/destination.ts
cjwooo: I agree this shouldn't happen. I put this check here as a just-in-case defensive measure.

PR #1586, opened 2024-07-12
FAI-12100 Enable debug level logging if configured

ypc-faros:


matiaslcoulougian (approved):

package-lock.json
matiaslcoulougian, line 1: Why are updating package-lock file?

ypc-faros: Added unintentionally. Removed

PR #1595, opened 2024-07-18
FAI-12238 | Copilot seats fix startedAt for team assignments

dbruno21:
## Description

Copilot seats fix startedAt for team assignments
Applied "same" logic as https://github.com/faros-ai/feeds/pull/2219, but dealing with source + destination separation.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/github-source/src/github.ts
```diff
@@ -383,16 +395,36 @@ export abstract class GitHub {
     try {
       for await (const res of iter) {
         for (const seat of res.data.seats) {
-          if (!seatsFound) seatsFound = true;
+          if (!seatsFound) {
```
chalenge, line 398: why not just `seatsFound = true` without the if statement?

sources/github-source/src/github.ts
```diff
@@ -383,16 +395,36 @@ export abstract class GitHub {
...
+          }
+          if (seat.assigning_team && !teamAddMemberTimestamps) {
+            // try to fetch team add member timestamps only if there are seats with team assignments
+            teamAddMemberTimestamps = await this.getTeamAddMemberTimestamps(
+              org,
+              'copilot team assignments',
+              cutoffDate
+            );
+          }
+          const userAssignee = seat.assignee.login as string;
```
chalenge, line 409: no need to nullish coalescing if they're already nulls
```typescript
const userAssignee = seat.assignee?.login?.toString()
const teamAssignee = seat.assigning_team?.slug;
```

sources/github-source/src/github.ts
```diff
@@ -383,16 +395,36 @@ export abstract class GitHub {
...
+          const teamAssignee = seat.assigning_team?.slug ?? null;
+          let startedAt = Utils.toDate(seat.created_at);
+          if (teamAssignee) {
```
chalenge, line 412:
```typescript
if (teamAssignee) {
  const joinedAt = teamAddMemberTimestamps?.[teamAssignee]?.[userAssignee];

  if (joinedAt && joinedAt > startedAt) {
    startedAt = joinedAt;
  }
}
```

maybe even just `joinedAt > startedAt`

chalenge (commented):
Mostly fine, see comments. Add to README with perms for CoPilot

sources/github-source/src/github.ts
dbruno21: seat.assignee.login it's explicitly typed as string because of a conflict in the API types ->

Screenshot 2024-07-18 at 8 44 52 PM

sources/github-source/src/github.ts
```diff
@@ -443,6 +473,70 @@ export abstract class GitHub {
...
+        per_page: PAGE_SIZE,
+      }
+    );
+    try {
+      for await (const res of iter) {
+        for (const log of res.data) {
+          logs.push(log);
+        }
+      }
+    } catch (err: any) {
```
chalenge, line 501: Missed this initially but let's follow-up in another PR. Can we check the exact status code it returns instead of `any`. If we have an intermittent issue we might incorrectly use the wrong API.

Also think we can have a better test for when to use this API by checking the [type](https://docs.github.com/en/get-started/learning-about-github/types-of-github-accounts) when we get the [organization](https://docs.github.com/en/enterprise-cloud@latest/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization) if it is `Enterprise`. Let's test that too.

chalenge (approved):
Nice 💥 . See comment for follow-up

sources/github-source/src/github.ts
dbruno21: Agree

PR #1596, opened 2024-07-18
FAI-11603: Add files to Pull Requests stream

matiaslcoulougian:
## Description
Pull PR files conditionally in PRs stream.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/github-source/src/github.ts
```diff
@@ -163,16 +170,71 @@ export abstract class GitHub {
...
+
+  private async getFilesForPullRequest(
+    owner: string,
+    repo: string,
+    number: number,
+    endCursor: string
+  ): Promise<PullRequestFile[]> {
+    const files: PullRequestFile[] = [];
+    const iter = this.octokit(
+      'farosai'
```
dbruno21, line 221:
```suggestion
      owner
```

sources/github-source/src/github.ts
```diff
@@ -163,16 +170,71 @@ export abstract class GitHub {
...
+      const remainingFiles = await this.getFilesForPullRequest(
+        org,
+        repo,
+        pr.number,
+        pr.files.pageInfo.endCursor
+      );
+      files = files.concat(remainingFiles);
+    }
+    return files;
+  }
```
dbruno21, line 211: We should probably do something generic for the possible nested fragments we can fetch. Can do on following PRs as we may see too many duplicated logic.

destinations/airbyte-faros-destination/src/converters/github/faros_pull_requests.ts
```diff
@@ -85,20 +99,36 @@ export class FarosPullRequests extends GitHubConverter {
...
-      ...pr.labels.nodes.map((label) => ({
+      ...pr.labels.map((label) => ({
         model: 'vcs_PullRequestLabel',
         record: {
           pullRequest: prKey,
           label: {name: label.name},
         },
       })),
+      ...pr.files.map((file) => {
+        this.collectFile(file.path, repoKey);
```
dbruno21, line 110: nit: I don't like collect fn being run inside map

sources/github-source/src/github.ts
matiaslcoulougian: Yes, makes sense

dbruno21 (approved):

faros-airbyte-common/resources/github/queries/pull-request-files-query.gql
```diff
@@ -0,0 +1,17 @@
+query pullRequestFiles($owner: String!, $repo: String!, $number: Int!, $cursor: String, $page_size: Int = 100) {
```
chalenge, line 1: not ideal but for APIs we are querying, I think we should use the [REST API](https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files) so we can distribute rate-limits across resources since we have different limits, see [here](https://docs.github.com/en/rest/rate-limit/rate-limit?apiVersion=2022-11-28)

sources/github-source/src/github.ts
```diff
@@ -163,16 +170,71 @@ export abstract class GitHub {
...
       }
     }
   }
 
+  private buildPRQuery(): string {
+    let query = PULL_REQUESTS_QUERY;
+    if (this.fetchFiles) {
+      query += FILES_FRAGMENT;
+    } else {
+      query = query.replace('...Files', '');
```
chalenge, line 190: 🤔

sources/github-source/src/github.ts
chalenge: I was wondering if the default graphql pagination library doesn't support this already?

sources/github-source/src/github.ts
```diff
@@ -524,6 +586,7 @@ export class GitHubToken extends GitHub {
       baseOctokit,
       cfg.bucket_id ?? DEFAULT_BUCKET_ID,
       cfg.bucket_total ?? DEFAULT_BUCKET_TOTAL,
+      cfg.fetch_files ?? true,
```
chalenge, line 589: is `?? true` necessary since the default in the spec is `true`?

sources/github-source/src/github.ts
dbruno21: https://github.com/octokit/plugin-paginate-graphql.js/?tab=readme-ov-file#unsupported-nested-pagination

There's also an interesting video explaining GitHub graphql API and some patterns to follow.

sources/github-source/src/github.ts
dbruno21: > is ?? true necessary since the default in the spec is true?

Same could be said for bucket_id and bucket_total values. I think the spec default isn't used when running via airbyte CLI, so it should be passed here as well.

faros-airbyte-common/package.json
```diff
@@ -31,6 +31,7 @@
     "./github": "./lib/github/index.js",
     "./github/generated": "./lib/github/generated/index.js",
     "./github/queries": "./lib/github/queries.js",
+    "./github/fragments": "./lib/github/fragments.js",
```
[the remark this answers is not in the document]
dbruno21: Originally, I found it cleaner to have split packages ("github" - "github/queries" - "github/generated") instead of having everything mixed up on the same package. But is possible.

sources/github-source/src/github.ts
matiaslcoulougian: Yes, when running via CLI defaults are not set using the spec

faros-airbyte-common/package.json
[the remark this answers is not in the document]
chalenge: Let's merge them

sources/github-source/resources/spec.json
```diff
@@ -199,6 +199,13 @@
       "title": "Total Number of Buckets",
       "description": "Total number of buckets to distribute organizations and repositories across. Use it when distributing the load between multiple sources",
       "default": 1
+    },
+    "fetch_files": {
```
chalenge, line 203: `fetch_pull_request_files`

sources/github-source/src/github.ts
```diff
@@ -163,16 +168,75 @@ export abstract class GitHub {
...
+    org: string,
+    repo: string
+  ): Promise<PullRequestFile[]> {
+    let files = pr.files.nodes;
+    if (pr.files.pageInfo.hasNextPage) {
+      const remainingFiles = await this.getPullRequestFiles(
+        org,
+        repo,
+        pr.number,
+        2 // Start from the second page to avoid re-fetching the first one
```
chalenge, line 204: hmmm, are we sure that the pages return the same things if we are using different APIs?

chalenge (commented):
See comment about fetching extra files

sources/github-source/src/github.ts
matiaslcoulougian: I found no documentation that assured this, but performed some tests that proved to me that the pages returned are the same, even though the pagination methods are different between the APIs.
I did a test comparing results returned from gql and from the rest api with files < 100 and results were identically the same with the same order.

matiaslcoulougian: Then tested with a pr with files > 100, using this logic to fetch the remaining files via the rest api listFiles. The collected files were correct, matching effectivly the files that pr has.

dbruno21: The edge case is the PR gets updated on the middle of these requests, but that's a common problem with page number pagination, not because we are combining APIs. Could be solved by using GraphQL cursor based pagination. But we have this tradeoff because of not trying to use graphql that much.

chalenge: Let's start from 1 to be completely sure and have guaranteed results

PR #1599, opened 2024-07-22
Fix bug in Jira Source Project filter

matiaslcoulougian:
## Description
Fix projects list check in project-board-filter. If `projects` is `[]` (empty array) we should sync all projects, which is the same behaviour as projects being `undefined`.

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


chalenge (approved):
🙇🏽

PR #1601, opened 2024-07-23
FAI-12304 | Workaround for copilot seats endedAt when using team assignments

dbruno21:
## Description

Workaround for copilot seats endedAt when using team assignments.
How it works:
- Track every team assignment
- Added to audit logs query filter for team.remove_member in addition to team.add_member
- If found from the audit logs a user that was removed from a team, after the team got copilot assigned, and we verify the user doesn't currently have copilot access, we assume the copilot seat was removed when the user was removed from the team

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


chalenge (commented):
Let's simplify the code for `wasUserRemovedAfter`.

chalenge (approved):
lgtm

PR #1603, opened 2024-07-25
FAI-12391 | Added permissions required for each GitHub stream

dbruno21:
## Description

Added permissions required for each GitHub stream

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/README.md
+++ b/sources/github-source/README.md
@@ -79,9 +79,10 @@ connectors, such as linting/formatting tools, should go in the root
 
 ### GitHub Required Permissions per Stream
 
-*Fine-grained permissions marked as required are always read-only.
+(*) Permissions marked as required are always read-only.\
+(**) Fine-grained access tokens only have access to one organization and must be explicitly passed in the configuration.
 
-| Stream                | Classic              | Fine-grained*                                    |
+| Stream                | Classic              | Fine-grained \| GitHub App (*)                   |
 |-----------------------|----------------------|--------------------------------------------------|
 | Commits               | repo                 | Repository: Metadata & Contents                  |
 | Copilot Seats         | read:org             | Organization: Administration                     |
--- a/sources/github-source/README.md
+++ b/sources/github-source/README.md
@@ -81,16 +81,16 @@ connectors, such as linting/formatting tools, should go in the root
 
 *Fine-grained permissions marked as required are always read-only.
 
-| Stream                | Classic              | Fine-grained*                          |
-|-----------------------|----------------------|----------------------------------------|
-| Commits               | repo                 | Repository: Metadata & Contents        |
-| Copilot Seats         | read:org             | Organization: Administration           |
-| Copilot Usage         | read:org             | Organization: Administration           |
-| Labels                | repo                 | Repository: Metadata & Issues          |
-| Organizations         | read:org             | Organization: Administration           |
-| Pull Requests         | repo                 | Repository: Metadata & Pull Requests   |
-| Pull Request Comments | repo                 | Repository: Metadata & Pull Requests   |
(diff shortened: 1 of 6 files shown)


sources/github-source/src/org-repo-filter.ts
```diff
@@ -100,10 +97,7 @@ export class OrgRepoFilter {
         this.filterConfig.reposByOrg.get(org).forEach((repoName) => {
           const repo = visibleRepos.find((r) => r.name === repoName);
           if (!repo) {
-            this.logger.warn(
-              `Skipping not found repository ${org}/${repoName}`
-            );
-            return;
+            throw new VError(`Repository not found: ${org}/${repoName}`);
```
dbruno21, line 100: I see on feeds there's a setting for this `fail_missing_repos`, should we add (on another PR)? @chalenge

dbruno21: reverted

sources/github-source/src/org-repo-filter.ts
```diff
@@ -63,19 +64,15 @@ export class OrgRepoFilter {
...
       if (!this.filterConfig.organizations) {
+        const visibleOrgs = await github.getOrganizations();
         visibleOrgs.forEach((org) => {
           if (!this.filterConfig.excludedOrganizations?.has(org)) {
             organizations.add(org);
           }
         });
       } else {
         this.filterConfig.organizations.forEach((org) => {
-          if (!visibleOrgs.some((o) => o === org)) {
```
dbruno21, line 75: Fine-grained tokens return an empty list when listing organizations for the authenticated user, so we can't validate against visible orgs and organizations in that case.

dbruno21: Left check agains visible orgs only if visible orgs list is not empty.

sources/github-source/src/org-repo-filter.ts
```diff
@@ -72,7 +73,11 @@ export class OrgRepoFilter {
...
+          // fine-grained tokens return an empty list for visible orgs,
+          // so we only run the check if the list is not empty
+          if (visibleOrgs.length === 0) {
```
chalenge: `visibleOrgs.length` works too.

sources/github-source/src/org-repo-filter.ts
```diff
@@ -64,6 +65,9 @@ export class OrgRepoFilter {
       const organizations = new Set<string>();
       const github = await GitHub.instance(this.config, this.logger);
       const visibleOrgs = await github.getOrganizations();
+      if (visibleOrgs.length === 0) {
```
chalenge, line 68: `if visibleOrgs.length` works fine

sources/github-source/src/org-repo-filter.ts
chalenge: Another qst, with fine-grained token does GET `/organization/{orgId}` work then?

sources/github-source/README.md
```diff
@@ -79,10 +79,18 @@ connectors, such as linting/formatting tools, should go in the root
...
 
-| Stream        | Classic               | Fine-grained*                            |
-|---------------|-----------------------|------------------------------------------|
-| Commits       | repo:status, read:org | Repository Contents, Repository Metadata |
-| Copilot Seats | read:org              | Organization Administration              |
-| Copilot Usage | read:org              | Organization Administration              |
-
-*Fine-grained permissions marked as required are always read-only.
\ No newline at end of file
+*Fine-grained permissions marked as required are always read-only.
```
chalenge, line 82: Can we add the comment about fine-grained permission only applying to one organization

chalenge (approved):
lgtm, after comment

PR #1605, opened 2024-07-25
FAI-11613 | Collect and write users on team memberships stream

dbruno21:
## Description

Collect and write users on team memberships stream

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


matiaslcoulougian (approved):
🚀

sources/github-source/src/streams/faros_team_memberships.ts
```diff
@@ -11,7 +11,7 @@ export class FarosTeamMemberships extends StreamWithOrgSlices {
   }
 
   get primaryKey(): StreamKey {
-    return ['org', 'team', 'user'];
+    return [['org'], ['team'], ['user', 'login']];
```
chalenge, line 14: should this be `return ['org', 'team', ['user, 'login']`. Did we test against a local Airbyte?

dbruno21: Local Airbyte works.
Syntax is also the correct one. Only string | string[] | string[][] are valid

another example https://github.com/faros-ai/airbyte-connectors/blob/44c744a4c8777733a9f616ff50b512f2b437d5b7/sources/bitbucket-server-source/src/streams/pull_requests.ts#L26

PR #1615, opened 2024-07-31
FAI-11608: GitHub Tags Source + Stream Converter

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Migration notes

## Extra info


faros-airbyte-common/src/github/types.ts
```diff
@@ -1,11 +1,13 @@
 import {Octokit} from '@octokit/rest';
 import {GetResponseDataTypeFromEndpointMethod} from '@octokit/types';
+import {exec} from 'child_process';
```
dbruno21, line 3:
```suggestion
```

sources/github-source/src/streams/faros_tags.ts
```diff
@@ -0,0 +1,55 @@
...
+    yield* github.getTags(org, repo, cutoffDate);
+  }
+
+  getUpdatedState(
+    currentStreamState: StreamState,
+    latestRecord: Tag,
+    slice: RepoStreamSlice
+  ): StreamState {
+    const latestRecordCutoff = Utils.toDate(
+      latestRecord?.commit.committedDate ?? 0
```
dbruno21, line 47: Can commit be null / undefined?

sources/github-source/src/github.ts
```diff
@@ -899,6 +903,40 @@ export abstract class GitHub {
...
+    org: string,
+    repo: string,
+    cutoffDate?: Date
+  ): AsyncGenerator<Tag> {
+    const iter = this.octokit(org).graphql.paginate.iterator<RepoTagsQuery>(
+      REPOSITORY_TAGS_QUERY,
+      {
+        owner: org,
+        repo,
+        per_page: this.pageSize,
```
dbruno21, line 916: There's no param to filter by date, right?

sources/github-source/src/github.ts
```diff
@@ -899,6 +903,40 @@ export abstract class GitHub {
...
+      for (const tag of res.repository.refs.nodes) {
+        let commit: TagsQueryCommitNode;
+        if (tag.target.type === 'Commit') {
+          commit = tag.target;
+        }
+        if (tag.target.type === 'Tag' && tag.target.target.type === 'Commit') {
+          commit = tag.target.target;
+        }
+        if (cutoffDate && Utils.toDate(commit.committedDate) <= cutoffDate) {
+          continue;
```
dbruno21, line 929: There's no way / param to sort by date, right? If there is or results are already sorted this should be a break;

dbruno21: How does feeds handle the incremental mode?

sources/github-source/src/streams/faros_tags.ts
matiaslcoulougian: Here it should not be null. But target could be null, will handle that case [here](https://github.com/faros-ai/airbyte-connectors/pull/1615/files#diff-c3ec475eeb0ddb699ccf89b07eb40f24741bc8974d505f77035e1a8d27f4e563R922):

sources/github-source/src/github.ts
matiaslcoulougian: Nope

sources/github-source/src/github.ts
matiaslcoulougian: Fees do not sync this incrementally, all tags are always pulled and written

matiaslcoulougian: Removed incremental sync so we don't miss any tags added to commits committed before cutoffDate.

dbruno21 (approved):

PR #1627, opened 2024-08-06
FAI-12482 | Added "Custom" run mode option including ALL streams

dbruno21:
## Description

Added "Custom" run mode option including ALL streams, so that Full/Standard (default) run mode can be a subset of all possible streams. When Custom is selected the stream selection is done through Airbyte stream settings.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


matiaslcoulougian (approved):
lgtm

PR #1629, opened 2024-08-07
FAI-11610: GitHub Projects Stream

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/github-source/src/streams/faros_projects.ts
```diff
@@ -0,0 +1,61 @@
...
+    syncMode: SyncMode,
+    cursorField?: string[],
+    streamSlice?: OrgStreamSlice,
+    streamState?: StreamState
+  ): AsyncGenerator<Project> {
+    const github = await GitHub.instance(this.config, this.logger);
+    const org = streamSlice?.org;
+    const state = streamState?.[StreamBase.orgKey(org)];
+    const cutoffDate = this.getUpdateStartDate(state?.cutoff);
+    for await (const project of github.getProjects(org, cutoffDate)) {
```
dbruno21, line 38: yield* github.getProjects(org, cutoffDate)
yield* github.getClassicProjects(org, cutoffDate)

I think this works as well

sources/github-source/src/github.ts
```diff
@@ -979,6 +982,60 @@ export abstract class GitHub {
...
+    cutoffDate?: Date
+  ): AsyncGenerator<Project> {
+    const iter = this.octokit(org).paginate.iterator(
+      this.octokit(org).projects.listForOrg,
+      {
+        org,
+        per_page: this.pageSize,
+      }
+    );
+    for await (const res of iter) {
```
dbruno21, line 1025: This will throw an error when running a version with API deprecated (won't happen at least until November). Maybe we should catch for 404's (unknown route), but hard to know if that would be the behavior when deprecated.

matiaslcoulougian: Hmmm, true, will think of how we can defend from that case

sources/github-source/src/streams/faros_projects.ts
matiaslcoulougian: I can try

dbruno21 (approved):

PR #1630, opened 2024-08-08
FAI-11606 | Backfilling support for GH incremental streams

dbruno21:
## Description

- Backfilling support for GH incremental streams
- Left comments on each of the incremental streams describing the strategy depending on what we can do with each API method

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Extra info


sources/github-source/resources/spec.json
```diff
@@ -169,47 +169,73 @@
...
       },
       "concurrency_limit": {
-        "order": 16,
+        "order": 17,
         "type": "integer",
         "title": "Concurrency limit",
         "description": "Maximum concurrency to run with",
         "default": 4
+      },
+      "backfill": {
```
matiaslcoulougian, line 221: Do we need this backfill boolean? Couldn't we just check for start_date and end_date? 

Also, I noticed we have it in Jira source but we are not using it.

chalenge: We need to pass it to [the CDK](https://github.com/faros-ai/airbyte-connectors/blob/b6f02a6cca0df5c544ed663d9b6a0c2d35ea5569/faros-airbyte-cdk/src/protocol.ts#L53) to ensure we do not write state.

sources/github-source/src/github.ts
```diff
@@ -414,21 +462,31 @@ export abstract class GitHub {
   async *getPullRequestComments(
     org: string,
     repo: string,
-    cutoffDate?: Date
+    startDate?: Date,
+    endDate?: Date
   ): AsyncGenerator<PullRequestComment> {
+    // query supports filtering by start date (since) but not by end date
```
chalenge, line 468: not sure this is better. For really large repos that have been around for years this would be really slow. Think we likely still better off starting from the top since we likely gonna backfill newer data

sources/github-source/src/github.ts
```diff
@@ -414,21 +462,31 @@ export abstract class GitHub {
...
+    // for backfill, we iterate in ascending order and stop when we reach the end date
     const iter = this.octokit(org).paginate.iterator(
       this.octokit(org).pulls.listReviewCommentsForRepo,
       {
         owner: org,
         repo: repo,
-        since: cutoffDate?.toISOString(),
-        direction: 'desc',
+        since: startDate?.toISOString(),
+        direction: this.backfill ? 'asc' : 'desc',
```
chalenge, line 476: see comment above

sources/github-source/src/github.ts
```diff
@@ -1057,21 +1126,21 @@ export abstract class GitHub {
...
-      const iter = this.octokit(org).paginate.iterator(
-        this.octokit(org).projects.listForOrg,
-        {
-          org,
-          per_page: this.pageSize,
-        }
-      );
       for await (const res of iter) {
         for (const project of res.data) {
-          if (cutoffDate && Utils.toDate(project.updated_at) <= cutoffDate) {
```
chalenge, line 1072: shouldn't we filter out `closed` projects that haven't been updated in our timeframe?

sources/github-source/resources/spec.json
```diff
@@ -169,47 +169,73 @@
         "description": "Fetch Pull Requests associated reviews",
         "default": true
       },
-      "cutoff_days": {
+      "fetch_projects_classic": {
```
chalenge, line 172: let's just fetch both classic and new ones

chalenge (commented):
Hmmm, really sucks that we have to burn APIs calls to get targeted dates.

sources/github-source/src/github.ts
dbruno21: Not following why would it be slower. The ascending order is just used only if it's a backfill. We start at the desired start date and stop when we start reading non-desired records. If it wasn't ascending order we would have to iterate through pages we don't need until starting reading records < end date.

chalenge: If we have a repo started in 2010 and we only need data from 2022 then we gonna go through all the years from 2010, that will be slow and expensive. It is likely that repos have been around for a while and think we always backfill mostly recent data. We could do some smart heuristics to figure direction but feels like an overkill

sources/github-source/src/streams/faros_projects.ts
```diff
@@ -34,9 +34,14 @@ export class FarosProjects extends StreamWithOrgSlices {
...
     const state = streamState?.[StreamBase.orgKey(org)];
-    const cutoffDate = this.getUpdateStartDate(state?.cutoff);
-    yield* github.getProjects(org, cutoffDate);
-    yield* github.getClassicProjects(org, cutoffDate);
+    const [startDate, endDate] =
+      syncMode === SyncMode.INCREMENTAL
+        ? this.getUpdateRange(state?.cutoff)
+        : this.getUpdateRange();
+    yield* github.getProjects(org, startDate, endDate);
+    if (this.config.fetch_projects_classic ?? DEFAULT_FETCH_PROJECTS_CLASSIC) {
```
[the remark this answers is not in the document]
dbruno21: That makes sense for GH cloud, but for servers running old versions it probably won't be deprecated at the same time (thinking we would end up adding a toggle setting anyway?)

sources/github-source/src/github.ts
dbruno21: Can skip those records yes. (We'll be still pulling everyone since API doesn't support sorting, nor filtering)

sources/github-source/src/github.ts
dbruno21: > query supports filtering by start date (since) but not by end date
That's why we use the "since" parameter. So for that case if start date is on 2022, we will only pull records from 2022 onwards.

chalenge: Ah gotcha, let's add to this `we iterate in ascending order from the start date ...`

sources/github-source/src/streams/faros_projects.ts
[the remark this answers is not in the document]
dbruno21: Also the reason for the default being false is to prevent us having all connectors failing someday and making us having to go 1 by 1 fixing the version / setting.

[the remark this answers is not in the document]
chalenge: Issues are rarely used so not keen on adding more flags. We can handle the errors from that API by default and just log a warning

sources/github-source/src/github.ts
```diff
@@ -1057,20 +1126,30 @@ export abstract class GitHub {
...
-            break;
+          if (
+            this.backfill &&
```
chalenge, line 1145: let's also check the project state

dbruno21: Not sure why's that? If we are sure the project is updated between the timeframe (based on updated_at), why do we care of the state to before writing a record for it?

sources/github-source/resources/spec.json
dbruno21: removed flag

sources/github-source/src/github.ts
chalenge: How do we sync projects that are `open` but hasn't been updated within the timeframe? Do projects always get updated when issues in it are updated?

dbruno21: Issues is a different stream with it's own state.

For projects we are only making use (picking) of these fields: 'name' | 'body' | 'created_at' | 'updated_at'. I'd assume projects get updated if name / body get updated and we'll only sync those. We're not making any distinction if a project is open / closed.

dbruno21: Updated projects stream to be full sync

chalenge (approved):
lgtm

PR #1644, opened 2024-08-20
FAI-12666 | Dynamic page size for PRs query to avoid timeout failures

dbruno21:
## Description

Dynamic page size for PRs query to avoid timeout failures on PR stream. Applied same workaround and values (page size, multipliers) from [feeds implementation](https://github.com/faros-ai/feeds/blob/2f9f1ed6b189252bd3f8fca2c6c9229aff0ce185/feeds/vcs/github-feed/src/feed.ts#L1121)

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/resources/spec.json
+++ b/sources/github-source/resources/spec.json
@@ -197,35 +197,42 @@
         "description": "Maximum number of items in a paginated response",
         "default": 100
       },
-      "timeout": {
+      "max_pull_requests_page_size": {
         "order": 15,
+        "type": "integer",
+        "title": "Maximum Pull Requests Page Size",
+        "description": "Maximum number of items in a paginated response for pull requests",
+        "default": 25
+      },
+      "timeout": {
+        "order": 16,
         "type": "integer",
         "title": "Request Timeout",
         "description": "Timeout in milliseconds for each request to the GitHub API",
         "default": 120000
       },
       "concurrency_limit": {
-        "order": 16,
+        "order": 17,
         "type": "integer",
         "title": "Concurrency limit",
         "description": "Maximum concurrency to run with",
         "default": 4
       },
       "backfill": {
-        "order": 17,
+        "order": 18,
         "type": "boolean",
         "title": "Backfill",
         "description": "Backfill data from the start date to the end date.",
         "default": false
       },
       "start_date": {
-        "order": 18,
+        "order": 19,
         "type": "string",
         "title": "Start Date",
         "description": "The date from which to start syncing data. Only use for backfilling."
       },
       "end_date": {
-        "order": 19,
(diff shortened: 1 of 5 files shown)


sources/github-source/resources/spec.json
```diff
@@ -197,35 +197,42 @@
         "description": "Maximum number of items in a paginated response",
         "default": 100
       },
-      "timeout": {
+      "max_pull_requests_page_size": {
```
chalenge, line 200: `pull_requests_page_size`. Same as the one above 😺

sources/github-source/src/github.ts
```diff
@@ -228,37 +230,71 @@ export abstract class GitHub {
       return;
     }
     const query = this.buildPRQuery();
-    const iter = this.octokit(org).graphql.paginate.iterator<PullRequestsQuery>(
-      query,
-      {
+    let currentPageSize = MAX_PR_PAGE_SIZE;
```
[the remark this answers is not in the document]
dbruno21: Updated

ypc-faros (approved):

sources/github-source/src/octokit.ts
```diff
@@ -36,10 +36,9 @@ export function makeOctokitClient(
   cfg: GitHubConfig,
   installationId?: number,
   logger?: AirbyteLogger,
-  onRateLimit?: RateLimitHandler,
   maxRetries = 3
 ): ExtendedOctokit {
-  const throttle = getThrottle(cfg, logger, onRateLimit, maxRetries);
```
chalenge, line 42: what is changing here?

sources/github-source/src/octokit.ts
```diff
@@ -114,9 +114,12 @@ function getOctokitAuth(
 function getThrottle(
   cfg: GitHubConfig,
   logger: AirbyteLogger,
-  onRateLimit: (reason?: Error) => boolean,
   maxRetries: number
 ): ThrottlingOptions & {global: Bottleneck.Group} {
+  const onRateLimit: RateLimitHandler = (error) => {
```
chalenge, line 119: Will this still do exponential retries?

sources/github-source/src/octokit.ts
dbruno21: Just removing this non-used param

sources/github-source/src/octokit.ts
dbruno21: I noticed from logs the rate limit we always hit is the one from `onSecondaryRateLimit` (that's the only one that's handling retries). I'll update `onRateLimit` to have the same behavior just in case.

PR #1645, opened 2024-08-20
FAI-12633 | Copilot usage metrics by team

dbruno21:
## Description

- Adds copilot usage metrics by team. 
- Metrics are the same ones as written by org, but adding an additional tag with the team name. 
- When fetching list of teams, catch error since it may require additional permissions to fetch teams (same as teams stream)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


ypc-faros (approved):
🚀

PR #1648, opened 2024-08-22
FAI-12259 | Google sheet source support for sheet filters and first row offset

dbruno21:
## Description

Google sheet source support for sheet filters and first row offset (local file and S3 support would need a bigger refactor).

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/sheets-source/src/sheets-reader.ts
+++ b/sources/sheets-source/src/sheets-reader.ts
@@ -36,6 +36,8 @@ type SpreadsheetSource = LocalSpreadsheet | GoogleSheet | S3Spreadsheet;
 
 export interface SheetsConfig {
   readonly spreadsheet_source: SpreadsheetSource;
+  readonly sheet_names?: string[];
+  readonly row_offset?: number;
   readonly sheet_page_size?: number;
   readonly stream_name?: string;
 }
@@ -86,7 +88,9 @@ export class SheetsReader {
 
         wb = await SheetsReader.loadSheets(
           config.spreadsheet_source.google_sheet_id,
+          config.sheet_names,
           config.sheet_page_size,
+          config.row_offset,
           creds,
           logger
         );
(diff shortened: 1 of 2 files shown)


sources/sheets-source/src/sheets-reader.ts
```diff
@@ -197,7 +201,9 @@ export class SheetsReader {
 
   static async loadSheets(
     sheetId: string,
+    sheetNames: string[],
```
ypc-faros, line 204: nit: Maybe use `Set<string>` here to simplify the logic

sources/sheets-source/src/sheets-reader.ts
```diff
@@ -274,11 +284,12 @@ export class SheetsReader {
     doc: GoogleSpreadsheet,
     sheetName: string,
     sheetPageSize: number,
+    rowOffset: number,
     logger?: AirbyteLogger
   ): Promise<XLSX.WorkSheet> {
     const sheet = SheetsReader.getSheetOrError(doc, sheetName);
 
-    await sheet.loadHeaderRow();
+    await sheet.loadHeaderRow(1 + (rowOffset ?? 0));
```
ypc-faros, line 292: nit: Same. Maybe do this in the call instead?

ypc-faros (approved):
A couple of minor comments. I'm okay either way.

sources/sheets-source/src/sheets-reader.ts
dbruno21: The string[] comes directly from the Airbyte spec (was trying to avoid doing too much handling of what comes from the config). Think we could refactor if we extend to the S3 and local file to parse config / validate / assign defaults at the root level for these new configs.

sources/sheets-source/src/sheets-reader.ts
dbruno21: I found it safer here. If it's forgotten on the call, this throws an error. Also, think we could refactor if we extend to the S3 and local file to parse config / validate / assign defaults at the root level  for these new configs.

PR #1656, opened 2024-08-29
FAI-12270: Bitbucket backfill dates

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/bitbucket-source/src/bitbucket.ts
```diff
@@ -146,10 +140,10 @@ export class Bitbucket {
   async *getCommits(
     workspace: string,
     repoSlug: string,
-    lastUpdated?: number
+    startDate?: Date,
```
ypc-faros, line 143: Why are these optional? Aren't we always passing values for start/end?

sources/bitbucket-source/resources/spec.json
```diff
@@ -87,6 +87,24 @@
...
+        "order": 11,
+        "type": "boolean",
+        "title": "Backfill",
+        "description": "Backfill data from the start date to the end date."
+      },
+      "start_date": {
+        "order": 12,
+        "type": "string",
+        "title": "Start Date",
+        "description": "The date from which to start syncing data. Only use for backfilling."
```
ypc-faros, line 101: nit: Remove `Only use for backfilling.`
We can technically use it for backfill or regular sync.

ypc-faros (approved):

PR #1657, opened 2024-08-29
FAI-12550 Add option to always full sync tasks and skip costly project-tasks streams

ypc-faros:
## Description

There are cases where the project-tasks relationship is so costly that it accounts for the majority of the records pulled in a sync.

This PR introduces an optimization where, if the number of tasks pulled in a full sync is << number of records in the project-tasks relationship (always full sync), it could be more efficient to pull tasks in full sync (the tasks records would include the associated projects) VS tasks incrementally + projects(full) + project-tasks(full).

If the `optimize_fetching_projects_and_tasks_with_full_tasks_sync` is enabled, we simply remove the unwanted streams from the catalog and add `tasks_full`. A new stream (pretty much an alias) is needed so that we make sure that the destination receives the appropriate destination sync mode (`OVERWRITE`)

Another wrinkle that I had to deal with is that the destination doesn't reset models unless ALL the streams that write the model have succeeded. In this case there's an intersection of models between `tasks_full` and (`projects`, `project_tasks`). I [marked](https://github.com/faros-ai/airbyte-connectors/pull/1657/files#diff-bdfcc178277d99d541ceb3db141b777acec0551173f74b0c17e4b868079903bcR25) these two as safe for reset (they'd never succeed, since they are not synced at all) so that the destination frees the respective models for reset.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/asana/project_tasks.ts
+++ b/destinations/airbyte-faros-destination/src/converters/asana/project_tasks.ts
@@ -15,6 +15,13 @@ export class ProjectTasks extends AsanaConverter {
     const source = this.streamName.source;
     const projectTask: ProjectTaskAssociation = record.record
       .data as ProjectTaskAssociation;
+    return ProjectTasks.convertProjectTask(projectTask, source);
+  }
+
+  static convertProjectTask(
+    projectTask: ProjectTaskAssociation,
+    source: string
+  ) {
     const taskKey = {uid: projectTask.task_gid, source};
     const res: DestinationRecord[] = [];
 
--- a/destinations/airbyte-faros-destination/src/converters/asana/projects.ts
+++ b/destinations/airbyte-faros-destination/src/converters/asana/projects.ts
@@ -16,9 +16,18 @@ export class Projects extends AsanaConverter {
     const source = this.streamName.source;
     const project: AsanaProject = record.record.data as AsanaProject;
 
+    return Projects.convertProject(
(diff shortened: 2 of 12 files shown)


dbruno21 (approved):

dbruno21:
> Another wrinkle that I had to deal with is that the destination doesn't reset models unless ALL the streams that write the model have succeeded

@ypc-faros I'm curious, what's the behavior if one of those streams is incremental?

ypc-faros:
> > Another wrinkle that I had to deal with is that the destination doesn't reset models unless ALL the streams that write the model have succeeded
> 
> @ypc-faros I'm curious, what's the behavior if one of those streams is incremental? Do the models get reset as well?

What I described only applies to full refresh (overwrite on the destination side) streams. Relevant code:
https://github.com/faros-ai/airbyte-connectors/blob/89e3b891b36af9498283919f757d03bdb585b9b2/destinations/airbyte-faros-destination/src/destination.ts#L1137
https://github.com/faros-ai/airbyte-connectors/blob/89e3b891b36af9498283919f757d03bdb585b9b2/destinations/airbyte-faros-destination/src/destination.ts#L559
https://github.com/faros-ai/airbyte-connectors/blob/89e3b891b36af9498283919f757d03bdb585b9b2/destinations/airbyte-faros-destination/src/converters/converter.ts#L136-L172

PR #1658, opened 2024-08-30
FAI-12835 | GitHub Security streams (sources)

dbruno21:
## Description

Adds to GitHub source 3 streams for Security module:
- Code Scanning Alerts
- Dependabot Alerts
- Secret Scanning Alerts

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/github-source/src/github.ts
```diff
@@ -1275,6 +1281,152 @@ export abstract class GitHub {
     }
   }
 
+  async *getCodeScanningAlerts(
+    org: string,
+    repo: string,
+    startDate?: Date,
```
ypc-faros, line 1287: The dates don't need to be optional. We're guaranteed to have a start, end.

sources/github-source/src/github.ts
```diff
@@ -1275,6 +1281,152 @@ export abstract class GitHub {
...
+        per_page: this.pageSize,
+        sort: 'updated',
+        direction: 'desc',
+      }
+    );
+    try {
+      for await (const res of iter) {
+        for (const alert of res.data) {
+          if (
+            this.backfill &&
```
ypc-faros, line 1304: We should check for this regardless of the value of `backfill`

To simplify logic like this across all sources, I think we should:

1) Always calculate start, end dates based on the relevant config fields (start date, end date, cutoff)
2) Always fetch data for the range [start, end)

The `backfill` flag should only be accessed by the cdk (if possible) to determine whether to use the state or not.

ypc-faros (approved):
🚀

sources/github-source/src/github.ts
```diff
@@ -894,6 +897,9 @@ export abstract class GitHub {
     } catch (err: any) {
       this.handleCopilotError(err, org, 'seats');
     }
+    this.logger.debug(
```
ypc-faros, line 900: Looks unrelated to this change. Flagging in case you need to remove

sources/github-source/src/github.ts
dbruno21: The reason I check if it's a backfill here and other incremental streams is because: let's say this is not a backfill, the end date is generated as the current timestamp at the beginning of the sync. My thought was there is no reason to skip records that could have been updated during the sync. We are already fetching them after all since the API doesn't provide a way to filter by range (not even one of since/until params) in most cases. It isn't a big impact, but don't see why it shouldn't be that way. Maybe would be cleaner / simpler if an end date is not calculated (leave as null) when it's not a backfill.

sources/github-source/src/github.ts
dbruno21: It's ok, I want to add it

sources/github-source/src/github.ts
ypc-faros: Fair point. I think it's a tradeoff then between saving in API calls for that delta past the calculated end date and less code. I'd personally go for less code (saves all the boilerplate of wiring through the `backfill` value and simplifies the predicate), but I don't have a strong opinion against the other.

PR #1665, opened 2024-09-02
FAI-12274: Bitbucket commits stream

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/converters/bitbucket/common.ts
```diff
@@ -127,4 +135,36 @@ export abstract class BitbucketConverter extends Converter {
       ) ?? BitbucketCommon.MAX_DESCRIPTION_LENGTH
     );
   }
+
+  protected collectUser(user: PartialUser, workspace: string): void {
+    if (!user?.accountId) return;
+    this.collectedUsers.set(user.accountId, {...user, workspace});
+  }
+
+  protected convertUsers(): DestinationRecord[] {
```
ypc-faros, line 144: Does this deal with records of the same user with/without some attributes? E.g., a record of the user with `displayName` and another of the same user without `displayName`?

I think we did some special handling for that in the GH converters...

chalenge (approved):
lgtm

ypc-faros (approved):
Fix tests (probably only need to update a snap)

PR #1666, opened 2024-09-03
FAI-12836 | GitHub Security streams (converters)

dbruno21:
## Description

Adds GitHub converters for the Security module 3 streams:

- Code Scanning Alerts
- Dependabot Alerts
- Secret Scanning Alerts

Source responses samples + schemas used to build the mappings:
- https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-a-repository
- https://docs.github.com/en/rest/dependabot/alerts?apiVersion=2022-11-28#list-dependabot-alerts-for-a-repository
- https://docs.github.com/en/rest/secret-scanning/secret-scanning?apiVersion=2022-11-28#list-secret-scanning-alerts-for-a-repository

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


dbruno21:
Here's the fields that couldn't be filled for the models within each converter https://github.com/faros-ai/airbyte-connectors/pull/1666/commits/ebe86f4f0c4f8d1c8e210a94550c759c614797f7

dbruno21:
For `sec_Vulnerability` model, I think we should add a `type` field and start with these possible standard values (we can define better names):
- Security (e.g. vanta + code scanning security alerts)
- CodingError (e.g. code scanning non-security alerts)
- SecretLeak (e.g. secret scanning)

cc @chalenge

destinations/airbyte-faros-destination/src/converters/github/faros_code_scanning_alerts.ts
```diff
@@ -0,0 +1,80 @@
...
+    const repository = GitHubCommon.repoKey(
+      alert.org,
+      alert.repo,
+      this.streamName.source
+    );
+    return [
+      {
+        model: 'sec_Vulnerability',
+        record: {
+          uid: alert.html_url,
```
ypc-faros, line 27: Not very in love with having the full URL as the `uid` here. Since the vulns are scoped to the repo, maybe we can do a concatenation of org/repo/vuln number?

dbruno21: That was my other alternative, but code-scanning / dependabot / secret-scanning should be added as well to avoid collision. Then I thought that the html_url is already pretty much it e.g. https://github.com/octocat/hello-world/code-scanning/4, and thought it could fit as well.

ypc-faros: I'd still prefer the concatenation. Will save storage, although probably not meaningful saving depending on volume. Better indexing if the `uid` has an index. And more importantly for me, readability and independence of GH URL scheme.

chalenge: Yeah, this doesn't look right. I would assume we should be able to have the same vulnerabilites be linked to multiple repos.

dbruno21: > Yeah, this doesn't look right. I would assume we should be able to have the same vulnerabilites be linked to multiple repos.

I don't think @ypc-faros meant the same.

I also thought at first that maybe a `sec_Vulnerability` could be a unique reference to a vulnerability, but:
1. The way it's done in Vanta source, it's already using the vanta uid which is the uid for tracking a vulnerability found within a system. (@omreego could you confirm this?)
2. I think it would be very hard to come up with a way to identify uniquely vulnerabilities. It's easy for CVE ones where we could just use the CVE ID maybe. But there are other finding that doesn't include or have a correlation to a global vulnerability database.
3. Do we really care of tracking it that way (like a vulnerability database) instead of just tracking the findings? We are already storing any external vulnerability id's (e.g. CVE) after all.

dbruno21: @ypc-faros Updated uid to be built with the concatenation of keys

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -219,6 +219,15 @@ export class GitHubCommon {
     };
   }
 
+  static vulnerabilityUid(
+    org: string,
+    repo: string,
+    alertType: string,
+    number: number
+  ): string {
+    return `${org}/${repo}/${alertType}/${number}`;
```
ypc-faros, line 228: nit: Let's do `toLower` on org/repo (or on the whole thing)

ypc-faros (dismissed):

destinations/airbyte-faros-destination/src/converters/github/faros_code_scanning_alerts.ts
omreego: Vanta source uses the vanta uid yes

omreego: sec_Vulnerability is a unique vulnerability, but it can be linked to multiple repos via the vcs_RepositoryVulnerability object

dbruno21: > sec_Vulnerability is a unique vulnerability, but it can be linked to multiple repos

@omreego It can be linked to multiple repos technically, but the records we are converting to sec_Vulnerability records are tight to a vulnerability finding on 1 repo, right? i.e. the uid returned by vanta is unique for a vulnerability finding within only ONE repo. Not saying it's bad, just trying to align how we think of this model.

ypc-faros:
Oops, missed your [comment](https://github.com/faros-ai/airbyte-connectors/pull/1666#discussion_r1742674430) @chalenge 
I guess we should look into the feasibility of your proposal

ypc-faros (commented):
Undo LGTM until we discuss what to do 😬

destinations/airbyte-faros-destination/src/converters/github/faros_code_scanning_alerts.ts
omreego: In the case of vanta, you are correct that they are tied 1 per repo, but that doesn't mean we need to have unique sec_Vulnerability per repo in all cases. When we create the table in Faros, the vcs_Vuln and cicd_Vuln objects are the ones that are counted as a single vuln. You can also make use of the "vulnerabilityIds" to take in CVE, etc if those are available

destinations/airbyte-faros-destination/src/converters/github/faros_code_scanning_alerts.ts
```diff
@@ -0,0 +1,86 @@
...
+            Utils.toDate(alert.dismissed_at) ??
+            null,
+          status: getStatus(alert),
+        },
+      },
+    ];
+  }
+}
+
+function getSeverity(alert: CodeScanningAlert) {
```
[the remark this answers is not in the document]
dbruno21: The field on the schema is [described](https://github.com/faros-ai/canonical-models/blob/8512c967608b16013aebac3bcdab1f1b7f3e4f77/resources/models/sec.gql#L13) to be a number between 0-10.

@omreego I took the mappings exactly as they are on Vanta source, is this something standard or something you came up with?

destinations/airbyte-faros-destination/src/converters/github/faros_dependabot_alerts.ts
```diff
@@ -0,0 +1,102 @@
...
+  async convert(
+    record: AirbyteRecord
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    const alert = record.record.data as DependabotAlert;
+    const repository = GitHubCommon.repoKey(
+      alert.org,
+      alert.repo,
+      this.streamName.source
+    );
+    const uid = GitHubCommon.vulnerabilityUid(
```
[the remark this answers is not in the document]
dbruno21: Updated to make use of the new model for relating to vulnerability identifiers (CVE, GHSA)

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -219,6 +219,15 @@ export class GitHubCommon {
     };
   }
 
+  static vulnerabilityUid(
+    org: string,
+    repo: string,
+    alertType: string,
+    number: number
+  ): string {
+    return toLower(`${org}/${repo}/${alertType}/${number}`);
```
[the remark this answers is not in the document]
dbruno21: `sec_Vulnerability` will be left to be used as a vulnerability finding and `sec_VulnerabilityIdentifier` should be used for relating the same vulnerability to multiple findings

destinations/airbyte-faros-destination/src/converters/github/common.ts
```diff
@@ -219,6 +227,84 @@ export class GitHubCommon {
...
+      default:
+        return {category: 'Custom', detail: state};
+    }
+  }
+
+  static vulnerabilitySeverity(alert: SecurityAlert) {
+    const level =
+      (alert as CodeScanningAlert).rule?.security_severity_level ??
+      (alert as DependabotAlert).security_vulnerability?.severity;
+    switch (level) {
```
chalenge, line 296: can we document where this category values came from, please

destinations/airbyte-faros-destination/src/converters/github/faros_code_scanning_alerts.ts
```diff
@@ -0,0 +1,60 @@
...
+    );
+    return [
+      {
+        model: 'sec_Vulnerability',
+        record: {
+          uid,
+          source: this.streamName.source,
+          type: GitHubCommon.vulnerabilityType(alert, this.alertType),
+          title: alert.rule.name ?? alert.rule.id,
+          description: Utils.cleanAndTruncate(alert.rule.description, 200),
```
chalenge, line 39: nit: why 200?

chalenge (approved):
lgtm, see nits

destinations/airbyte-faros-destination/src/converters/github/common.ts
dbruno21: The field on our schema is [described](https://github.com/faros-ai/canonical-models/blob/8512c967608b16013aebac3bcdab1f1b7f3e4f77/resources/models/sec.gql#L13) to be a number between 0-10.

@omreego I took the mappings exactly as they are on Vanta source, is this something standard or something you came up with?

omreego: Yes exactly

PR #1667, opened 2024-09-03
FAI-12718: Add Jira project version release status

chalenge:
## Description

- Add Jira project version release status
- Limit project versions we sync based on descending release date

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/jira-source/src/streams/common.ts
```diff
@@ -155,3 +166,38 @@ export abstract class StreamWithBoardSlices extends StreamBase {
     }
   }
 }
+
+export abstract class ProjectStreamSliceWithStaticCutoff extends StreamWithProjectSlices {
+  getUpdatedState(
```
matiaslcoulougian, line 171: Maybe we could add an extra comment to explain the use of this getUpdatedState function and its difference with 
the one that uses [getUpdatedStreamState](https://github.com/faros-ai/airbyte-connectors/blob/a3027379552561b3b0f03e9b5682d657bc379ba9/sources/jira-source/src/streams/common.ts#L123)

matiaslcoulougian (approved):
It took me some time to understand the logic but it looks good 👍

sources/jira-source/src/streams/common.ts
chalenge: Added comment. Let me know if it makes sense and/or needs to be updated

PR #1691, opened 2024-09-17
[FAI-12637] Write source and destination connector version to Poseidon

jeniii:
## Description

> Provide description here

todo: make sure local airbyte run also pushing the version.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/destination.ts
+++ b/destinations/airbyte-faros-destination/src/destination.ts
@@ -74,6 +74,7 @@ interface SyncErrors {
   dst: SyncMessage[];
 }
 
+// This is used to wrap the version string and allow us to pass the source version back via function calls.
 interface SourceVersion {
   version?: string;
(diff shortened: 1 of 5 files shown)


destinations/airbyte-faros-destination/src/destination.ts
```diff
@@ -73,6 +74,10 @@ interface SyncErrors {
   dst: SyncMessage[];
 }
 
+interface SourceVersion {
```
cjwooo, line 77: Comment here explaining that it's wrapped in an interface because the version is fetched and set later

cjwooo (approved):

destinations/airbyte-faros-destination/src/destination.ts
jeniii: will do

PR #1696, opened 2024-09-19
Add implementation of Zephyr APIs

chalenge:
## Description

- Add implementation of Zephyr APIs

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/zephyr-source/src/types.ts
+++ b/sources/zephyr-source/src/types.ts
@@ -8,10 +8,11 @@ export interface UserPassword {
 }
 
 export type Authentication = AccessToken | UserPassword;
+export type Project<T = any> = T[];
 export interface ZephyrConfig {
   readonly url: string;
   readonly authentication: Authentication;
-  readonly projects: ReadonlyArray<string>;
+  readonly projects: ReadonlyArray<Project>;
   readonly api_timeout?: number;
   readonly api_page_limit?: number;
   readonly api_max_retries?: number;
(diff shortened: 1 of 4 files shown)


sources/zephyr-source/src/types.ts
```diff
@@ -8,10 +8,11 @@ export interface UserPassword {
 }
 
 export type Authentication = AccessToken | UserPassword;
+export type Project<T = any> = T[];
```
matiaslcoulougian, line 11: Can we define the Project type?

sources/zephyr-source/src/zephyr.ts
```diff
@@ -119,23 +120,174 @@ export class Zephyr {
...
+  ): Promise<ReadonlyArray<any>> {
     const testExecutions: any[] = [];
-    testExecutions.push({
-      id: 1,
-      key: 'key',
-      name: 'name',
-    });
+    const testCycles = await this.getTestCycles(project);
+
+    // filter fetch cycles to match the configured cycles
```
matiaslcoulougian, line 229: We could define a function to filter versions and cycles with configured lists

dbruno21 (approved):

matiaslcoulougian (approved):
Some nits but looks good

sources/zephyr-source/src/types.ts
chalenge: yeah, will add types once we get sample data

PR #1697, opened 2024-09-19
FAI-13091 Improve resolving full merge commit sha

ypc-faros:
The API used by the PRs stream returns the short sha of the merge commit. The current solution to resolve the full sha is to accumulate all the commit records in memory (destination side) and, in the PRs converter, try to find the full commit sha that start with the prefix (short sha). This is wasteful both in terms of memory (we keep the full commit records in memory, not just the `sha`s) and time => linear search for the match.

This PR proposes a solution on the source side: while reading the commits stream, we add the full `sha`s to a `Set` in memory. While reading the PRs stream, if we find that the PR was merged and we have a short sha for the merge commit, we look it up using binary search over all the full commit `sha`s collected previously. This is better in terms of memory (only storing the full sha of the commits in memory) and lookup time => O(k log n) (`k`: length of short sha, `n`: number of full `sha`s)

I thought of a couple of more elegant (and also kind of too complicated vs a binary search) alternatives (prefix tree, map of prefixes to full sha) but they'd improve the runtime at the expense of more memory usage. I opted for this one since the sorting only happens once in a sync, and the complexity of the lookup is reasonable.


matiaslcoulougian (approved):
Nice! 🔥👌🏽

matiaslcoulougian:
Nit: remove unused imports marked by Sonarcloud

PR #1699, opened 2024-09-25
FAI-12277 Improve Bitbucket PR diff stats 

ypc-faros:


destinations/airbyte-faros-destination/src/converters/bitbucket/pull_requests_with_activities.ts
```diff
@@ -85,15 +102,55 @@ export class PullRequestsWithActivities extends BitbucketConverter {
...
+          additions: diff.linesAdded,
+          deletions: diff.linesRemoved,
+        },
+      };
+    };
+
+    if (diff.old?.path) {
+      res.push(processDiffPath(diff.old.path));
+    }
+    if (diff.new?.path && diff.new.path !== diff.old?.path) {
```
matiaslcoulougian, line 141: If new path != old path, should we still process old path?

matiaslcoulougian: The only case they are both old and new defined and different from each other is when it is a file rename. In that case we would write both files with 0 lines added and 0 lines removed.

matiaslcoulougian (approved):
LGTM. Remove unused imports

destinations/airbyte-faros-destination/src/converters/bitbucket/pull_requests_with_activities.ts
ypc-faros: Yeah 👍 
Added one test record to illustrate this case: https://github.com/faros-ai/airbyte-connectors/pull/1699/commits/81cfdfab33c40618769180d1030e194c9e40fa1a
It's working as intended

PR #1706, opened 2024-09-30
FAI-13080 | FarosGraph query mode for Jira boards

dbruno21:
## Description

Added boolean parameter and logic to use Faros Graph for Jira boards selection (inclusions / exclusions)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/jira-source/resources/spec.json
+++ b/sources/jira-source/resources/spec.json
@@ -112,7 +112,7 @@
         "order": 15,
         "type": "boolean",
         "title": "Use Faros Graph for Boards selection",
-        "description": "Query mode for selecting which Jira boards should be fetched (Faros credentials are required).",
+        "description": "Use Faros Graph for selecting which boards should be fetched (Faros credentials are required).",
         "default": false
       },
       "projects": {
--- a/sources/jira-source/resources/spec.json
+++ b/sources/jira-source/resources/spec.json
@@ -257,7 +257,7 @@
       "query_mode": {
         "order": 33,
         "title": "Query Mode",
-        "description": "Query mode for selecting what Jira boards should be fetched",
+        "description": "Query mode for selecting which Jira boards should be fetched",
         "type": "string",
         "default": "StaticLists",
         "enum": [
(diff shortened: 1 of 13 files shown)


sources/jira-source/resources/spec.json
```diff
@@ -253,6 +253,17 @@
         "type": "string",
         "title": "Organization ID",
         "description": "Atlassian organization ID for fetching teams. Required only when using Fetch Teams with Jira Cloud. See https://confluence.atlassian.com/jirakb/what-it-is-the-organization-id-and-where-to-find-it-1207189876.html for information on how to find it."
+      },
+      "query_mode": {
```
ypc-faros, line 257: I was never a fan of this enum :D 
Should we use a boolean toggle instead?

sources/jira-source/src/project-board-filter.ts
```diff
@@ -100,9 +183,53 @@ export class ProjectBoardFilter {
     }
   }
 
+  private async loadBoards(): Promise<void> {
```
ypc-faros, line 186: We should move this to `faros-airbyte-common` (doesn't need to happen in this PR)

sources/jira-source/resources/spec.json
ypc-faros: `use_faros_graph_for_board_selection` ?

sources/jira-source/src/project-board-filter.ts
dbruno21: I'll try to generalize it when redoing it for GitHub

sources/jira-source/resources/spec.json
dbruno21: @chalenge Any thoughts? Would you prefer a toogle / boolean instead?

sources/jira-source/resources/spec.json
```diff
@@ -108,8 +108,15 @@
         "title": "Use Users Prefix Search",
         "default": false
       },
-      "projects": {
+      "use_faros_graph_boards_selection": {
         "order": 15,
+        "type": "boolean",
+        "title": "Use Faros Graph for Boards selection",
+        "description": "Query mode for selecting which Jira boards should be fetched (Faros credentials are required).",
```
ypc-faros, line 115: Update description

ypc-faros (approved):

PR #1712, opened 2024-10-02
FAI-13204 | FarosGraph query mode for GitHub repositories

dbruno21:
## Description

Added boolean parameter and logic to use Faros Graph for GitHub repositories selection (inclusions / exclusions).
Moved shared logic to common.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/github-source/src/org-repo-filter.ts
```diff
@@ -132,4 +158,57 @@ export class OrgRepoFilter {
...
+    }
+    if (excludedRepositories?.length) {
+      collectReposByOrg(excludedReposByOrg, excludedRepositories);
+    }
+    for (const org of reposByOrg.keys()) {
+      if (excludedReposByOrg.has(org)) {
+        this.logger.warn(
+          `Both repositories and excluded_repositories are specified for organization ${org}, excluded_repositories for organization ${org} will be ignored.`
+        );
+        excludedReposByOrg.delete(org);
```
dbruno21, line 205: I'm not sure if keeping to apply inclusion / exclusion rules per each organization is worth or if it is confusing (both static lists and faros graph cases)

i.e. when using Faros graph let's say that I include faros-ai/airbyte-connectors, but I also have repos from other orgs where none are included / excluded. Right now, those repos from other orgs will be selected.

sources/github-source/src/github.ts
```diff
@@ -83,6 +83,8 @@ export const DEFAULT_FETCH_PR_REVIEWS = true;
 export const DEFAULT_CUTOFF_DAYS = 90;
 export const DEFAULT_BUCKET_ID = 1;
 export const DEFAULT_BUCKET_TOTAL = 1;
+export const DEFAULT_API_URL = 'https://prod.api.faros.ai';
```
ypc-faros, line 86: nit: `DEFAULT_FAROS_API_URL`

sources/github-source/src/github.ts
```diff
@@ -74,7 +74,7 @@ import {ExtendedOctokit, makeOctokitClient} from './octokit';
 import {RunMode} from './streams/common';
 import {AuditLogTeamMember, GitHubConfig, GraphQLErrorResponse} from './types';
 
-export const DEFAULT_API_URL = 'https://api.github.com';
+export const DEFAULT_URL = 'https://api.github.com';
```
ypc-faros, line 77: nit: `DEFAULT_GITHUB_API_URL`

ypc-faros (approved):
🔥

sources/github-source/resources/spec.json
```diff
@@ -149,90 +156,111 @@
...
       },
       "bucket_total": {
-        "order": 13,
+        "order": 14,
         "type": "integer",
         "title": "Total Number of Buckets",
         "description": "Total number of buckets to distribute organizations and repositories across. Use it when distributing the load between multiple sources",
         "default": 1
       },
+      "api_url": {
```
chalenge, line 200: this should be `faros_api_url`, same as below

sources/github-source/src/org-repo-filter.ts
chalenge: Hmmm. 
@cjwooo what was the Autodesk use case? Was it using static lists to discover/get repos, then use faros graph for included/excluded repos?

sources/github-source/resources/spec.json
dbruno21: I agree that would be better, but I thought the consistency with Jira was more important
https://github.com/faros-ai/airbyte-connectors/blob/df3012c7b2b79c36fc993627319c6dac6670c8b1/sources/jira-source/resources/spec.json#L204-L224

PR #1721, opened 2024-10-08
FAI-13320: Filter JIRA boards in a project by the board location not using projectKeyOrId query param

chalenge:
## Description

- When using `projectKeyOrId` query param for [GET All Boards API](https://developer.atlassian.com/cloud/jira/software/rest/api-group-board/#api-rest-agile-1-0-board-get) it `filters results to boards that are relevant to a project. Relevance means that the jql filter defined in board contains a reference to a project.` This seems to result in some boards in a project not showing up if they do not explicitly reference the project. 

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/jira-source/src/jira.ts
+++ b/sources/jira-source/src/jira.ts
@@ -867,11 +867,19 @@ export class Jira {
   }
 
   @Memoize()
-  async getBoards(
-    projectKey?: string
+  async getProjectBoards(
+    projectKey: string
   ): Promise<ReadonlyArray<AgileModels.Board>> {
+    const allBoards = await this.getAllBoards();
+    return allBoards.filter(
+      (board) => board.location?.projectKey === projectKey
+    );
+  }
+
+  @Memoize()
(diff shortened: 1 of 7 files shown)


ypc-faros (approved):

sources/jira-source/src/jira.ts
```diff
@@ -866,6 +866,33 @@ export class Jira {
...
+    // Get all boards located in the project
+    const allBoards = await this.getBoards();
+    allBoards.forEach((board) => {
+      if (board.location?.projectKey === projectKey) {
+        boardMap.set(board.id, board);
+      }
+    });
+
+    // Get boards relevant to the project
+    const projectBoards = await this.getBoards(projectKey);
```
ypc-faros, line 888: nit: Can we make `getBoards` private and maybe rename to `getRelevantBoards`. Don't want us to trip and use it accidentally from outside this class

ypc-faros (approved):

ypc-faros:
🚀

PR #1723, opened 2024-10-08
FAI-12648: Task board relationship hotspot

ted-faros:
## Description

Conditionally disable reset. 

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/jira-source/src/index.ts
+++ b/sources/jira-source/src/index.ts
@@ -84,12 +84,34 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
   }
 
   override async onBeforeRun(config: JiraConfig): Promise<JiraConfig> {
-    return {
-      ...config,
-      ...(config.use_faros_board_issue_tracker && {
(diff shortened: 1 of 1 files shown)


sources/jira-source/src/index.ts
```diff
@@ -84,12 +84,34 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
...
+    ) {
+      const farosClient = this.makeFarosClient(config);
+      try {
+        const res: any = await farosClient.request(
+          'GET',
+          `/accounts/${config.faros_source_id}/state`
+        );
+        return {
+          ...config,
+          ...(res.state && {
```
cjwooo, line 104: `res.state == true` if it's an empty object or array. Is that the desired outcome?

ted-faros: `res.state` is a string.  If that string is not empty, I want to reset.

cjwooo (approved):

PR #1730, opened 2024-10-10
FAI-13429 - ServiceNow client resiliency

willmarks:
## Description

We should stop processing and pickup where we left off if there is an issue. This PR also fixes the service identifier maps, they weren't working correctly.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/servicenow-source/src/servicenow/servicenow.ts
```diff
@@ -119,6 +127,7 @@ export class ServiceNow {
                 cmdb_ci_Map.set(cmdb_ci_sys_id, cmdb_ci_identifier);
               } catch (err: any) {
                 this.logger.warn(`Error retrieving cmdb_ci: ${cmdb_ci_sys_id}`);
+                cmdb_ci_Map.set(cmdb_ci_sys_id, undefined);
```
willmarks, line 130: Want to save the fact that we tried and failed to retrieve the service. This will save us from trying over and over on a service that isn't there.

snabar (approved):

PR #1738, opened 2024-10-14
FAI-13434 | Write labels on PRs/Issues streams and remove Labels stream from standard mode

dbruno21:
## Description

Write labels on PRs/Issues streams and remove Labels stream from standard mode

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


ypc-faros (approved):
Do we need the `faros_labels` stream at all?

dbruno21:
> Do we need the `faros_labels` stream at all?

I don't think we'll ever use it, but I think @chalenge had some reasons to keep it

PR #1743, opened 2024-10-16
FAI-13399 | GitHub request timeouts reimplemented to work with retries and rate limits

dbruno21:
## Description

GitHub request timeouts reimplemented to work with retries and rate limits

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/src/octokit.ts
+++ b/sources/github-source/src/octokit.ts
@@ -1,8 +1,11 @@
 import {createAppAuth} from '@octokit/auth-app';
+import {Octokit as OctokitCore} from '@octokit/core';
 import {paginateGraphql} from '@octokit/plugin-paginate-graphql';
 import {retry} from '@octokit/plugin-retry';
 import {throttling, ThrottlingOptions} from '@octokit/plugin-throttling';
-import {Octokit} from '@octokit/rest';
+import {RequestError} from '@octokit/request-error';
+import {Octokit as OctokitRest} from '@octokit/rest';
+import {EndpointDefaults, OctokitResponse} from '@octokit/types';
 import Bottleneck from 'bottleneck';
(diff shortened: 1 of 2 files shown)


sources/github-source/src/octokit.ts
```diff
@@ -167,3 +184,40 @@ function beforeRequestHook(request: any, logger: AirbyteLogger): void {
...
+          controller.abort(); // aborts request after timeout
+        }, timeoutMs);
+      });
+      try {
+        return (await Promise.race([
+          request(options),
+          timeout,
+        ])) as OctokitResponse<any, number>;
+      } catch (err: any) {
+        if (err.name === 'AbortError') {
```
chalenge, line 206: Instead of failing the process and not saving state, what you think about logging a warning and and exiting instead of throwing a failure, so we collect the results and start syncing from that point. See [this](https://github.com/faros-ai/airbyte-connectors/pull/1730/files#diff-5494a2e962a569b739abc3dfa232173be5f13929c3d21f0636a085048f6c26a5R108-R111).

Not sure how easy it will to implement that here

dbruno21: I understand, but do not think it's related to the PR (we could do it on another). The purpose of the PR is removing the timeout we were doing at a higher level (github.ts) that didn't consider that the request could have gone through rate limit / retries.

This specific line is catching the timeout / aborted request error and throwing a simulated 500 so that it's retried by the retry-plugin ([similar to this](https://github.com/octokit/plugin-retry.js/blob/de9e3d457ba247aa6e1c5ea4e47ec514dc0e147e/src/wrap-request.ts#L35-L48)).

chalenge (approved):

ypc-faros (approved):

PR #1749, opened 2024-10-18
Fix origin logic in dst for round robin syncs

ypc-faros:
When using `round_robin_bucket_execution`, we must suffix the `origin` with the bucket id so that resets only operate on records written by the corresponding sync (otherwise, e.g., the sync for bucket 2 would delete the records written by the sync for bucket 1 for full refresh streams)

In addition, when running a reset sync, we need to make sure we're deleting data for the origin AND for the origins with bucket suffix (which may or may not exist)

### First sync
Screenshot 2024-10-18 at 9 03 41 AM
Screenshot 2024-10-18 at 9 02 57 AM
Screenshot 2024-10-18 at 9 02 28 AM
### Second sync
Screenshot 2024-10-18 at 9 08 53 AM
Screenshot 2024-10-18 at 9 08 23 AM
Screenshot 2024-10-18 at 9 07 57 AM
### Reset sync
Screenshot 2024-10-18 at 11 11 51 AM
Screenshot 2024-10-18 at 11 39 16 AM

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-client.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-client.ts
@@ -29,6 +29,7 @@ import traverse from 'traverse';
 import {assert, Dictionary} from 'ts-essentials';
 import {VError} from 'verror';
 
+import {StreamNameSeparator} from '../converters/converter';
 import {OriginProvider} from './graphql-writer';
 import {
   DeletionRecord,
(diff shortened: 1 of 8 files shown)


destinations/airbyte-faros-destination/src/common/graphql-client.ts
```diff
@@ -371,18 +373,34 @@ export class GraphQLClient {
...
+      this.logger.info(
+        `Also resetting data for bucketed origins (matching ${origin}__bucket__<number>)`
+      );
+    }
+
+    const originFilter = isResetSync
+      ? {
+          _or: [
+            {origin: {_eq: origin}},
+            {origin: {_like: `${origin}__bucket__%`}},
```
dbruno21, line 391: instead of __ maybe it's better to use same constant as here https://github.com/faros-ai/airbyte-connectors/pull/1749/files#diff-03a77eca599dd7ac1e468cb3aacadf006ce5355ad78a9e54672669a413284eeeR113

dbruno21 (approved):

PR #1753, opened 2024-10-21
[FAI-13519] - GitHub Copilot usage recordedAt timestamp

crupley:
## Description

Populates the `recordedAt` field on `vcs_UserToolUsage` from the `emitted_at` timestamp on the source.

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
+++ b/destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
@@ -80,6 +80,7 @@ export class FarosCopilotSeats extends GitHubConverter {
         record: {
           userTool,
           usedAt: Utils.toDate(activeSeat.last_activity_at),
+          recordedAt: Utils.toDate(record.record.emitted_at),
         },
       });
     }
--- a/destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
+++ b/destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
@@ -34,6 +34,7 @@ Array [
   },
   Object {
     "vcs_UserToolUsage": Object {
+      "recordedAt": "2024-06-20T09:54:42.836Z",
       "source": "GitHub",
       "usedAt": "2021-10-14T06:53:32.000Z",
       "userTool": Object {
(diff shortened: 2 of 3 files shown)


destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
```diff
@@ -34,6 +34,7 @@ Array [
   },
   Object {
     "vcs_UserToolUsage": Object {
+      "recordedAt": "2024-06-20T09:54:42.836Z",
```
crupley, line 37: I updated this manually to match the value in the records file; is there a better way?

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
```diff
@@ -80,6 +80,7 @@ export class FarosCopilotSeats extends GitHubConverter {
         record: {
           userTool,
           usedAt: Utils.toDate(activeSeat.last_activity_at),
+          recordedAt: Utils.toDate(record.record.emitted_at),
```
crupley, line 83: Is it safe to use the `emitted_at` field from the stream here?

destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
dbruno21: Make sure to build before running converter tests:
`npm run build`

Run faros_github converter tests and pass the flag to update the snapshot:
`npm run test faros_github -- -u`

crupley: confirmed that `npm run test faros_github -- -u` yields the same snapshot file as what's here

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
ypc-faros: You can safely assume that field is present (it's written by the CDK)
It represents the point in time at which the record was emitted by the source

ypc-faros (approved):

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
crupley: That's exactly what I want to capture here. But maybe I should have asked: Is it safe to use a system field like this or would it be better to write a value from the source as part of the record `data` and pick that up in the destination?

ypc-faros: I wouldn't bother if it's effectively $now

PR #1755, opened 2024-10-25
FAI-13600 | Add retries for 422 error on PR files API (GitHub)

dbruno21:
## Description

Add retries for 422 error on PR files API (GitHub)
- Added custom plugin / hook because retries plugin doNotRetry parameter is fixed after initializing Octokit instance ([#1](https://github.com/octokit/plugin-retry.js/blob/e3f082e7b6dc996ca1d157b1a5c322ab708f1c84/src/error-request.ts#L10)  [#2](https://github.com/octokit/plugin-retry.js/blob/main/src/index.ts#L14))  and to avoid having to do retry logic manually.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/github-source/src/octokit.ts
+++ b/sources/github-source/src/octokit.ts
@@ -186,6 +186,9 @@ function beforeRequestHook(
   }
 }
 
+// Fake HTTP status code used by manually thrown errors to trigger retries by the retry-plugin
+const RETRYABLE_STATUS_CODE = 1000;
+
 function timeout(octokit: OctokitCore, octokitOptions: any) {
   const timeoutMs = octokitOptions.timeout?.ms;
   if (timeoutMs > 0) {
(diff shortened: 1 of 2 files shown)


sources/github-source/src/octokit.ts
```diff
@@ -204,10 +205,10 @@ function timeout(octokit: OctokitCore, octokitOptions: any) {
         ])) as OctokitResponse<any, number>;
       } catch (err: any) {
         if (err.name === 'AbortError') {
-          // simulate 500 so that retry plugin retries the request
+          // simulate request error so that retry plugin retries the request
           throw new RequestError(
             `GitHub request timed-out after ${timeoutMs} ms`,
-            500,
+            1000,
```
ypc-faros, line 211: What's this `1000` ?

dbruno21: While testing I thought that it could be confusing seeing an error log saying it's a 500 error when it wasn't actually
e.g. see this example (the log at the end) where I enabled retries for a 404 when querying a non-existant org
image

ypc-faros: Got it, thanks!

Let's extract it as a constant then `FAKE_STATUS` or similar with a small comment?

ypc-faros (approved):

PR #1766, opened 2024-11-05
FAI-13823: Better handling for geo locations and tests

chalenge:
## Description

- Fix nulls for address and co-ordinates
- Validate geo_Coordinates values are numbers

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/destinations/airbyte-faros-destination/src/converters/common/geo.ts
+++ b/destinations/airbyte-faros-destination/src/converters/common/geo.ts
@@ -67,7 +67,7 @@ export class LocationCollector {
       const rawCoordinates = location.coordinates;
       const lat = rawCoordinates?.lat ? toNumber(rawCoordinates.lat) : null;
       const lon = rawCoordinates?.lon ? toNumber(rawCoordinates.lon) : null;
-      const coordinates = isFinite(lat) || isFinite(lon) ? {lat, lon} : null;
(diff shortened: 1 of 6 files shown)


destinations/airbyte-faros-destination/src/converters/common/geo.ts
```diff
@@ -1,77 +1,113 @@
...
+        }
       }
-      if (location.coordinates) {
-        const coordinates = location.coordinates;
-        const coordinatesKey = `${coordinates.lat}-${coordinates.lon}`;
+
+      const rawCoordinates = location.coordinates;
+      const lat = rawCoordinates?.lat ? toNumber(rawCoordinates.lat) : null;
+      const lon = rawCoordinates?.lon ? toNumber(rawCoordinates.lon) : null;
+      const coordinates = isFinite(lat) || isFinite(lon) ? {lat, lon} : null;
```
ypc-faros, line 70: Shouldn't we require both to be valid?

chalenge: 🤷🏽  Was being less restrictive if one was a valid number. Fine will make it restrictive

ypc-faros (approved):
🚀

PR #1768, opened 2024-11-07
FAI-13924: Add ability to sync only users and not groups to Azure AD source

chalenge:
## Description

- Add ability to sync only users and not groups to Azure AD source

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/azureactivedirectory-source/src/index.ts
```diff
@@ -47,4 +50,24 @@ export class AzureActiveDirectorySource extends AirbyteSourceBase<AzureActiveDir
...
+    config: AzureActiveDirectoryConfig,
+    catalog: AirbyteConfiguredCatalog,
+    state?: AirbyteState
+  ): Promise<{
+    config: AzureActiveDirectoryConfig;
+    catalog: AirbyteConfiguredCatalog;
+    state?: AirbyteState;
+  }> {
+    const streamNames = [Users.name];
+    if (config.fetch_teams) {
```
ypc-faros, line 64: This is not honoring the default. Do `config.fetch_teams ?? true` ...

ypc-faros (approved):

PR #1769, opened 2024-11-07
FAI-13924: Fix stream filtering

chalenge:
## Description

- Should use snakecase names

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/azureactivedirectory-source/src/index.ts
+++ b/sources/azureactivedirectory-source/src/index.ts
@@ -60,9 +60,9 @@ export class AzureActiveDirectorySource extends AirbyteSourceBase<AzureActiveDir
     catalog: AirbyteConfiguredCatalog;
     state?: AirbyteState;
   }> {
-    const streamNames = [Users.name];
+    const streamNames = [Users.name.toLowerCase()];
     if (config.fetch_teams ?? true) {
-      streamNames.push(Groups.name);
+      streamNames.push(Groups.name.toLowerCase());
     }
     const streams = catalog.streams.filter((stream) =>
       streamNames.includes(stream.stream.name)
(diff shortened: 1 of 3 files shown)


ypc-faros:
I'm surprised you need to do this. See the impl for `name` : https://github.com/faros-ai/airbyte-connectors/blob/e763191a928ba4ed1b8b2c420f5eb406a690cf0f/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L23

ypc-faros:
Maybe a test would help figure out what's going on. See [here](https://github.com/faros-ai/airbyte-connectors/blob/e763191a928ba4ed1b8b2c420f5eb406a690cf0f/sources/jira-source/test/index.test.ts#L757-L764) for an example

chalenge:
> I'm surprised you need to do this. See the impl for `name` :
> 
> https://github.com/faros-ai/airbyte-connectors/blob/e763191a928ba4ed1b8b2c420f5eb406a690cf0f/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L23

That would only work if I had instantiated the class `new Users(config, this.logger)`, that was the bug otherwise it returns the JavaScript's built-in name property of a class. Not worth instantiate a whole class for a name here.

ypc-faros:
> > I'm surprised you need to do this. See the impl for `name` :
> > https://github.com/faros-ai/airbyte-connectors/blob/e763191a928ba4ed1b8b2c420f5eb406a690cf0f/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L23
> 
> That would only work if I had instantiated the class `new Users(config, this.logger)`, that was the bug otherwise it returns the JavaScript's built-in name property of a class. Not worth instantiating a whole class for a name here. Added test

OMG, js!!! 🤦‍♂️

ypc-faros (approved):

PR #1770, opened 2024-11-08
FAI-13835 | Write new copilot chat metrics and added setting to enable use of GA API

dbruno21:
## Description

Write new copilot chat metrics and added setting to enable use of GA API

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/github-source/resources/spec.json
```diff
@@ -196,98 +196,105 @@
         "description": "Fix the licenses dates for Copilot seats when using team assignments (requires Audit Log API)",
         "default": true
       },
-      "cutoff_days": {
+      "copilot_metrics_ga": {
```
dbruno21, line 199: We can remove this param later when beta API gets removed and GA API is the only option

sources/github-source/resources/spec.json
```diff
@@ -196,98 +196,105 @@
...
         "order": 14,
+        "type": "boolean",
+        "title": "Copilot Metrics GA",
+        "description": "Use Copilot Metrics GA API instead of Copilot Usage Beta API",
+        "default": false
```
dbruno21, line 204: Default to false until reports get updated with the new chat metrics

ypc-faros (approved):

PR #1773, opened 2024-11-13
FAI-13837 PR code coverage from commit statuses (codecov, codeclimate)

ypc-faros:


sources/bitbucket-source/test/resources/config.json
```diff
@@ -3,5 +3,7 @@
   "password": "secret",
   "bucket_id": 1,
   "bucket_total": 1,
-  "run_mode": "Custom"
+  "run_mode": "Custom",
+  "start_date": "2024-01-01",
```
ypc-faros, line 7: A test for bitbucket started failing out of the blue: https://github.com/faros-ai/airbyte-connectors/actions/runs/11823274716

We filter commits on [our side](https://github.com/faros-ai/airbyte-connectors/blob/923a9597351edc3a0aabb8be75acd7b164758279/sources/bitbucket-source/src/bitbucket.ts#L186) since the API doesn't support a time filter. One of the [mock records](https://github.com/faros-ai/airbyte-connectors/blob/923a9597351edc3a0aabb8be75acd7b164758279/sources/bitbucket-source/test/resources/commits/commits.json#L83) is too old (beyond the default cutoff) and is being filtered out 🫠

sources/github-source/src/github.ts
```diff
@@ -349,6 +389,79 @@ export abstract class GitHub {
...
+            `Failed to parse ${status.context} status description: ${status.description}`
+          );
+          return undefined;
+        }
+      };
+
+      for await (const res of iter) {
+        for (const status of res.data) {
+          if (status?.context === 'codeclimate/diff-coverage') {
+            codeClimateResult = processStatus(status);
```
dbruno21, line 443: why do we need to have different vars and conditions for codeClimate vs codeCov if they are processed in the same way? Is it for the case we add support to an alternative that needs to be processed differently?

dbruno21 (approved):

PR #1774, opened 2024-11-14
FAI-14047 | Write Copilot metrics to new specific model in Faros schema

dbruno21:
## Description

Write Copilot metrics to new specific model in Faros schema (`vcs_AssistantMetric`)

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
```diff
@@ -74,13 +75,41 @@ export class FarosCopilotSeats extends GitHubConverter {
         }),
       },
     });
-    if (activeSeat.last_activity_at) {
+    if (
+      activeSeat.last_activity_at &&
+      ctx?.config?.edition_configs?.edition !== Edition.COMMUNITY
```
dbruno21, line 80: Is it ok to leave this check and add the new model to CE schema later? @chalenge

chalenge: Yeah, fine. Can't remember @crupley do we have the CE Copilot dashboards done?

chalenge (approved):

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
```diff
@@ -75,14 +76,45 @@ export class FarosCopilotSeats extends GitHubConverter {
...
+          recordedAt,
         },
       });
+      if (ctx?.config?.edition_configs?.edition !== Edition.COMMUNITY) {
+        res.push({
+          model: 'vcs_AssistantMetric',
+          record: {
+            uid: [
+              GitHubTool.Copilot,
+              AssistantMetric.LastActivity,
```
ypc-faros, line 95: Maybe we should compute a hash based on all these components? I assume we only need to guarantee uniqueness, since the individual parts are in the record itself...

dbruno21: I think there would be no benefit in doing that. We would still rely on passing manually to the hash function the desired components to end up with an opaque string. We could still use the uid for debugging.

ypc-faros: I usually prefer these kind of ids to be opaque. I just don't like downstream consumers of the data to rely on a specific format/scheme of the key.

What way would you use it for debugging that is not doable if the uid is hashed?

dbruno21: > What way would you use it for debugging that is not doable if the uid is hashed

The only case would be 2 different records that I'd have expected to write the same uid (or the opposite 2 records that should be writing a different uid) and then I'd need to realize how did I mess it up 😅

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
```diff
@@ -75,14 +76,45 @@ export class FarosCopilotSeats extends GitHubConverter {
...
           userTool,
-          usedAt: Utils.toDate(activeSeat.last_activity_at),
-          recordedAt: Utils.toDate(record.record.emitted_at),
+          usedAt: lastActivityAt,
+          recordedAt,
         },
       });
+      if (ctx?.config?.edition_configs?.edition !== Edition.COMMUNITY) {
+        res.push({
+          model: 'vcs_AssistantMetric',
```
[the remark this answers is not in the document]
dbruno21: Is it safe to add if this model is thought to be written by 2 different streams (`faros_copilot_seats` and `faros_copilot_usage`) ?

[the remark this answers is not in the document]
ypc-faros: Yeah it shouldn't be an issue (since these are the stream for which we never delete data, right?).
We still need to declare the models so that the Airbyte reset sync works properly

PR #1777, opened 2024-11-15
FAI-14050 | Improved GitHub PR commentCount consistency and issue comments stream

dbruno21:
## Description

Improved GitHub PR commentCount consistency and added issue comments stream

There can be different type of comments on GitHub PRs:
1. Pull request review comments (comments that are made on a line in the diff - Files changes tab)
2. Issue comments (comments that are made not on the diff - Conversation tab)
3. Pull request review submission comment (the optional comment that can be sent when submitting review)

The behavior before this change was summing up 1 + 2 for vcs_PullRequest.commentCount (on Pull Requests stream) and writing vcs_PullRequestComment's just for 1 (on Pull Request Comments stream).

With this change:
- vcs_PullRequest.commentCount is always the sum of 1 + 2 + 3
- 1 will be written if Pull Request Comments stream is enabled
- 2 will be written if Issue Comments stream is enabled
- 3 will be written if Fetch PR Reviews option is enabled (on Pull Request stream)

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/github/faros_issue_comments.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  readonly destinationModels: ReadonlyArray<DestinationModel> =
+    this.alias.destinationModels;
+
+  async convert(
+    record: AirbyteRecord
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    const comment = record.record.data as IssueComment;
+    // skip issue comments that were not made on a pull request
+    if (comment.html_url.split('/').reverse()[1] !== 'pull') {
```
chalenge, line 19: can we filter these out in the source? Saw field `issue_url` is it always populated even for PRs?

chalenge (approved):
Nice 🚀

destinations/airbyte-faros-destination/src/converters/github/faros_issue_comments.ts
dbruno21: > Saw field issue_url is it always populated even for PRs

Yes, since every PR is an issue as well, and the only field found on the response from this API that we can use to determine if the issue is a PR is just this `html_url`. We could move this check to skip non-PR comments on the source instead, but thought it's more correct to be here.

PR #1756, opened 2024-10-28
FAI-13464: Write board inclusion for Jira boards when fetching board issues

chalenge:
## Description

- Write board inclusion for Jira boards when fetching board issues

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

PR #1778, opened 2024-11-19
FAI-13464 | Write faros options for repos being synced (GitHub)

dbruno21:
## Description

Write faros options for repos being synced (GitHub)
Same behavior as implemented for Jira boards on https://github.com/faros-ai/airbyte-connectors/pull/1756

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/github-source/resources/schemas/farosRepositories.json
```diff
@@ -45,6 +45,9 @@
     },
     "archived": {
       "type": "boolean"
+    },
+    "syncNestedData": {
```
dbruno21, line 49: @chalenge this is what you called syncIssues in Jira, but couldn't think of a better name for this case

chalenge: `syncRepoData`

chalenge (approved):

PR #1782, opened 2024-11-21
FAI-13692 Always rewrite the PR/file associations + Add explicit FLUSH operation

ypc-faros:
The graph after opening this PR with a single file changed.
Screenshot 2024-11-21 at 1 48 21 PM
With a dummy change to the README
Screenshot 2024-11-21 at 1 50 51 PM
Reverting the change in the README
Screenshot 2024-11-21 at 1 53 46 PM

### DIFF
--- a/destinations/airbyte-faros-destination/src/common/graphql-writer.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-writer.ts
@@ -38,6 +38,11 @@ export class GraphQLWriter {
       );
       return false;
     } else if (Object.values(Operation).includes(operation as Operation)) {
+      if (operation === Operation.FLUSH) {
+        await this.graphQLClient.flush();
+        return true;
+      }
+
       const timestampedRecord: TimestampedRecord = {
         ...result.record,
         model: baseModel,
--- a/destinations/airbyte-faros-destination/src/common/graphql-writer.ts
+++ b/destinations/airbyte-faros-destination/src/common/graphql-writer.ts
@@ -56,7 +56,9 @@ export class GraphQLWriter {
       // been processed yet.
       if (operation === Operation.DELETION && !result.record.at) {
         await this.writeTimestampedRecord(timestampedRecord);
-        await this.graphQLClient.flush();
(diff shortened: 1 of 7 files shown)


destinations/airbyte-faros-destination/src/common/types.ts
```diff
@@ -63,3 +64,8 @@ export interface DeletionRecord extends TimestampedRecord {
   operation: Operation.DELETION;
   where: Dictionary<any>;
 }
+
+export const FLUSH = {
```
chalenge, line 68: does this force an commit/write in the DB

chalenge (approved):

destinations/airbyte-faros-destination/src/common/types.ts
ypc-faros: Yes, it flushes the graphql buffers

ypc-faros: Actually, this might result in an excessive number of calls to Hasura (if the sync is pulling many PRs)

I think I should accumulate the deletes (1) and the association records (2). When the stream finishes processing then: process (1), FLUSH and process (2). (so that we take advantage of batching)

ted-faros: @ypc-faros sgtm.

destinations/airbyte-faros-destination/src/common/graphql-writer.ts
```diff
@@ -51,7 +56,9 @@ export class GraphQLWriter {
       // been processed yet.
       if (operation === Operation.DELETION && !result.record.at) {
         await this.writeTimestampedRecord(timestampedRecord);
-        await this.graphQLClient.flush();
+        if (result.record.flushRequired ?? true) {
```
ypc-faros, line 59: I had to do this to maintain backwards compatibility.

Maybe I'll hunt down all occurrences and swap the default if someone has strong opinions. (In another PR)

ted-faros (approved):
lgtm

PR #1784, opened 2024-11-22
FAI-13813: Add Tromzo findings converter

chalenge:
## Description

- Add Tromzo Findings converter
- Add shared Vulnerabilities utility for mapping scores and generating keys for types

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/tromzo/findings.ts
```diff
@@ -0,0 +1,145 @@
...
+    finding: any,
+    vulnerabilityKey: {uid: string; source: string}
+  ): DestinationRecord {
+    return {
+      model: 'vcs_RepositoryVulnerability',
+      record: {
+        vulnerability: vulnerabilityKey,
+        repository: GitHubCommon.parseRepositoryKey(
+          finding?.asset?.name,
+          'GitHub' // TODO - Figure out what source to use
```
matiaslcoulougian, line 114: If the API response doesn't contain a field to determine the integration from which the vulnerability is coming, we might need to query repos from Faros Graph.

matiaslcoulougian (approved):

destinations/airbyte-faros-destination/src/converters/tromzo/findings.ts
chalenge: Will update in next PR. Merging to unblock

PR #1814, opened 2024-11-27
Fix azure repos name separator

dbruno21:
## Description

Fix azure repos name separator:
- Use `:` instead of `_` as separator between project and repo for vcs_Repository.name because project name or repo name could contain `_` chars making it harder to identify the parts
- Added azure repos converter records test snapshot

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


ypc-faros (approved):

destinations/airbyte-faros-destination/src/converters/azure-repos/common.ts
```diff
@@ -52,7 +52,7 @@ export abstract class AzureReposConverter extends Converter {
    * @returns           The identifier for a repo unique across projects
    */
   getProjectRepo(repository: CommitRepository): string {
-    return `${repository?.project?.name}_${repository?.name}`;
+    return `${repository?.project?.name}/${repository?.name}`;
```
chalenge, line 55: how about `:`, worried about `/` since it can be used for paths

chalenge (approved):

PR #1819, opened 2024-11-28
Fix Azure Pipelines - Support individual state per slice

dbruno21:
## Description

Fix Azure Pipelines - Support individual state per slice

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

### DIFF
--- a/sources/azurepipeline-source/src/azurepipeline.ts
+++ b/sources/azurepipeline-source/src/azurepipeline.ts
@@ -10,7 +10,7 @@ import {
   Release,
   ReleaseResponse,
 } from 'faros-airbyte-common/azurepipeline';
-import {makeAxiosInstanceWithRetry} from 'faros-js-client';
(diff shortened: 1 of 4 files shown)


chalenge (approved):

ypc-faros:
I think that utility function expects the `cutoff` to be a number: https://github.com/faros-ai/airbyte-connectors/blob/472ac942066536f2f1ee7de69bab95057d5be70f/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L245

ypc-faros:
> I think that utility function expects the `cutoff` to be a number:
> 
> https://github.com/faros-ai/airbyte-connectors/blob/472ac942066536f2f1ee7de69bab95057d5be70f/faros-airbyte-cdk/src/sources/streams/stream-base.ts#L245

Maybe it works all the same but I'd prefer uniformity across sources

PR #1838, opened 2024-12-06
Tell jest to preprocess only typescript files

ypc-faros:
To get rid of the annoying noise in the tests output

Screenshot 2024-12-06 at 3 04 32 PM


chalenge (approved):
THANK YOU!!!! 🥇

PR #1839, opened 2024-12-06
Handle PR files unavailable error

dbruno21:
## Description

Handle PR files unavailable error

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/github-source/src/github.ts
+++ b/sources/github-source/src/github.ts
@@ -586,19 +586,25 @@ export abstract class GitHub {
         pull_number: number,
         per_page: this.pageSize,
         page: startingPage,
-        request: {
-          retryAdditionalError: (err: RequestError) => err.status === 422,
-        },
       }
     );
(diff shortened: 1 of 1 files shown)


sources/github-source/src/github.ts
```diff
@@ -586,19 +586,25 @@ export abstract class GitHub {
         pull_number: number,
         per_page: this.pageSize,
         page: startingPage,
-        request: {
-          retryAdditionalError: (err: RequestError) => err.status === 422,
```
dbruno21, line 590: Learnt that this thing doesn't work inside paginate.iterator(). Couldn't find a fix, but anyway, retries are not needed in this case since it's not recoverable when this API is failing like this based on what we've seen.

ypc-faros (approved):

chalenge (approved):

PR #1840, opened 2024-12-06
FAI-14237 - Fix azure repos incremental PR fetch logic

cjwooo:
## Description

We're currently fetching any non-completed PR regardless of any related timestamps.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/azure-repos-source/src/azure-repos.ts
```diff
@@ -429,8 +429,13 @@ export class AzureRepos {
       }
     )) {
       for (const pullRequest of pullRequestRes?.data?.value ?? []) {
-        const closedDate = DateTime.fromISO(pullRequest.closedDate);
-        if (pullRequest.status === 'completed' && closedDate <= since) {
```
cjwooo, line 433: I'm not sure why the `pullRequest.status === 'completed'` check here. @willmarks do you know?

ypc-faros (approved):

chalenge:
@cjwooo @willmarks what is left here. Can we close or merge?

cjwooo:
> @cjwooo @willmarks what is left here. Can we close or merge?

I'd like @willmarks to comment on the `pullRequest.status === completed` check before taking any action.

sources/azure-repos-source/src/azure-repos.ts
willmarks: I think that we wanted to continue to pull updates for PRs that were still in progress as there wasn't a great way to determine the last update so we needed to consider all incomplete PRs. If a PR is completed then we only wanted to consider it within the timeframe? Your proxy for lastUpdate this seems like a better idea. Perhaps we also want to consider other forms of updates before considering creation date tho? We just need to make sure that we don't miss long lived PR updates if they are not moved to closed.

willmarks:
Sorry @cjwooo, missed your question from last week. Let me know if what I said makes sense. Happy to discuss.

sources/azure-repos-source/src/azure-repos.ts
cjwooo: Hm so in order to get the comment timestamps we have make that second API call to get the PR threads, which I was hoping to avoid.

cjwooo: I'm going to close this PR -- I agree that we should take comment timestamps into account, so if we have to make the individual PR thread call anyways, we might as well upload the PR to Faros -- more data is better than less data.

cjwooo:
https://github.com/faros-ai/airbyte-connectors/pull/1840#discussion_r1884586413

PR #1843, opened 2024-12-08
Always persist state (even in backfill syncs)

ypc-faros:
It's safe to persist the state even when running in backfill because we don't modify the streams' state. But we want to persist the state anyway so that we advance the bucket id when using round robin bucketing in a backfill

### DIFF
--- a/destinations/airbyte-faros-destination/src/destination.ts
+++ b/destinations/airbyte-faros-destination/src/destination.ts
@@ -885,7 +885,7 @@ export class FarosDestination extends AirbyteDestination<DestinationConfig> {
           }
         });
 
-        if (stateMessage && !isBackfillSync) {
+        if (stateMessage) {
           yield stateMessage;
         }
       }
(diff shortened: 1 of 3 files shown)


chalenge:
Not sure I understand this part "because we don't modify the streams' state". If we persist state, how will we not override current stream state?

chalenge (commented):
Not sure I understand this part "because we don't modify the streams' state". If we persist state, how will we not override current stream state?

ypc-faros:
> Not sure I understand this part "because we don't modify the streams' state". If we persist state, how will we not override current stream state?

Added a test to illustrate my point @chalenge 

In `streams - pull requests backfill with bucketing and round robin execution only affects bucketing state` you'll see that even when the state should have changed for the `faros_pull_requests` stream (`github/hello-world` slice), it did not because we're using `backfill`. However the `__bucket_execution_state` section of the state changed because we want the `last_executed_bucket_id` to advance.

ypc-faros:
> > Not sure I understand this part "because we don't modify the streams' state". If we persist state, how will we not override current stream state?
> 
> Added a test to illustrate my point @chalenge
> 
> In `streams - pull requests backfill with bucketing and round robin execution only affects bucketing state` you'll see that even when the state should have changed for the `faros_pull_requests` stream (`github/hello-world` slice), it did not because we're using `backfill`. However the `__bucket_execution_state` section of the state changed because we want the `last_executed_bucket_id` to advance.

PS: It wasn't straightforward to add an equivalent test to the CDK tests. We should probably work on that...

chalenge (approved):

PR #1848, opened 2024-12-09
FAI-13813: Make Tromzo source incremental

chalenge:
## Description

- Make Tromzo source incremental
- Update findings gql with new fields

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


dbruno21 (approved):

sources/tromzo-source/src/streams/findings.ts
```diff
@@ -46,10 +48,43 @@ export class Findings extends AirbyteStreamBase {
...
+      [streamSlice.tool]: {
+        cutoff: cutoff.getTime(),
+      },
+    };
+  }
+
+  private getUpdateRange(cutoff?: number): [Date, Date] {
+    return [
+      cutoff ? Utils.toDate(cutoff) : this.config.startDate,
+      this.config.endDate,
```
ypc-faros, line 87: Didn't see `endDate` being used anywhere

sources/tromzo-source/src/types.ts
ypc-faros, line 1: (Doesn't need to happen in this PR but) we should move these to `faros-airbyte-common` so that we can share with the destination converters

Except for the `TromzoConfig`, which we usually keep in the client class itself

PR #1852, opened 2024-12-10
FAI-14379 | Override mode for GitHub / Jira

dbruno21:


sources/jira-source/src/index.ts
```diff
@@ -44,6 +44,13 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
     return 'jira';
   }
 
+  mode(config: JiraConfig): string | undefined {
+    if (!config.url) {
+      return undefined;
```
dbruno21, line 49: if url not present, throws error on checkConnection

tovbinm, line 49: How can this be no url? Let's assume cloud then.

sources/jira-source/src/index.ts
```diff
@@ -44,6 +44,13 @@ export class JiraSource extends AirbyteSourceBase<JiraConfig> {
...
+    }
+    return config.url.endsWith('.atlassian.net') ? 'cloud' : 'server';
```
ypc-faros, line 51: nit: Better check against [this](https://github.com/faros-ai/airbyte-connectors/blob/13818da80e10a69c28d18c412500d2fbc1850e29/sources/jira-source/src/jira.ts#L118) regex

ypc-faros (approved):

sources/jira-source/src/index.ts
chalenge: +1

sources/jira-source/src/jira.ts
```diff
@@ -115,7 +115,7 @@ const prRegex = new RegExp(
     'stateCount=(?<count>[0-9]+)})'
 );
 
-const jiraCloudRegex = /^https:\/\/(.*)\.atlassian\.net/g;
+export const JIRA_CLOUD_REGEX = /^https:\/\/(.*)\.atlassian\.net/g;
```
tovbinm, line 118: what if the url does not end with `/`?

dbruno21: There's no `/` at the end, only the regexp delimiter

PR #1853, opened 2024-12-11
FAI-13813: Bug fix state for comparing state

chalenge:
## Description

Fix checking for new cutoff

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/tromzo-source/src/streams/findings.ts
+++ b/sources/tromzo-source/src/streams/findings.ts
@@ -71,14 +71,20 @@ export class Findings extends AirbyteStreamBase {
     const currentState = Utils.toDate(
       currentStreamState?.[streamSlice.tool]?.cutoff
     );
-    const cutoff =
-      latestRecordCutoff > currentState ? latestRecordCutoff : currentState;
-    return {
-      ...currentStreamState,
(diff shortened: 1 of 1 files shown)


sources/tromzo-source/src/streams/findings.ts
```diff
@@ -71,14 +71,20 @@ export class Findings extends AirbyteStreamBase {
     const currentState = Utils.toDate(
       currentStreamState?.[streamSlice.tool]?.cutoff
     );
-    const cutoff =
-      latestRecordCutoff > currentState ? latestRecordCutoff : currentState;
```
chalenge, line 75: 🤦🏽

ypc-faros (approved):

sources/tromzo-source/src/streams/findings.ts
ypc-faros: We should look into providing a default impl for this. Most `getUpdatedState` are pretty similar.

chalenge: 💯. I'll look into it

PR #1855, opened 2024-12-11
FAI-13912 | GitHub - Change faros graph inclusion behavior when org has no filters

dbruno21:
## Description

Before this change when an organization had no inclusions / exclusions (faros graph options), all repositories for that organization would be synced and included. We only want this behavior if ALL the organizations have no inclusions / exclusions (e.g. only during first sync).

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


dbruno21:
No need to change after all

PR #1864, opened 2024-12-20
FAI-14493 Leverage workday converter for okta users

ypc-faros:
Also removed `groups` stream from source. It's not useful


destinations/airbyte-faros-destination/src/converters/okta/users.ts
```diff
@@ -1,93 +1,98 @@
...
+export class Users extends OktaConverter {
+  private workdayConverter = new Customreports();
+  private _config: OktaConfig = undefined;
 
-    const fullName = [profile.firstName, profile.middleName, profile.lastName]
-      .filter((x) => x)
-      .join(' ');
+  private initialize(ctx: StreamContext): void {
+    this._config =
+      this._config ?? ctx.config.source_specific_configs?.okta ?? {};
```
dbruno21, line 35: need to add to spec.json?

dbruno21 (approved):

PR #1867, opened 2024-12-24
FAI-13967, FAI-8202: Jira source add source name qualifier to write non-uid ids for multiple instances

chalenge:
## Description

- This will allow writing of Jira entities without UIDs from same source and multiple instances but have the same IDS.
- For example Jira ids are incrementally generated. When syncing data from two instances e.g. two Jira servers the system IDs for Sprints and Boards will be similar and our primary keys of ID and Source will not be sufficient to ensure uniqueness. The keys will override each other hence we need to add an option to differentiate each other.
- This will update associated object keys as well

Example tms_SprintBoardRelationship with for qualifier `TeamB`
```gql
{
  "tms_SprintBoardRelationship": {
    "board": {
      "source": "Jira",
      "uid": "49",
    },
    "sprint": {
      "source": "Jira",
      "uid": "55",
    },
  },
}
```
will map to 
```gql
{
  "tms_SprintBoardRelationship": {
    "board": {
      "source": "Jira_TeamB",
      "uid": "49",
    },
    "source": "Jira_TeamB",
    "sprint": {
      "source": "Jira_TeamB",
      "uid": "55",
    },
  },
}
```

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


ypc-faros:
Oof, not sure about the implications of this one. It can potentially open a can of worms since, for example, we sometimes query by source for updates/deletions (example [1](https://github.com/faros-ai/airbyte-connectors/blob/ca8fd94dd84df030ae58eaf654b3b9e5eae2a504/destinations/airbyte-faros-destination/src/converters/trello/cards.ts#L307) of many)

I understand the motivation, but is it possible to do this in the source instead 😬 ? Maybe we can change the system ids instead of the source part of the key?

chalenge:
> Oof, not sure about the implications of this one. It can potentially open a can of worms since, for example, we sometimes query by source for updates/deletions (example [1](https://github.com/faros-ai/airbyte-connectors/blob/ca8fd94dd84df030ae58eaf654b3b9e5eae2a504/destinations/airbyte-faros-destination/src/converters/trello/cards.ts#L307) of many)
> 
> I understand the motivation, but is it possible to do this in the source instead 😬 ? Maybe we can change the system ids instead of the source part of the key?

Basically I want to port this [change over](https://github.com/faros-ai/feeds/commit/331a2fea99df99f6f77fc476159b3c0cec6ccdf2). Think we decided updating the source instead. Good point about when we need to query about the source, so yes kinda also need it in the source too. Initial design was to preserve source the keys since those should remain recognizable for the user. Can keep it as a source specific config for now. I had thought to generalize completely.

ypc-faros:
> > Oof, not sure about the implications of this one. It can potentially open a can of worms since, for example, we sometimes query by source for updates/deletions (example [1](https://github.com/faros-ai/airbyte-connectors/blob/ca8fd94dd84df030ae58eaf654b3b9e5eae2a504/destinations/airbyte-faros-destination/src/converters/trello/cards.ts#L307) of many)
> > I understand the motivation, but is it possible to do this in the source instead 😬 ? Maybe we can change the system ids instead of the source part of the key?
> 
> Basically I want to port this [change over](https://github.com/faros-ai/feeds/commit/331a2fea99df99f6f77fc476159b3c0cec6ccdf2). Think we decided updating the source instead. Good point about when we need to query about the source, so yes kinda also need it in the source too. Initial design was to preserve source the keys since those should remain recognizable for the user. Can keep it as a source specific config for now. I had thought to generalize completely.

It's doable, we just need to be super careful 😬 
Some thoughts:

1. The config field should be both in the sources config and in the destination config (with the source config one taking precedence) so that we can support this for non Faros sources.
2. Make sure we use the qualified `source` in the sources code (I can only think of Faros [options](https://github.com/faros-ai/airbyte-connectors/blob/42e7e9bed3ded5765a644a4d22087ea7ca86b790/faros-airbyte-common/src/common/index.ts#L117) but we'd have to look with a fine tooth comb)
3. Same as (2) for the destination. Probably 99% of the work would be to replace [this](https://github.com/faros-ai/airbyte-connectors/blob/42e7e9bed3ded5765a644a4d22087ea7ca86b790/destinations/airbyte-faros-destination/src/converters/jira/faros_issues.ts#L54) kind of statements with a function that computes the qualified `source`.

I'm okay with starting with Jira and then generalizing

ypc-faros:
Another one: https://github.com/faros-ai/airbyte-connectors/blob/88a362d9bc0fae8a5d2139207c4c4ce734441c4d/sources/jira-source/src/project-board-filter.ts#L222

destinations/airbyte-faros-destination/src/converters/jira/common.ts
```diff
@@ -51,6 +53,19 @@ export abstract class JiraConverter extends Converter {
     return ctx.config.source_specific_configs?.jira ?? {};
   }
 
+  /**
+   * If the source is not qualified, we need to add the qualifier to the source name
+   */
+  protected initializeSource(ctx: StreamContext): string {
```
chalenge, line 59: @ypc-faros any other ideas for forcing this initialization?

ypc-faros: No good ideas, but I think that's fine for now.
We can maybe make it a responsibility of the context when we generalize

ypc-faros (approved):

PR #1907, opened 2025-02-03
FAI-14913: GitHub source enable fetching public organizations

chalenge:
## Description

- GitHub source enable fetching public repos
- Also try fetching for organization directly

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/sources/github-source/src/org-repo-filter.ts
+++ b/sources/github-source/src/org-repo-filter.ts
@@ -93,43 +93,74 @@ export class OrgRepoFilter {
   @Memoize()
   async getOrganizations(): Promise<ReadonlyArray<string>> {
     if (!this.organizations) {
-      const organizations = new Set<string>();
       const github = await GitHub.instance(this.config, this.logger);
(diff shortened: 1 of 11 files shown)


sources/github-source/src/org-repo-filter.ts
```diff
@@ -93,43 +93,74 @@ export class OrgRepoFilter {
...
+      return true;
     }
 
-    return Array.from(this.organizations);
+    // Attempt direct organization lookup if not in visibleOrgs
+    try {
+      await github.getOrganization(lowerOrg);
+      return true;
+    } catch (error) {
+      this.logger.warn(`Skipping not found organization ${lowerOrg}`);
```
dbruno21, line 161: We should check error is 404 else throw

chalenge: Fixed the message, we shouldn't throw at all here for one problematic organization. We have multiple other failure points. Will log the error and try processing.

dbruno21 (approved):

PR #1908, opened 2025-02-04
FAI-14070/FAI-14071: Add use_board_ownership flag to Jira Source

matiaslcoulougian:
## Description
We are adding use_board_ownership flag, like the one we had on the feed, to be able to keep using it with 1 board per project.

Default behavior is use_board_ownership:true, as it was working as today.

When flag is set to false:
- Board list is populated with project uids
- Project-board relation created 1-1 with project uid.
- The board and board_issues streams are removed from list.
- Task-board relationship is written on faros_issues stream using project uid.
- Sprints and sprint history records are associated to the project board. If the same sprint is in more than one Jira Board, it will override it with the last one fetched.

TODO
- [x] Add more tests
- [x] Refactor to reduce duplication

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


sources/jira-source/src/project-board-filter.ts
```diff
@@ -132,6 +132,10 @@ export class ProjectBoardFilter {
       // Ensure included / excluded boards are loaded
       await this.loadSelectedBoards();
 
+      if (!this.useBoardOwnership()) {
```
chalenge, line 135: how does this work with the loadSelectedBoards calls above. Do we need both?

sources/jira-source/src/project-board-filter.ts
```diff
@@ -214,6 +218,23 @@ export class ProjectBoardFilter {
...
+   * Use the Jira API to get all boards for each project and use the UIDs to populate
+   * the internal boards map.
+   * Only includes boards based on the inclusion criteria determined by getBoardInclusion.
+   * With this method we generate a unique board per project to represent all tasks in the project
+   *
+   * @returns A Promise that resolves when all boards have been processed.
+   */
+  private async getBoardsFromProjects(): Promise<void> {
+    for (const project of this.projects) {
+      const {included, issueSync} = await this.getBoardInclusion(project);
```
chalenge, line 231: are we handling if someone provides a list of board IDs by mistake?

destinations/airbyte-faros-destination/src/converters/jira/faros_issues.ts
```diff
@@ -129,6 +129,16 @@ export class FarosIssues extends JiraConverter {
       },
     });
 
+    if (!this.useBoardOwnership(ctx)) {
```
chalenge, line 132: Is this the only place this will be used? Can we rename the flag to something more meaningful like `use_project_as_board`

chalenge (changes requested):
How are sprint report boards being handled, are they all gonna be phantoms. Can we add some tests to validate these mappings

sources/jira-source/src/streams/faros_boards.ts
```diff
@@ -24,6 +24,22 @@ export class FarosBoards extends StreamWithProjectSlices {
...
+          uid: projectKey,
+          name: `Tasks in project ${projectKey}`,
+          projectKey,
+          type: 'Custom',
+          issueSync,
+        };
+      }
+      return;
+    }
+
```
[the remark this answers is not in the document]
matiaslcoulougian: The dummy board bellow would be used in the case where use board ownership is true... For this special case we only yield one board per project and return

destinations/airbyte-faros-destination/resources/source-specific-configs/jira.json
```diff
@@ -11,11 +11,12 @@
           "const": "Jira",
           "order": 0
         },
-        "use_board_ownership": {
+        "use_projects_as_boards": {
```
chalenge, line 14: Won't this break the feeds until they migrate?

destinations/airbyte-faros-destination/src/converters/jira/common.ts
```diff
@@ -35,7 +35,7 @@ export const JiraStatusCategories: ReadonlyMap<string, string> = new Map(
 export interface JiraConfig {
   additional_fields_array_limit?: number;
   exclude_fields?: string[];
-  use_board_ownership?: boolean;
+  use_projects_as_boards?: boolean;
```
chalenge, line 38: see above

sources/jira-source/src/project-board-filter.ts
```diff
@@ -149,13 +149,17 @@ export class ProjectBoardFilter {
       // Ensure projects is populated
       await this.getProjects();
 
-      // Ensure included / excluded boards are loaded
-      await this.loadSelectedBoards();
-
-      if (this.isWebhookSupplementMode() && this.hasFarosClient()) {
-        await this.getBoardsFromFaros(jira);
+      if (this.useProjectsAsBoards()) {
```
chalenge, line 152: I am not sure I still follow the logic in this filter. Let's split this up in another PR

sources/jira-source/src/project-board-filter.ts
```diff
@@ -239,6 +243,20 @@ export class ProjectBoardFilter {
     }
   }
 
+  /**
+   * Use the Jira API to get all boards for each project and use the UIDs to populate
```
chalenge, line 247: This comment doesn't seem correct. I don't see us using the API, but only returning the list of projects

sources/jira-source/src/project-board-filter.ts
```diff
@@ -149,13 +149,17 @@ export class ProjectBoardFilter {
...
-      await this.loadSelectedBoards();
-
-      if (this.isWebhookSupplementMode() && this.hasFarosClient()) {
-        await this.getBoardsFromFaros(jira);
+      if (this.useProjectsAsBoards()) {
+        await this.getBoardsFromProjects();
       } else {
-        await this.getBoardsFromJira(jira);
+        // Ensure included / excluded boards are loaded
+        await this.loadSelectedBoards();
```
chalenge, line 156: does this still work when using faros_graph for board selection?

chalenge (commented):
See comments. Can we validate feeds works with current destination

sources/jira-source/src/jira.ts
```diff
@@ -978,6 +979,8 @@ export class Jira {
     return sprints;
   }
 
+  // add a get sprint and sprint report by project or smth like that
```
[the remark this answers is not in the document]
matiaslcoulougian: Nope, they are old comments I left by mistake.

sources/jira-source/src/project-board-filter.ts
matiaslcoulougian: It was not working... I added the support for it + handling the case were use_faros_graph_boards_selection is false and boards/excludedBoards configs are set by mistake.

sources/jira-source/src/project-board-filter.ts
```diff
@@ -6,121 +6,155 @@ import {Memoize} from 'typescript-memoize';
...
     }
+    return excludedItems;
+  }
 
+  // Verifies that inclusion and exclusion lists are not specified when using Faros Graph for selection, and
+  // returns the lists to use.
+  protected verifyListsWithGraphSelection(
+    items: ReadonlyArray<string>,
+    excludedItems: ReadonlyArray<string>,
+    key: 'boards' | 'projects' = 'boards'
```
chalenge, line 104: when do we ever fetch projects? We only fetch selected boards from the graph

sources/jira-source/src/project-filter.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  constructor(
+    config: JiraConfig,
+    logger: AirbyteLogger,
+    farosClient?: FarosClient
+  ) {
+    super(config, logger, farosClient);
+
+    const {items: projects, excludedItems: excluded_projects} =
```
chalenge, line 36: this feels a little more complicated and we can simplify this. I am thinking when 
1. `projects_as_boards` is selected, ignore board list completely. filterConfig.projects = projects.
2. `projects_as_boards` is selected and `use_faros_graph`, ignore initial projects config too. then filterConfig.projects === boards come from the graph via selected boards
WYDT?

sources/jira-source/src/project-board-filter.ts
```diff
@@ -6,121 +6,155 @@ import {Memoize} from 'typescript-memoize';
...
     if (!this.projects) {
       this.logger.info('Generating list of projects to sync');
-      this.projects = new Set();
+      this.projects = new Map();
 
       const jira = await Jira.instance(this.config, this.logger);
       if (!this.filterConfig.projects?.size) {
         const projects =
-          this.isWebhookSupplementMode() && this.hasFarosClient()
+          this.isWebhookSupplementMode && this.hasFarosClient()
```
chalenge, line 135: we should prolly move this out too in next PR

sources/jira-source/src/project-filter.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  @Memoize()
+  async getProjects(): Promise<ReadonlyArray<ProjectOrBoardInclusion>> {
+    if (!this.projects) {
+      this.logger.info('Generating list of projects to sync');
+      this.projects = new Map();
+
+      const jira = await Jira.instance(this.config, this.logger);
+
+      if (!this.filterConfig.projects?.size) {
```
chalenge, line 59: looks awry similar to the other method. Okay for now but should look to generalize

destinations/airbyte-faros-destination/test/converters/__snapshots__/jira.test.ts.snap
```diff
@@ -4,8 +4,8 @@ exports[`jira process records from all streams 1`] = `
...
-  "Would write 72 records",
-  "Would write records by model: {\\\\\\"tms_Epic\\\\\\":1,\\\\\\"tms_Project\\\\\\":1,\\\\\\"tms_ProjectReleaseRelationship\\\\\\":3,\\\\\\"tms_Release\\\\\\":3,\\\\\\"tms_Sprint\\\\\\":10,\\\\\\"tms_Task\\\\\\":5,\\\\\\"tms_TaskAssignment\\\\\\":1,\\\\\\"tms_TaskBoard\\\\\\":1,\\\\\\"tms_TaskBoardProjectRelationship\\\\\\":1,\\\\\\"tms_TaskBoardRelationship\\\\\\":5,\\\\\\"tms_TaskDependency\\\\\\":1,\\\\\\"tms_TaskProjectRelationship\\\\\\":5,\\\\\\"tms_TaskReleaseRelationship__Deletion\\\\\\":1,\\\\\\"tms_TaskReleaseRelationship__Upsert\\\\\\":3,\\\\\\"tms_TaskTag\\\\\\":2,\\\\\\"tms_User\\\\\\":29}\\"},\\"type\\":\\"LOG\\"}",
+  "Would write 71 records",
```
chalenge, line 7: why less?

chalenge (approved):
Looks good, see comments.

sources/jira-source/src/project-board-filter.ts
matiaslcoulougian: Did this change because if we leave the RunMode enum import from comon.ts as it was before, we end up with a circular dependency between imports: project-board-filter -> common -> project-filter -> project-board-filter.

sources/jira-source/src/project-board-filter.ts
matiaslcoulougian: We don't fetch projects. But we fetch boards that are actually projects and that we need to verify against config project list and use as project filter config, when use_projects_as_boards is enabled.

sources/jira-source/src/project-filter.ts
matiaslcoulougian: So the check inside the constructor would be if `use_faros_graph` is enabled, ignore projects and excluded_projects config. 
And if `use_faros_graph` is not enabled, we need to check that only projects or excluded_projects are defined and not both.

That is basically what is being done inside [verifyListsWithGraphSelection](https://github.com/faros-ai/airbyte-connectors/pull/1908/files#diff-51f5e440a3db84efa710db1709f27c37890c8a47e3d71f52703594ec45560d6dR101) function

PR #1916, opened 2025-02-18
FAI-14966: Support fetching users in Server Instance Types

chalenge:
## Description

- Support fetching users in Server Instance Types using Teams and Team Memberships

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


sources/azure-workitems-source/src/azure-workitems.ts
```diff
@@ -69,9 +76,16 @@ export class AzureWorkitems {
     if (AzureWorkitems.azure_Workitems) return AzureWorkitems.azure_Workitems;
 
     AzureWorkitems.validateConfig(config);
+    if (config.instance_type?.type) {
```
dbruno21, line 79: Should we add to azure-repos the same logs?

dbruno21 (approved):

sources/azure-workitems-source/src/azure-workitems.ts
chalenge: It is not really using that logic. We need to re-write all this to have common functions 😭

PR #1918, opened 2025-02-19
FAI-15182 Address memory usage issues in circleci source

ypc-faros:
## Description

In the circleci source, we [cache](https://github.com/faros-ai/airbyte-connectors/blob/352c2f37d53592c73e8cd0900e898511c2cba7a9/sources/circleci-source/src/circleci/circleci.ts#L345) the workflows for (pipelineId, since). This cached data is only used if the `since` lines up between streams (pipelines and tests). Otherwise we’re wasting memory without any benefit.

I think we should:

(This PR)
Remove the memoization from the source to try to resolve the OOM.

(Next steps)
Fix the caching issues by: fusing the pipelines and tests streams OR changing the logic to improve the odds of a cache hit.

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/circleci-source/src/circleci/circleci.ts
+++ b/sources/circleci-source/src/circleci/circleci.ts
@@ -341,7 +341,6 @@ export class CircleCI {
     }
   }
 
-  @Memoize()
   async fetchWorkflows(pipelineId: string, since: Date): Promise<Workflow[]> {
     const url = `/pipeline/${pipelineId}/workflow`;
     const results: Workflow[] = [];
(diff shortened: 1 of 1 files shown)


sources/circleci-source/src/circleci/circleci.ts
```diff
@@ -341,7 +341,6 @@ export class CircleCI {
     }
   }
 
-  @Memoize()
```
chalenge, line 344: what is the impact on the API. Will it get hit much more?

ypc-faros: It's already being hit (if I'm correct, the memoization is most likely not helping at all)

But for sure we should make it work.

PR #1919, opened 2025-02-19
FAI-15197 | Handle possible null requested reviewer on GitHub pull requests

dbruno21:
## Description

Handle possible null requested reviewer on GitHub pull requests
Based on [GitHub gql schema](https://docs.github.com/en/graphql/reference/objects#reviewrequest), requestedReviewer is nullable and we experienced null read errors because of this. Haven't found a reason, documentation or proof on why this would happen (maybe the referenced user was deleted from the GitHub instance?)

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


chalenge (approved):
Any other user objects we should check, like commentor, author etc?

dbruno21:
> Any other user objects we should check, like commentor, author etc?

Looks like we already have in place checks for those

PR #1922, opened 2025-02-21
FAI-14966: Azure Workitems use supported node client

chalenge:
## Description

- Implement using https://github.com/microsoft/azure-devops-node-api client instead of axios

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


chalenge:
@cjwooo @dbruno21 since you're now the AzDo experts

dbruno21 (approved):

dbruno21:
@chalenge I assume there would be a separate PR for converters?

chalenge:
> @chalenge I assume there would be a separate PR for converters?

Yeah. The PR was getting too large. I will do the converters in a separate one when I am done with the sources.

PR #1911, opened 2025-02-05
Make it possible to store compact representation of input record in ctx for dependent streams

ypc-faros:
## Description

This helps with memory usage. For example, for Wolken, the [dependent stream](https://github.com/faros-ai/airbyte-connectors/blob/0d3c38c8f0e7c287d26d434e9882297e9e623c4a/destinations/airbyte-faros-destination/src/converters/wolken/incidents.ts#L133) only really needs the `serviceIdFlexField` from the input `ConfigurationItem`

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

> Describe migration notes if any

## Extra info

> Add any additional information

PR #1924, opened 2025-02-25
FAI-15233 | Wolken support for user lookup extra fields

dbruno21:
## Description

- Allow specifying (dst) `application_tag_flex_field_user_lookup_names`. Same use as `application_tag_flex_field_names` but values are also treated as user references (can do a lookup on users records)
  - With (dst) `user_lookup_extra_fields_mapping` fields from the user lookup can be written as additional tags
  - With (src) `flex_field_user_lookup_names` users refs found on flex fields with that name will be retrieved with their full details (invididual call per each)
- Write incident category and subcategory as incident tags

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


destinations/airbyte-faros-destination/src/converters/wolken/configuration_items.ts
```diff
@@ -1,10 +1,23 @@
...
+  DestinationRecord,
+  StreamContext,
+  StreamName,
+} from '../converter';
 import {WolkenConverter} from './common';
 
 export class ConfigurationItems extends WolkenConverter {
+  static readonly usersStream = new StreamName('wolken', 'users');
+
+  override get dependencies(): ReadonlyArray<StreamName> {
```
chalenge, line 17: Not sure how many these users are and how big the records but can we store compact reps of them [like here](https://github.com/faros-ai/airbyte-connectors/pull/1911)

destinations/airbyte-faros-destination/src/converters/wolken/configuration_items.ts
```diff
@@ -85,6 +98,45 @@ export class ConfigurationItems extends WolkenConverter {
         }
       }
 
+      const applicationTagFlexFieldUserLookupNames =
```
chalenge, line 101: can you extract to separate method to split up this already large method?

chalenge (approved):
See comments

destinations/airbyte-faros-destination/src/converters/wolken/configuration_items.ts
dbruno21: It's there -> `destinations/airbyte-faros-destination/src/converters/wolken/users.ts`

PR #1926, opened 2025-02-26
FAI-14652: Datadog source and converter add SLOs

chalenge:
## Description

- Add ability to pull SLOs from datadog source
- Migrate datadog client and ensure type correctness in source and destination

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/test/converters/datadog.test.ts
```diff
@@ -41,28 +31,7 @@ describe('datadog', () => {
       configPath,
       catalogPath: 'test/resources/datadog/catalog.json',
       inputRecordsPath: 'datadog/all-streams.log',
+      checkRecordsData: (records) => expect(records).toMatchSnapshot(),
     });
-
-    const recordedRequests = await segmentMock.getSeenRequests();
```
chalenge, line 46: removed this for actual data check with snapshot above hence the massive diff size

destinations/airbyte-faros-destination/src/converters/datadog/common.ts
```diff
@@ -39,4 +48,66 @@ export abstract class DatadogConverter extends Converter {
...
+      const colonIndex = tag.indexOf(':');
+      const [key, value] =
+        colonIndex === -1
+          ? [tag, null]
+          : [tag.substring(0, colonIndex), tag.substring(colonIndex + 1)];
+
+      records.push({
+        model: 'faros_Tag',
+        record: {
+          uid: value ? `${key}:${value}` : key,
```
willmarks, line 101: Depending on how these tags are being associated, sometimes its nice to have a piece of the uid be the uid for the object being associated so that if that object is updated then it replaces the existing tag. It looks like the colonIndex stuff makes sure every tag's key is unique tho?

destinations/airbyte-faros-destination/src/converters/datadog/service_level_objectives.ts
```diff
@@ -0,0 +1,140 @@
...
+        return {category: 'Custom', detail: String(state._data)};
+    }
+  }
+
+  /**
+   * Service Level Objective timestamps are in Unix time (seconds). Convert to milliseconds.
+   * @param number - The Unix timestamp
+   * @returns The Date object
+   */
+  private toDate(number?: number): Date | undefined {
```
willmarks, line 134: Maybe `unixSecondsToDate` or something. We have a bunch of `toDate` functions floating around..

destinations/airbyte-faros-destination/src/converters/datadog/service_level_objectives.ts
```diff
@@ -0,0 +1,140 @@
...
+        return {category: 'Breached', detail: state};
+      case 'warning':
+        return {category: 'Warning', detail: state};
+      case 'ok':
+        return {category: 'Healthy', detail: state};
+      case 'no_data':
+        return {category: 'Pending', detail: state};
+      default:
+        // Handled for UnparsedObject in SLOState type
+        return {category: 'Custom', detail: String(state._data)};
```
willmarks, line 125: is `_data` a typescript thing?

willmarks (approved):

destinations/airbyte-faros-destination/src/converters/datadog/common.ts
chalenge: Do you mean the uid for example if the tag is `env:prod` for an incident then make the uid of the faros_Tag be something like `incident_env:prod`?

destinations/airbyte-faros-destination/src/converters/datadog/service_level_objectives.ts
chalenge: it's a util class for the datadog client to handle any unparsed objects not matching values for the type. Not sure where it is used

destinations/airbyte-faros-destination/src/converters/datadog/common.ts
willmarks: yeah for example an application tag's uid might be `<application>_<tag_key>` so any changes to the value update the tag for that application. This assumes you only want a single instance of a given tag per tagged entity. Not exactly sure if that fits this situation tho.

chalenge: Not really, think we want the same tag to be associated with all entities. For example the same tag for a service can be used on an SLO can also be on an incident

PR #1938, opened 2025-03-11
FAI-15518: Migrate Azure Repos source to use shared node client

chalenge:
## Description

- Migrate Azure Repos source to use shared node client

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


dbruno21 (approved):
Starting to look good!

PR #1943, opened 2025-03-14
[FAI-13889] Update placeholders for airbyte cli wizard autofill

jeniii:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change


faros-airbyte-cdk/src/prompts.ts
```diff
@@ -39,9 +45,24 @@ export function runSelect(cfg: SelectConfig): Promise<string> {
...
+     * If the config is a leaf, skip EXAMPLE and ENUM.
+     * That means we don't autofill EXAMPLE and ENUM values as that might get confused with DEFAULT values.
+     */
     for (const symbol of Object.values(ChoiceType)) {
-      if (choices.has(symbol)) {
+      if (
+        choices.has(symbol) &&
+        !(
+          cfg.name === SelectConfigName.LEAF &&
+          (symbol === ChoiceType.EXAMPLE || symbol === ChoiceType.ENUM)
```
jeniii, line 63: @cjwooo  i want to not autofill examples and enums when there's no default for the configs. i think it will be confused with the defaults.

If the config doesn't have default, then it will fall back to placeholders. for `PLACEHOLDER_NUMBER` and `PLACEHOLDER_BOOLEAN`, it might still be confusing but i don't know if there a better way to get around that.

destinations/airbyte-faros-destination/resources/spec.json
```diff
@@ -24,10 +24,10 @@
         "type": "object",
         "oneOf": [
           {
-            "$ref": "community.json"
+            "$ref": "cloud.json"
```
jeniii, line 27: the default is edition is `cloud`. i make it list first so it can be picked up as default for wizard autofill. 
I'm not sure how to set a default in oneOf. do you know? @ypc-faros

PR #1949, opened 2025-03-20
FAI-15519: Fix generic pagination for continuation token

chalenge:
## Description

- Fix generic pagination for continuation token

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


faros-airbyte-common/src/azure-devops/azure-devops.ts
```diff
@@ -94,7 +94,14 @@ export abstract class AzureDevOps {
...
 
-    const top = config.page_size ?? DEFAULT_PAGE_SIZE;
+    const pageSize = isNaN(config.page_size) ? undefined : config.page_size;
+    if (pageSize && pageSize <= 1) {
+      logger.warn(
+        'Page size must be greater than 1, will use default minimum of 2'
+      );
+    }
+
+    const top = Math.max(2, pageSize ?? DEFAULT_PAGE_SIZE);
```
willmarks, line 104: Should we enforce the minimum at the spec level?

faros-airbyte-common/src/azure-devops/azure-devops.ts
```diff
@@ -202,40 +226,53 @@ export abstract class AzureDevOps {
...
-        // continuationTokenParam as the continuation token
-        const lastRecord = result[result.length - 1];
-        continuationToken = lastRecord[continuationTokenParam];
-      } else {
-        continuationToken = undefined;
+      if (!Array.isArray(result)) {
+        throw new VError(
+          `Expected array result but got ${typeof result} ${JSON.stringify(
+            result
+          )}`
```
willmarks, line 254: Any danger of leaking creds?

willmarks (approved):

faros-airbyte-common/src/azure-devops/azure-devops.ts
```diff
@@ -202,40 +226,53 @@ export abstract class AzureDevOps {
...
+        );
       }
+
+      // Remove the first record when we have a continuation token to avoid duplication
+      // since we already processed it in the previous page
+      const records = continuationToken ? result.slice(1) : result;
```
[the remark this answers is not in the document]
chalenge: This is a workaround for [issue linked](https://github.com/microsoft/azure-devops-node-api/issues/609) on the method docs. Normally the continuationToken would be the first record in next page but the client is not returning the continuation token in the response, so doing this inefficient hack of using last record in current page, which will then be first record in next page hence why I have to remove it. 

Other option would be to call REST endpoint directly for these calls.

faros-airbyte-common/src/azure-devops/azure-devops.ts
chalenge: Don't think so.

PR #1956, opened 2025-03-24
FAI-15701: Azure pipelines use node client types

chalenge:
## Description

Deprecate manual types and use types from https://github.com/microsoft/azure-devops-node-api

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/azurepipeline/common.ts
```diff
@@ -35,9 +27,6 @@ export abstract class AzurePipelineConverter extends Converter {
   getOrganizationFromUrl(url: string): string | undefined {
     try {
       const parsed = new URL(url);
-      if (parsed.hostname !== 'dev.azure.com') {
```
chalenge, line 38: breaks when the hostname is not cloud is results

dbruno21 (approved):

PR #1967, opened 2025-04-01
FAI-15474 | Copilot metrics support fixed teams config

dbruno21:
## Description

Copilot metrics support fixed teams config

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/github-source/resources/spec.json
+++ b/sources/github-source/resources/spec.json
@@ -352,6 +352,24 @@
         "description": "Fetch public organizations",
         "default": false,
         "order": 34
+      },
+      "copilot_metrics_teams": {
+        "type": "array",
+        "title": "Copilot Metrics Teams",
+        "description": "List of teams to fetch Copilot metrics for. If not specified, all teams for the organization will be fetched.",
+        "items": {
+          "type": "string"
+        },
+        "order": 35
+      },
+      "copilot_metrics_enterprise_teams": {
+        "type": "array",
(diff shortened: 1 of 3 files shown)


sources/github-source/resources/spec.json
```diff
@@ -352,6 +352,24 @@
         "description": "Fetch public organizations",
         "default": false,
         "order": 34
+      },
+      "copilot_metrics_teams": {
+        "type": "array",
+        "title": "Copilot Metrics Teams",
+        "description": "List of teams to fetch Copilot metrics for. If not specified, all teams for the organization will be fetched.",
+        "items": {
```
chalenge, line 360: would prefer to have one param and handle not finding the team in either copilot or enterprise copilot with `No GitHub Copilot usage found for org ${org} - team ${teamSlug}.` when we are fetching from both.

chalenge (approved):
lgtm

sources/github-source/resources/spec.json
dbruno21: Updated it to be just 1 param, I think it makes sense to set up multiple sources if needed both use cases, after all even with keeping separate configs, we would also have a similar problem if we consider the fact that multiple orgs / enterprises could be fetched as well (and each of them with different teams)

PR #1982, opened 2025-04-10
Vuln: Upgrade axios to 1.8.4

omreego:
## Description
Vuln fix

### DIFF
--- a/destinations/airbyte-faros-destination/package.json
+++ b/destinations/airbyte-faros-destination/package.json
@@ -34,7 +34,7 @@
   "dependencies": {
     "@esm2cjs/p-queue": "^7.3.0",
     "@segment/analytics-node": "^2.0.0",
-    "axios": "^1.7.4",
+    "axios": "^1.8.4",
     "date-format": "^4.0.6",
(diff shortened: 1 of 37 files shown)


cjwooo (approved):

destinations/airbyte-faros-destination/package.json
```diff
@@ -34,7 +34,7 @@
   "dependencies": {
     "@esm2cjs/p-queue": "^7.3.0",
     "@segment/analytics-node": "^2.0.0",
-    "axios": "^1.7.4",
+    "axios": "^1.8.4",
```
chalenge, line 37: I wonder if we should just at this point add it to the [main package](https://github.com/faros-ai/airbyte-connectors/blob/main/package.json) and deal with it in only one place

omreego: Do the sub-packages override the main one if it exists in both?

chalenge: Yeah, module can override as necessary. Btw we have `faros-js-client` in main package which imports `axios`, so think we can remove these too 😬

PR #1985, opened 2025-04-11
FAI-15881: Upgrade to node 22

chalenge:
## Description

- Upgrade to node 22
- Removed engine from sub-packages and moved to main

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "root",
   "private": true,
+  "packageManager": "^npm@10.9.2",
   "engines": {
     "node": ">=22"
   },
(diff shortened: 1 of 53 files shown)


package.json
```diff
@@ -1,7 +1,10 @@
 {
   "name": "root",
   "private": true,
-  "packageManager": "^npm@9.6.0",
+  "packageManager": "^npm@10.9.2",
```
dbruno21, line 4: Can you remove `^` here? I found on another repo that the reason dependabot PRs don't update package-lock.json is because of this

dbruno21 (approved):

package.json
chalenge: Really, it's never been an issue here. But can remove. We lose package-manager enforcement though.

PR #1994, opened 2025-04-15
FAI-16144 | Workday allows to add Employee Tag for Vendor

omreego:
## Description
Adds org_EmployeeTag matching employee to vendor - if vendor is provided by Workday.
Can potentially filter them by an input or by some other method-

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


meldiner (approved):
will this new tag show up in the reports db, or do we need to update dbt?

omreego:
I believe it will already appear in the Org Table under the column Descendant Faros Tags Vendor

chalenge (approved):

PR #1998, opened 2025-04-17
FAI-16059: Azurepipelines fetch runs instead of builds

chalenge:
## Description

- List builds returns only classic pipelines/ build definitions. Switch to using pipelines runs which returns everything

## Type of change
- [X] Bug fix
- [X] New feature
- [ ] Breaking change


sources/azurepipeline-source/src/azurepipeline.ts
```diff
@@ -51,78 +57,125 @@ export class AzurePipelines extends AzureDevOps {
...
   }
 
-  async *getPipelines(project: ProjectReference): AsyncGenerator<Pipeline> {
-    const pipelines = await this.client.pipelines.listPipelines(project.id);
-    for (const pipeline of pipelines) {
-      yield {
+  @Memoize((project: ProjectReference) => project.id)
+  async getPipelines(project: ProjectReference): Promise<Pipeline[]> {
+    const pipelines = [];
+    const result = await this.client.pipelines.listPipelines(project.id);
```
dbruno21, line 63: Isn't pagination needed or is it handled somewhere else?

sources/azurepipeline-source/src/azurepipeline.ts
```diff
@@ -51,78 +57,125 @@ export class AzurePipelines extends AzureDevOps {
...
-      };
+      });
     }
+    return pipelines;
   }
 
-  // https://docs.microsoft.com/en-us/rest/api/azure/devops/build/builds
-  async *getBuilds(
+  // https://learn.microsoft.com/en-us/rest/api/azure/devops/pipelines/runs/list
+  async *getRuns(
```
dbruno21, line 74: Do any of the API calls inside need pagination handling or is it already handled somewhere else?

dbruno21 (approved):
see comments

sources/azurepipeline-source/src/azurepipeline.ts
chalenge: This specific API doesn't support any pagination and returns top 10000 runs for a particular pipeline. Updated the method docs

sources/azurepipeline-source/src/azurepipeline.ts
chalenge: Not paginating coz the client doesn't return the continuation token and can fetch all pipelines in one call. If we start hitting a true limit of results in one call then we will have to switch to directly using REST api.

Updated method docs

PR #2000, opened 2025-04-23
Enforce using LF

jeniii:
## Description

> Provide description here

Due to the different EOF on Windows, users are encountering issues when building/running the docker images on their local machines. Enforce to use LF for all files.

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
jeniii, line 1: this should enforce all files to be LF and doesn't require users to enforce it in git config.

jeniii: lmk if you foresee issues using all LF on Windows

cjwooo (approved):

PR #2009, opened 2025-05-02
FAI-15916: Fix Azure boards get user and state

chalenge:
## Description

- Fix Azure boards get user
- Fix setting state when it is initially null
- Handle when get workitem updates returns nothing

## Type of change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change


sources/azure-workitems-source/src/azure-workitems.ts
```diff
@@ -194,6 +194,11 @@ export class AzureWorkitems extends types.AzureDevOps {
     for await (const update of this.getPaginated(getUpdatesFn)) {
       updates.push(update);
     }
+    if (!updates.length) {
+      this.logger.warn(
+        `Failed to get updates found for work item ${id} in project ${project}`
```
ypc-faros, line 199: nit: Remove `found` ?

ypc-faros (approved):
Fix failing test

sources/azure-workitems-source/src/azure-workitems.ts
chalenge: damn GenAI

PR #2012, opened 2025-05-06
FAI-16457 - Faros org import converters

willmarks:
## Description

Creates converters that can import org data into Faros. Functions very similarly to how the existing org importers work with some streamlining in order to not need cross stream dependencies.

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


destinations/airbyte-faros-destination/src/converters/faros-org-import/employees.ts
```diff
@@ -0,0 +1,262 @@
...
+      ctx?.config?.source_specific_configs?.faros_org_import?.resolve_locations,
+      ctx.farosClient
+    );
+  }
+
+  async convert(
+    record: AirbyteRecord,
+    ctx?: StreamContext
+  ): Promise<ReadonlyArray<DestinationRecord>> {
+    ctx.logger?.debug(JSON.stringify(record));
```
ypc-faros, line 52: nit: This will log every single input record when in debug mode. Maybe that's too much?

ypc-faros (approved):

PR #2038, opened 2025-05-22
FAI-16665 | New stream for GitHub Copilot User Engagement API (src + dst)

dbruno21:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


crupley (approved):

sources/github-source/src/github.ts
```diff
@@ -2065,6 +2069,54 @@ export abstract class GitHub {
...
+      0,
+      ...data.map((record) => Utils.toDate(record.date).getTime())
+    );
+    if (latestDay <= cutoffDate) {
+      this.logger.info(
+        `GitHub Copilot user engagement metrics for enterprise ${enterprise} is already up-to-date: ${new Date(cutoffDate).toISOString()}`
+      );
+      return;
+    }
+    const axios = makeAxiosInstanceWithRetry({
```
chalenge, line 2100: is this not yet available in octokit?

chalenge (approved):
🧑‍🍳

sources/github-source/src/github.ts
dbruno21: At first I have it working using the octokit instance, but if you take a look at the only example we have (test resources), the blob_uri doesn't seem to be autententicated and outside of GitHub's domain, so better not to send GitHub auth headers if not needed.

PR #2067, opened 2025-06-03
[FAI-16838] - Update github copilot to use CodingAssistant tool type

crupley:
## Description

Change github copilot user tool from

```js
{category: 'GitHubCopilot', detail: null}
```

to

```js
{category: 'CodingAssistant', detail: 'GitHubCopilot'}
```

First time working on this connector; I was unsure on a few things. Left comments to highlight.

## Type of change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Related issues

https://github.com/faros-ai/canonical-reports/pull/1082

## Migration notes

Could break customer charts. See here for examples to fix:

https://github.com/faros-ai/aion/pull/1789


destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
crupley, line 1: I updated this by deleting it and then running the test to regenerate it; is that correct?

faros-airbyte-common/src/github/types.ts
```diff
@@ -250,8 +250,12 @@ export type CopilotSeatsEmpty = {
...
 };
 
-export enum GitHubTool {
-  Copilot = 'GitHubCopilot',
+export enum ToolCategory {
+  CodingAssistant = 'CodingAssistant',
+}
+
+export enum ToolDetail {
+  GitHubCopilot = 'GitHubCopilot',
```
crupley, line 258: can eventually use this for Cursor, Windsurf, etc. connectors

destinations/airbyte-faros-destination/test/converters/__snapshots__/faros_github.test.ts.snap
ypc-faros: You can just run `npm run test`. That will fail if there's a diff in the snapshot and it'll show you the diff. If you're okay with the diff you can update the snapshots with `npm run test -- -u`

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
```diff
@@ -222,7 +223,7 @@ function userToolKey(
   return {
     user: {uid: userLogin, source},
     organization: {uid: toLower(orgLogin), source},
-    tool: {category: GitHubTool.Copilot},
+    tool: {category: ToolCategory.CodingAssistant, detail: ToolDetail.GitHubCopilot},
```
dbruno21, line 226: @crupley I realized that this change would have a very bad outcome. The tool field is used as part of vcs_UserTool key. So records generated with a different tool value, would be treated as different / new vcs_UserTool records.

How bad would it be to not touch the connector and just handle the transform on the reports layer for this specific case of GitHubCopilot ? We just need to remember for any new connectors or data ingestion of other tools to use the new format to avoid having to handle them as well on the reports.

dbruno21 (commented):
see comment

destinations/airbyte-faros-destination/src/converters/github/faros_copilot_seats.ts
crupley: My thinking here was that having 2 sets of tools wouldn't be that much of a problem. There will be a point in time where everyone switches to the new set, but there will be no difference in the reports data. I think it's better for the graph data and the reports data to agree and they would as soon as this change goes out.

Other than just having a duplicate of the UserTool in the graph, do you see other problems or confusion coming from this change? Could is mess up how we pull other related things like licenses?

dbruno21: > There will be a point in time where everyone switches to the new set
Old records already in the graph won't change. We will be creating new records (as if it was a different tool). I suppose that the reports can handle that "duplication" if not already handled, but I think that when we have to see the data on the graph (e.g. for debugging) it will be very confusing.

> do you see other problems or confusion coming from this change?
It also adds complexity here for the logic we do to query active seats and "infer" from the latest API response what seats should set to inactive. We would need to query both old records and new records.

Besides the added confusion and complexity this change could add, I don't see any benefit since we're already forced to live with the old records format that are already in the graph.

dbruno21: There could be a destination config, but it would be a manual thing and no one will remember or understand what it is to set it to the correct value when creating a new connector instance. Even then, I still think we just need to keep 1 format, and since we are forced to keep old records we are tight to the old format. 

@chalenge @ypc-faros thoughts?

ypc-faros: Yeah I think it's not worth the extra complexity.

Either we commit to transform the existing records and do the code cleanup afterwards (full migration) or keep it as is.

crupley:
https://github.com/faros-ai/airbyte-connectors/pull/2067#discussion_r2127034091

Decided against this change

PR #2082, opened 2025-06-10
Implement GitLab UserCollector for enhanced user deduplication and commit author resolution

ypc-faros:
## Summary
Introduces comprehensive user management improvements to the GitLab connector by implementing a UserCollector pattern and simplifying destination-side processing.

### Source-side improvements (gitlab-source):
- **UserCollector class**: New centralized user collection, deduplication, and merging logic
- **Enhanced user processing**: Collect users from all groups with comprehensive data merging
- **Commit author resolution**: Map commit author display names to usernames via UserCollector  
- **Stream dependencies**: Ensure users are collected before commits for proper author resolution
- **Updated User type**: Add `group_id` field for better organization mapping
- **Simplified users stream**: Direct user collection and yielding instead of complex slice processing

### Destination-side simplification (airbyte-faros-destination):
- **Simplified faros_users converter**: Remove redundant deduplication logic since it's now handled at source
- **Direct conversion**: Convert each user record immediately instead of batch processing
- **Leverage group_id**: Use new `group_id` field instead of parsing group information
- **Remove onProcessingComplete**: No longer needed since deduplication happens at source

## Benefits
- ✅ Eliminates duplicate deduplication logic between source and destination
- ✅ Improved commit author attribution through username resolution  
- ✅ More efficient memory usage at destination
- ✅ Better user data integrity through source-side merging
- ✅ Cleaner, more maintainable converter logic
- ✅ Comprehensive test coverage for new UserCollector functionality

## Test Plan
- [x] All existing GitLab source tests pass
- [x] All GitLab destination converter tests pass  
- [x] New UserCollector unit tests with 12 test cases covering:
  - User collection and deduplication
  - Data merging with first-non-empty-wins strategy
  - Commit author resolution with conflict handling
  - Edge cases and error scenarios
- [x] Stream dependency tests ensuring proper execution order
- [x] Updated test snapshots to reflect new data format


dbruno21 (approved):
What is/are the benefit/s of doing the user lookup in the source instead of destination? Should we have done the same for GitHub?

destinations/airbyte-faros-destination/src/converters/gitlab/faros_users.ts
```diff
@@ -15,14 +16,55 @@ export class FarosUsers extends GitlabConverter {
...
+          name: user.name,
+          email: user.email,
+          htmlUrl: user.web_url,
+          type: {category: 'User', detail: 'user'},
+        },
+        (value) => isNil(value) || isEmpty(value)
+      ),
+    });
+
+    return res;
```
dbruno21, line 68: Wouldn't it be better to continue using collectUser + convertUsers, even if we already know there won't be duplicates, so that we don't have to duplicate the mappings? If we make a change we need to modify it on both places.

dbruno21 (commented):
I understand that for other converters we still continue collecting users for dedupe + convert, is that right?

destinations/airbyte-faros-destination/src/converters/gitlab/faros_users.ts
ypc-faros: That's a good point. We'd be trading some memory usage for maintainability. I'm 99% for maintainability

PR #2101, opened 2025-06-21
ypc/async

ypc-faros:
- **async**
- **async**

## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info

### DIFF
--- a/sources/gitlab-source/src/gitlab.ts
+++ b/sources/gitlab-source/src/gitlab.ts
@@ -54,7 +54,7 @@ export class GitLab {
 
   constructor(
     readonly config: GitLabConfig,
-    protected readonly logger: AirbyteLogger,
+    protected readonly logger: AirbyteLogger
   ) {
     this.client = new GitlabClient({
       token: this.getToken(),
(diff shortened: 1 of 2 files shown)


sources/gitlab-source/src/gitlab.ts
```diff
@@ -160,43 +163,42 @@ export class GitLab {
     }
   }
 
-  async getProjects(groupId: string): Promise<FarosProjectOutput[]> {
-    const projects = (await this.keysetPagination(
+  async *getProjects(groupId: string): AsyncGenerator<FarosProjectOutput> {
```
chalenge, line 166: Don't we need to Memoize this for other streams?

chalenge (approved):

sources/gitlab-source/src/gitlab.ts
ypc-faros: We do that in the group filter:
https://github.com/faros-ai/airbyte-connectors/blob/34e771c07763fc22c00da277f70c29180beecadf/sources/gitlab-source/src/group-filter.ts#L116-L121
https://github.com/faros-ai/airbyte-connectors/blob/34e771c07763fc22c00da277f70c29180beecadf/sources/gitlab-source/src/streams/common.ts#L114-L125

PR #2168, opened 2025-07-10
FAI-17327: Jira add audit events stream to track deleted Jira issues

chalenge:
- Add sync_audit_events configuration option to enable audit log processing
- Implement FarosAuditEvents stream that fetches audit records via Jira API
- Filter audit events to only process ISSUE_DELETE events for synced projects
- Add comprehensive test coverage with mocked audit API responses
- Support both Jira Cloud and Server 7.6+ with graceful error handling
- Output IssueCompact format with isDeleted flag for destination processing

## Type of change
- [ ] Bug fix
- [X] New feature
- [ ] Breaking change


sources/jira-source/src/streams/faros_audit_events.ts
```diff
@@ -0,0 +1,108 @@
...
+  async *readRecords(
+    syncMode: SyncMode,
+    cursorField?: string[],
+    streamSlice?: any,
+    streamState?: AuditEventState
+  ): AsyncGenerator<any> {
+    const jira = await Jira.instance(this.config, this.logger);
+
+    // Get the projects that are being synced
+    const projects = await this.projectBoardFilter.getProjects();
```
dbruno21, line 60: Why don't we use project slices instead?

dbruno21 (approved):

sources/jira-source/src/streams/faros_audit_events.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    return currentStreamState;
+  }
+
+  async *readRecords(
+    syncMode: SyncMode,
+    cursorField?: string[],
+    streamSlice?: any,
+    streamState?: AuditEventState
+  ): AsyncGenerator<any> {
```
ypc-faros, line 56: nit: `any` => `AuditRecord`

sources/jira-source/src/streams/faros_audit_events.ts
```diff
@@ -0,0 +1,108 @@
...
+    currentStreamState: AuditEventState,
+    latestRecord: AuditRecord
+  ): AuditEventState {
+    const recordDate = latestRecord?.created;
+    if (!recordDate) {
+      return currentStreamState;
+    }
+
+    const currentCutoff = currentStreamState?.cutoff;
+    if (!currentCutoff || recordDate > currentCutoff) {
```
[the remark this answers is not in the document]
dbruno21: Also could store cutoff as number as we usually do

sources/jira-source/src/streams/faros_audit_events.ts
chalenge: Actually, did some testing and will use slices. Think for super large instances where we are fetching only a few projects it will be pretty wasteful and slow

PR #2171, opened 2025-07-11
FAI-17340 | Add fetch_issue_comments config to Jira source

dbruno21:
## Summary
- Add boolean config `fetch_issue_comments` with default `false`
- Include `comment` field in Jira API requests when enabled
- Update issue transformer to handle comment field mapping
- Add performance warning in config description


sources/jira-source/resources/spec.json
```diff
@@ -342,6 +342,13 @@
         "examples": [
           "Instance2"
         ]
+      },
+      "fetch_issue_comments": {
```
dbruno21, line 346: @chalenge Do you agree with this flag?

destinations/airbyte-faros-destination/src/converters/jira/faros_issues.ts
```diff
@@ -225,6 +225,30 @@ export class FarosIssues extends JiraConverter {
...
+        results.push({
+          model: 'tms_TaskComment',
+          record: {
+            task: {uid: issue.key, source},
+            uid: comment.id,
+            comment: Utils.cleanAndTruncate(
+              comment.body,
+              this.truncateLimit(ctx)
+            ),
+            createdAt: comment.created,
```
chalenge, line 239: Think you need `Utils.toDate(comment.created)`. Same below

destinations/airbyte-faros-destination/src/converters/jira/faros_issues.ts
```diff
@@ -225,6 +225,30 @@ export class FarosIssues extends JiraConverter {
       });
     }
 
+    if (issue.comments) {
```
chalenge, line 228: You need to do that delete comments then re-write them logic we do for labels

sources/jira-source/src/jira.ts
```diff
@@ -1009,6 +1013,10 @@ export class Jira {
         'summary',
         'updated',
       ].forEach((field) => fieldIds.add(field));
+
+      if (this.fetchIssueComments) {
+        fieldIds.add('comment');
```
chalenge, line 1018: Do we need to handle when there's more than 100 comments? Can you verify the API returns all comments. Saw something similar for something else when I migrated to new endpoint.

Also is this the same response for Jira server?

chalenge (commented):
See comments

PR #2178, opened 2025-07-16
FAI-17373: Default to startDate if activatedDate is null for openedAt field

matiaslcoulougian:
## Description

> Provide description here

## Type of change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Related issues

> Fix [#1]() 

## Migration notes

## Extra info


matiaslcoulougian:
just opened but then thought that with this we will need to do a backfill of sprints... closing and adding the change on dbt instead

PR #2276, opened 2025-09-22
FAI-18021 | Add inactive user detection to Claude users converter

dbruno21:
## Summary
- Added inactive user detection logic to Claude users converter following the pattern from GitHub Copilot seats converter
- Track current active users during conversion and mark missing users as inactive
- Query existing active users from Faros GraphQL API to identify users who are no longer active

## Changes Made
- **User Tracking**: Added `currentUsers` Set to track active users during conversion
- **Inactive Detection**: Implemented `onProcessingComplete()` method to query existing users and mark inactive ones
- **GraphQL Query**: Added USER_TOOL_QUERY to fetch existing active users from Faros
- **Test Updates**: Updated tests to mock GraphQL endpoint for proper testing

## Technical Details
- Uses the same pattern as GitHub Copilot seats converter for consistency
- Queries existing `vcs_UserTool` records to identify previously active users
- Sets `inactive: true` and `endedAt` timestamp for users no longer in current list
- Gracefully handles missing Faros client configuration

## Test Plan
- [x] Verify converter builds successfully
- [x] Test with mock GraphQL endpoint
- [x] Confirm inactive user detection logic
- [x] Validate test snapshots update correctly


destinations/airbyte-faros-destination/src/converters/claude/users.ts
```diff
@@ -51,20 +55,101 @@ export class Users extends ClaudeConverter {
...
+              organization: {
+                uid: this.streamName.source,
+                source: this.streamName.source,
+              },
+              tool: {
+                category: VCSToolCategory.CodingAssistant,
+                detail: VCSToolDetail.ClaudeCode,
+              },
+              inactive: true,
+              ...(!previousUser.endedAt && {endedAt: now.toISOString()}),
```
matiaslcoulougian, line 117: just in case, there's no way to obtain the actual date the user was set to inactive right? Cause this way the `endedAt` date will be tied to the next sync date after the user is set to inactive

matiaslcoulougian (approved):
LGTM, small nit question

